PR #54, opened 2023-09-25
[FEAT] Updated Tokens

### LINKED ISSUE #10: [FEAT] Design Tokens should allow bidirectional updates
**Description**:

As a designer, I want to the repositories tokens to automatically update when I change design tokens in Figma. As a developer, I want the Figma tokens to automatically update when I change one or more tokens in the repository. 

**Rationale**:

A bidirectional sync for our common design language is needed to prevent divergence between Design (Figma) and Development (components).

---

> ⚠️ The first draft for the tokens was already implemented in #9.
> You may use this package (@sick-design-system/design-tokens) as a base for further work.

reichst1SICKAG:
# Pull Request

### 🎫 Issues

Closes #10 

### General

- [ ] I have updated the project documentation to reflect my changes.

### Component

- [ ] I have added changes to an already existing component.
- [ ] I have written integration tests (or have adjusted already existing ones).
- [ ] I have created or updated any stories affected by my changes.


packages/design-tokens/tokens/semantic/typography-default.json
schilchSICKAG, line 1: I get multiple warnings because of this file. The issue may be caused because typo default and small are using the same definitions, but just one definition can be loaded at the same time. This leaves the question how we want to mitigate this. Should we skip semantic tokens completely, use typo-default or typo-small? Also, as far as I can tell, each token should have a `value` prop that has the final value included, but for those files we have a structure similiar to the one below:

```javascript
        "xs": {
          "value": {
            "fontFamily": "{font.family.default}",
            "fontWeight": "{font.weight.bold}",
            "lineHeight": "{font.lineheight.140}",
            "fontSize": "{font.size.size-200}"
          },
          "type": "typography"
        },
```

Suggestion:

Should use the following structure:

```javascript
        "xs": {
            "fontFamily": {
              "value": "{font.family.default}"
            },
            "fontWeight": {
              "value": "{font.weight.bold}"
            }
```

packages/design-tokens/tokens/core/color.json
schilchSICKAG, line 1: Colors are nested one level too deep, resulting in final names like `--sds-color-neutral-neutral-900` or `--sds-color-primary-primary-400`. Suggestion: Change this:

```javascript
  "color": {
    "primary": {
      "primary-3": {
        "value": "#E5F6FF",
        "type": "color",
        "description": "Primary-3"
      },
```

to

```javascript
  "color": {
    "primary": {
      "3": {
        "value": "#E5F6FF",
        "type": "color",
        "description": "Primary-3"
      },
```

to make names less verbose.

schilchSICKAG (changes requested):
I have found a couple of problems with the new tokens. We should talk about those before merging.

packages/design-tokens/tokens/core/color.json
reichst1SICKAG: yeah, the naming looks like duplicates which is not ;)
the first 'primary' is the group and the name ist 'primary-3' because just '3' is not a good name

Maybe it works better when we change the group name from primary to 'main' or 'brand' like `--sds-color-brand-primary-3`
This works quite good, if you look at the status colors.

packages/design-tokens/tokens/semantic/typography-default.json
reichst1SICKAG: Another idea is to remove those few semantic tokens for now, and insert them when we need them. Altough it would be good to have them in Figma to use them for documentation text or other stuff.

I only inserted them to test the semantic structure in Figma and to define the heading/body text definitions somewhere as tokens.

schilchSICKAG: I have removed the semantic tokens from the current bundle for the time being.

schilchSICKAG (approved):

schilchSICKAG:
Adjusted documentation to make it work again, removed semantic tokens from building for the time being.

PR #56, opened 2023-09-27
✨ Feat(core) #55: Rename package namespace to `@synergy-design-system`

### LINKED ISSUE #55: [FEAT] Adjust package names to use @synergy-design-system as prefix
All packages must be prefixed using `@synergy-design-system`

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the changes for renaming

### 🎫 Issues

Closes #55.

## 👩‍💻 Reviewer Notes

- I have renamed all static files and also the lock files.
- You will have to run `pnpm i -r` in the root to make sure all packages are linked correctly.
- The prefix in the current `component` package is still `sds-` and not `syn-`, as we will remove this package anyways.

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -7,7 +7,7 @@
     "@synergy-design-system/design-tokens": "workspace:^",
     "@synergy-design-system/icons": "workspace:^"
   },
-  "description": "Various CSS styles used in the SICK Design System",
+  "description": "Various CSS styles used in the Synergy Design System",
   "devDependencies": {
     "@synergy-design-system/eslint-config-sds": "workspace:^",
     "@synergy-design-system/stylelint-config-sds": "workspace:^",
@@ -38,6 +38,7 @@
     "Styles",
     "Reset",
     "SDS",
+    "Synergy Design System",
     "SICK",
     "SICK Design System"
   ],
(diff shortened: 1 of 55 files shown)


packages/css/package.json
```diff
@@ -4,13 +4,13 @@
     "url": "https://www.sick.com"
   },
   "dependencies": {
-    "@sick-design-system/design-tokens": "workspace:^",
-    "@sick-design-system/icons": "workspace:^"
+    "@synergy-design-system/design-tokens": "workspace:^",
+    "@synergy-design-system/icons": "workspace:^"
   },
   "description": "Various CSS styles used in the SICK Design System",
```
kirchsuSICKAG, line 10: Do we also need to adapt the naming here and say "Synergy Design System" instead of "SICK Design System"? Same question for the other package.json files

packages/css/package.json
```diff
@@ -42,7 +42,7 @@
     "SICK Design System"
```
kirchsuSICKAG, line 42: Do we also need to adapt the naming here and say "Synergy Design System" instead of "SICK Design System"? Or maybe just add it as keyword? Same question for the other package.json files

packages/css/package.json
schilchSICKAG: Good catch, will update accordingly

schilchSICKAG: Done in latest push

kirchsuSICKAG (approved):

PR #70, opened 2023-10-16
revised github templates

klaarseSICKAG:
# Pull Request

## 📖 Description

We decided to revise github templates for this repo

### 🎫 Issues

https://github.com/SickDesignSystem/Internal/issues/4


.github/ISSUE_TEMPLATE/component-request-new-component.md
```diff
@@ -9,19 +9,24 @@ labels: component/new, status/triage, enhancement
 Thank you for taking your time to request a new component for the Synergy Design System. Please make sure to answer the questions below to give us more information about your request.
 -->
 
-**Version**:
+**Questions**
 
-I am using the following version of the component library: _____
+- [ ] I have looked at the documentation site and was not able to find the component. <!-- link zur seite??? -->
```
schilchSICKAG, line 14: Please remove comment before pushing :)

schilchSICKAG: Site URL will be available in the local Documentation (e.g. README.md) when ready.

.github/ISSUE_TEMPLATE/component-request-new-component.md
```diff
@@ -9,19 +9,24 @@ labels: component/new, status/triage, enhancement
...
+- [ ] I have looked into the Figma boards and was not able to find the component. <!-- kommt jeder auf figma??? -->
```
schilchSICKAG, line 15: There will be a public Figma board later (I assume). A link will also be part of the documentation. Please remove comment before pushing ;)

.github/ISSUE_TEMPLATE/component-request-new-component.md
```diff
@@ -9,19 +9,24 @@ labels: component/new, status/triage, enhancement
...
+- [ ] I could (or even want) to implement the component myself
 
-**Questions**:
+---
 
-- [ ] I have looked at the documentation site and was not able to find the component.
-- [ ] I have looked into the Figma boards and was not able to find the component.
+**Description**
```
schilchSICKAG, line 20: I would suggest adding a whitespace here

.github/ISSUE_TEMPLATE/generic-bug.md
```diff
@@ -24,7 +24,7 @@ The issue happens in the following browser(s): _____
 
 **Questions**:
 
-- [ ] I have verified the bug exists in the latest version of `@synergy-design-system/core` (e.g. via `npm install @synergy-design-system/core@latest`).
+- [ ] I have verified the bug exists in the latest version of `@synergy-design-system/core` (e.g. via `npm install @synergy-design-system/core@latest`). <!-- core??? -->
```
schilchSICKAG, line 27: Good catch ;). Seems like a leftover from the original version. Should be `@synergy-design-system/components`.

schilchSICKAG (changes requested):

schilchSICKAG (dismissed):

mariohamann:
I see a valid use case for the checkboxes at the beginning for externals, but question this if it helps for our current process in our team. But I don't have strong opinions on that.
---

I personally would like to have "Rationale" at the beginning and explicitly rename it to "User Story" and provide a sample text we can edit.

Something like this:
> As a [PERSONA] of the Synergy Design System, I would like to [REQUIRE], as well as [FURTHER_REQUIREMENT], so that [PRODUCT_VALUE], and [FURTHER_PRODUCT_VALUE].

This helps especially in Refinements to get a common understanding why we're doing things.

mariohamann:
I'm missing a `chore` template. Currently all templates force us to new version numbers. Sometimes we have just stuff going on inside our repo that doesn't change any output, shouldn't bump a version and especially shouldn't add stuff to the changelog.

schilchSICKAG:
I have just merged back main into this branch and created a stale issue template for chores. @klaarseSICKAG: You could use this as a starting point for formulating a good template. I also think @mariohamann is right about the rationale: We have the chance to create a clean template that guides the user in creating issues, so it would be the right time to include something like this in the template.

.github/ISSUE_TEMPLATE/generic-docs.md
```diff
@@ -8,7 +8,7 @@ labels: docs, status/triage
 **Description**:
 
 <!--
-Please describe what you want to change for the documentation.
+Please describe what documentation should be changed.
 -->
 
 **Rationale**:
```
schilchSICKAG, line 14: You have used `# HEADLINE` semantics in other templates, I think we should either use the current version (`**HEADLINE**:`) or the new version (`# HEADLINE`) in all templates.

schilchSICKAG (changes requested):

schilchSICKAG (approved):

PR #88, opened 2023-10-20
feat: ✨ React component support

### LINKED ISSUE #43: feat: ✨ React component support
**Description**:

As a developer using Synergy, I want to be able to use the components in my React applications with correct intelli-sense, event and attribute handling.

**Rationale**:

Better DX for React developers.

### LINKED ISSUE #62: feat: ✨ syn-input: react wrapper
**Description**:

As a user, I want to be able to use the text-field in React the same way a native react component is used.

Best would be to generate it via the CEM

schilchSICKAG:
# Pull Request

## 📖 Description

This PR provides a new package named `@synergy-design-system/react` that makes it possible to use the web components from `@synergy-design-system/components` in React.js based applications in an easy way.

### 🎫 Issues

Closes #43
Closes #62

## 👩‍💻 Reviewer Notes

- There is currently no semantic-release support. This has to be done in another ticket.
- Please also read the provided `README.md` to make sure I have not forgotten anything :)
- I refactored the original build system to a more streamlined and modular system. I think this better fits our needs than the original system from components. Looking forward for your opinions :)

## 📑 Test Plan

- Check if all scripts work as intended.
- Check if the provided demo works.
- Play around with the demo


mariohamann:
@schilchSICKAG I thought we aligned on having the build step in the components package and the react-package just being a "wrapper" package, which is automatically generated in the components build step with added `react` package in package.json? Why are there explicit component files and duplicated build files fin the react package? How is the plan to have the build steps in sync? On top I see problems with semantic release, which analyzes commits in a package – if we change something in components, the changes wouldn't be recognized in the react-package. What are the benefits over the current behaviour of Shoelace, which does all components automatically?

schilchSICKAG:
> * I thought we aligned on having the build step in the components package and the react-package just being a "wrapper" package, which is automatically generated in the components build step with added `react` package in package.json. Why would you deviate from this behaviour? Did you see problems here?

We will have a build step that calls `exec` for the react build in the components package. There will be problems in the future when we do not go this (or a similar) route. For example, we will not be able to publish the angular wrappers in the same way because they need a different build infrastructure that does not mix well with the one we have in shoelace currently.

The other problem is that you are doing this blind. When you just create the wrappers you **assume** they are correct when bundling and publishing them.

I have added a simple vite configuration so we can play with the wrappers and see if they are working. This is currently a manual step (and I do not plan to automate this currently), but at least you can have a look at the components that you are shipping.

> * Why are there explicit component files and duplicated build files in the react package? I see a ton of manual updating here and I don't see them being integrated with `vendorism.js`.

The build directory is currently partly duplicated. The reason for this is that I did not want to do the build-script refactoring in the same task. We already have left the path of shoelace when it comes to building, so we can create a cleaner build infrastructure in the long term.

My plan would be to make the react/scripts/jobs folder a shared internal package next and add the steps in there, so we can reference them from components, react and angular. With this, we will be better handle the build in the long run, making it go from 141 lines (current) to maybe 10 with shared features (for example the cleanup and esbuild jobs).

When using vendorism, you will also have to do a manual build in the components directory. I would add the step as detailed above.

For vendorism, maybe it would also feasable to allow for function callbacks on before and after hooks for the future? Something like:

```javascript
// Vendorism-config:
// ...more config
hooks: {
  // before: "echo ⌛️ Setting up target...",
  // after: "echo ✅ Target setup complete.",
  before: () => {
    console.log('⌛️ Setting up target...'),
  },
  after: (files)) => {
    console.log(files); // array of changed files so we can do something with them if needed?
    console.log('✅ Target setup complete. Creating new build...');
    exec('scripts/build.js');
  }
},
// ...more config
```

> * How is the plan to have the build steps in sync?

There should be no duplicated build files, but only the raw react components and their dependency for the wrapped components, as outlined above.

> * On top I see problems with semantic release, which analyzes commits in a package – if we change something in components, the changes wouldn't be recognized in the react-package.

There are multiple ways to tackle this. My approach would be to reuse the config in semantic release (e.g. I would add the following):

```json
      [
        "@semantic-release/git",
        {
          "message": "chore(release/components): ${nextRelease.version} [skip actions]\n\n${nextRelease.notes}",
          "assets": [
            "CHANGELOG.md",
            "package.json"
          ]
        }
      ],
```

to the react package. I did not want to do this right now because the issue with semantic release building currently is in review.

> * What are the benefits over the current behavior of Shoelace, which does all components automatically? And do they really outweigh the burden which comes with the approach being outlined here?

First, shoelace does always ship with the `lib/react` package which makes no sense for the components package. It also has devDependencies on `react` and `react-dom` for the same reason. I would like to separate it as we did with the docs package to have a clean components package. Otherwise, we would also have to add angular as a dependency in the components package to create its meta build, which feels unclean.

> Summing up: From a maintenance perspective I see way too much manual effort here and would have expected a more generic solution, which doesn't involve our developers as much as possible.

From my point of view, developers will not need to do anything at all with the package when the shared build approach is used in the future. The wrappers will make sure to always have the latest version of the components package via its current build step. Data is always drawn directly from components as a single source of truth and the wrappers will only change when there is for example a new component.

> As long as automation is possible, we should use it.

That is true and we will be able to archive it with this approach when the build script is integrated in the components package.

> I don't have any opinions on the final output and how it looks like, but from my perspective we should ensure, that effort and DX on our isn't harmed because of a wrapper.

I think we should discuss that tomorrow as I do not see a loose in DX. However, maybe there are some things I am missing. Looking forward for a discussion on this.

.github/workflows/release.yml
```diff
@@ -26,21 +26,8 @@ jobs:
...
-          NPM_TOKEN: ${{ secrets.NPM_RELEASE_TOKEN_CS }}
-        run: cd packages/components && pnpm release
-
-      - name: (tokens) Run Semantic Release
-        env:
-          GH_TOKEN: ${{ secrets.GH_RELEASE_BUILDUSERSICKAG }}
-          NPM_TOKEN: ${{ secrets.NPM_RELEASE_TOKEN_CS }}
-        run: cd packages/tokens && pnpm release
-
+        run: pnpm build && pnpm release && pnpm publish --tag alpha
```
mariohamann, line 33:
```suggestion
        run: pnpm build && pnpm release && pnpm publish --tag main
```

Shouldn't this be `main` to stick to the actual version name?

package.json
```diff
@@ -21,7 +21,7 @@
   "scripts": {
     "build": "pnpm --recursive --if-present build",
     "lint": "pnpm --recursive --if-present lint",
-    "cli": "node ./scripts/unified-design-system.js"
+    "release": "pnpm --recursive --if-present release"
```
mariohamann, line 24: ❤️

packages/_internal/stylelint-config-syn/package.json
```diff
@@ -35,6 +35,7 @@
   ],
   "license": "BSD-3-Clause",
   "name": "@synergy-design-system/stylelint-config-syn",
+  "private": true,
```
mariohamann, line 38: Should we rename the folder `_internal` to `_private` to have a common naming?

packages/assets/package.json
```diff
@@ -73,7 +73,7 @@
       [
         "@semantic-release/npm",
         {
-          "npmPublish": true
+          "npmPublish": false
         }
       ],
```
mariohamann, line 78: Do we have to add the plugin at all if don't publish?

packages/react/README.md
```diff
@@ -0,0 +1,86 @@
...
+To create a new version of this package, proceed in the following way:
+
+1. Check out the [Synergy Design System Repository](https://github.com/SickDesignSystem/synergy).
+2. Run `pnpm i -r` to install all dependencies.
+3. Build the `@synergy-design-system/components` package (or run `pnpm build in the project root to build everything).
+4. Move to to `packages/react` and use `pnpm start` to spin up a local vite project using react and typescript to validate your changes.
+5. Apply all fixes displayed when using `pnpm lint`.
+
+> ⚠️ The build process will always try to sync this packages `package.json.version` field with the latest version from `@synergy-design-system/components`!
+> Therefore, it is best to not alter the version string
```
mariohamann, line 86: Is this still valid after your changes?

packages/react/package.json
```diff
@@ -0,0 +1,122 @@
...
+    "eslint-plugin-jsx-a11y": "^6.7.1",
+    "eslint-plugin-react": "^7.33.2",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "eslint-plugin-react-refresh": "^0.4.3",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "typescript": "^5.0.2",
+    "vite": "^4.5.0"
+  },
+  "description": "React wrappers the Synergy Design System",
```
mariohamann, line 31:
```suggestion
  "description": "React wrappers for the Synergy Design System",
```

packages/react/package.json
```diff
@@ -0,0 +1,122 @@
...
+    "React Components",
+    "React Design System",
+    "Design System",
+    "SDS",
+    "Synergy Design System",
+    "SICK",
+    "SICK Design System"
+  ],
+  "license": "BSD-3-Clause",
+  "name": "@synergy-design-system/react",
```
mariohamann, line 56: Remember that you have to publish once locally, that the actions run correctly on GitHUb.

.github/workflows/release.yml
schilchSICKAG: Good question. I have opted for alpha as this is the way most packages do stuff like that when using publish tags. For me (personally), it is more common ground installing package scopes via `npm install --save @synergy-design-system/components@alpha` than via `npm install --save @synergy-design-system/components@main`. But I am open for this as we do not have yet an audience ;)

packages/_internal/stylelint-config-syn/package.json
schilchSICKAG: > Should we rename the folder `_internal` to `_private` to have a common naming?

Propably, I will change it in the branch directly.

.github/workflows/release.yml
mariohamann: Yah, don't have an opinion neither. It's just that we already published something with main and we're currently in prerelease pahse.... but I don't care. 😆

packages/assets/package.json
schilchSICKAG: I have left it in because in this case, it will not publish, but will at least adjust the `package.version` field.

mariohamann: Doesn't do it without the package? :O

schilchSICKAG: semantic-release is unopinionated, we could also have a python project where everything is managed with pip. Think this is the reason you need the module at all. See https://www.npmjs.com/package/@semantic-release/npm#options

packages/react/README.md
schilchSICKAG: No, I am currently in the process of removing all other stuff in the react folder so we get a real auto-generation. The build information will get removed, but we need to have a README.md at least.

.github/workflows/release.yml
schilchSICKAG: Done in latest commmit

packages/_internal/stylelint-config-syn/package.json
schilchSICKAG: Done in latest version

packages/react/README.md
schilchSICKAG: Plese have a look at the latest version of the readme. Builds are now always only triggered via `pnpm build` and the demo was moved to `_private/react-demo`, where it still works fine.

packages/react/package.json
schilchSICKAG: We have to create a new issue for this to make it work

packages/react/package.json
```diff
@@ -0,0 +1,97 @@
...
+        }
+      ],
+      [
+        "@semantic-release/github",
+        {
+          "successComment": false
+        }
+      ]
+    ]
+  },
```
mariohamann, line 85: We don't need this here, as versions and everything are handled from the components package.

packages/react/package.json
mariohamann: Why? You can just publish locally with the current version. We just need it once there as public before we can use Semantic Release actions.

schilchSICKAG: Ah, now I get it. Will do when your review is done

packages/_private/react-demo/README.md
```diff
@@ -0,0 +1,86 @@
+# @synergy-design-system/react
```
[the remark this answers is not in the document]
schilchSICKAG: I have updated this to reflect the nature of the demo package.

packages/react/package.json
schilchSICKAG: Removed in latest version

mariohamann (approved):

PR #108, opened 2023-11-02
feat: ✨ add syn-input

Gwenevere:
# Pull Request

## 📖 Description

The syn-input is a form field used to input any kind of single-line text, which could string, number etc...
This PR serves as a way to add the syn-input component with the new styling to the component library. 

### 🎫 Issues

- https://github.com/SickDesignSystem/synergy/issues/57

## 👩‍💻 Reviewer Notes

- I recommend that this is looked at by @LeaLeRose and @eileenwegner to make sure the design and functionalities are correct and work as expected.

### General

- [x] I have updated the project documentation to reflect my changes.

### Component

- [x] I have added changes to an already existing component.
- [x] I have written integration tests (or have adjusted already existing ones).
- [x] I have created or updated any stories affected by my changes.


mariohamann:
@Gwenevere Please re-merge main to have a cleaner baseline. :)

mariohamann:
@Gwenevere Please make an interactive story for the error state, so that we get a robust screenshot test.

packages/docs/stories/components/input.stories.ts
```diff
@@ -97,6 +97,28 @@ export const ReadonlyInputs: Story = {
   },
 };
 
+/** The focus attribute provides feedback to the users, informing them that the input component is ready for use.  */
```
mariohamann, line 100: It's not an attribute. :) Please rephrase, maybe together with @LeaLeRose so she can provide it in Figma as well.

mariohamann:
The error state is not showing up:

CleanShot 2023-11-13 at 09 27 44

Maybe you have to wait until the input is ready, too.

mariohamann:
@LeaLeRose @Gwenevere What I realize: "error" is not the correct term, we would have to rename it to "invalid". Could you please do in Figma and Code?

packages/docs/stories/components/input.stories.ts
Gwenevere: Yeah, I wasn't sure what to write

LeaLeRose: the documentary is now a lot more than the things we provide in Figma or Shoelace? Is this planned? do I have to change this as well?

packages/components/.stylelintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "@synergy-design-system/stylelint-config-syn"
+  "extends": "@synergy-design-system/stylelint-config-syn",
+   "rules": {
```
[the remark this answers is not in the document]
schilchSICKAG: This is now in `@synergy-design-system/stylelint-config-syn`. I have opted to remove the bits that where just floating around.

packages/components/scripts/vendorism.js
```diff
@@ -155,13 +155,13 @@ const config = {
       // add custom styles to the end of `${component}.styles.ts`
       (path, content) => {
         let newContent;
-        components.forEach((component) => {
+        [...components, 'form-control'].forEach((component) => {
```
schilchSICKAG, line 158: Where is this coming from? What are we doing with `form-control` here?

packages/components/src/components/input/input.custom.styles.ts
schilchSICKAG, line 1: Currently, there is no dark-mode support as it seems?

packages/components/web-test-runner.config.js
schilchSICKAG, line 1: Why was this removed? The configuration is needed for running the integration tests.

packages/docs/package.json
```diff
@@ -25,6 +26,8 @@
     "@storybook/web-components-vite": "^7.5.3",
     "@synergy-design-system/eslint-config-syn": "workspace:*",
     "@synergy-design-system/stylelint-config-syn": "workspace:*",
+    "@testing-library/react": "^14.1.0",
```
schilchSICKAG, line 29: What is this needed for?

packages/docs/src/helpers/component.ts
schilchSICKAG, line 1: Would need @mariohamann opinion on this, as this currently fixes the (still persisting) problem of wrong control types. I noticed it just when using the input, even with newest storybook. I think I also know where it is coming from, but wanted a quick fix here.

Problem seems to be whitespace in `components.json` in the attributes field. It has an empty line on the definition, which may trick your plugin of thinking it is numeric (somehow?).

packages/components/src/components/input/input.custom.styles.ts
```diff
@@ -1,5 +1,85 @@
...
+    font-size: var(--syn-font-size-2x-large);
+   }
+  
+  :host([data-user-invalid]) .input--standard {
+    border-color: var(--syn-input-border-color-focus-error);
+  }
+
+  :host([data-user-invalid]) .input--standard.input--focused:not(.input--disabled) {
+    border-color: var(--syn-input-border-color-focus-error);
+    box-shadow: 0 0 0 var(--syn-focus-ring-width) var(--syn-input-focus-ring-error);
```
schilchSICKAG, line 78: Focus-Ring-Width is currently set to `1px`, however, it should use `2px` in Figma. This is the reason it looks so strange currently: Just drawing a 1px blue border around a black border does not quite look nice ;). Please talk to @mariohamann about how to proceed here.

packages/components/src/styles/form-control.custom.styles.ts
schilchSICKAG, line 1: @mariohamann: Should we add an indicator to labels (e.g. at least `cursor: pointer`?)

schilchSICKAG (changes requested):

packages/docs/package.json
Gwenevere: to create interactive stories, like the invalid one, where a user's action is mimicked

packages/assets/.env.example
```diff
@@ -0,0 +1 @@
+FIGMA_PERSONAL_ACCESS_TOKEN = ""
```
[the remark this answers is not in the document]
Gwenevere: It's not needed for the input, is needed for the fetching of assets. It is documented, yes. I just hadn't added this when I worked on the project's packages

packages/components/scripts/vendorism.js
Gwenevere: the form-control is a component that is common to some components like the input and the textarea, since they are input components. It is a part of them.

schilchSICKAG (approved):
Ok from my point of view

PR #114, opened 2023-11-06
docs/syn-radio

LeaLeRose:
### General

- [ ] I have updated the project documentation to reflect my changes.

### Component

- [ ] I have added changes to an already existing component.
- [ ] I have written integration tests (or have adjusted already existing ones).
- [ ] I have created or updated any stories affected by my changes.

### DIFF
--- a/packages/tokens/src/figma-tokens/_docs.json
+++ b/packages/tokens/src/figma-tokens/_docs.json
@@ -584,16 +584,6 @@
         "type": "text"
       }
     },
-    "radio": {
-      "description": {
-        "value": "Radios allow the user to select a single option from a group.",
-        "type": "text"
-      },
-      "title": {
-        "value": "Radio",
-        "type": "text"
-      }
-    },
     "range": {
       "description": {
         "value": "Ranges allow the user to select a single value within a given range using a slider.",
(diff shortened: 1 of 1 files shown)


packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -773,6 +763,54 @@
...
       }
+    },
+    "radio": {
+      "default": {
+        "title": {
+          "value": "Radio",
+          "type": "text"
+        },
+        "description": {
+          "value": "Radios allow the user to select a single option from a group. \nRadios are designed to be used with radio groups.",
```
schilchSICKAG, line 774: Does that mean we need a `<syn-radio-group />` element to make use of `<syn-radio />`? In default html, radios are defined as a group of radio controls that have the same `name` attribute and do not require a wrapper.

packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -773,6 +763,54 @@
...
+          "type": "text"
+        }
+      },
+      "examples": {
```
schilchSICKAG, line 778: In https://github.com/SickDesignSystem/synergy/pull/112/files#diff-35864d9aee16b937faed6a2d8d9b635446c65983c9c77c1bbd86b8dac5d2e16cR749, there is a mention of a label property for text fields and areas, which makes completely sense. However, I think that for the sake of completeness, radio buttons should also allow a label to be used. Either it is not documented or missing from the component at all.

@mariohamann : What do you think about this?

packages/tokens/src/figma-tokens/_docs.json
mariohamann: Yes. https://shoelace.style/components/radio-group How did you solve this with DaVinci?

packages/tokens/src/figma-tokens/_docs.json
mariohamann: It's in the group.

https://shoelace.style/components/radio-group

mariohamann: + radios theirself have a label, too. It's just the default slot there.

packages/tokens/src/figma-tokens/_docs.json
schilchSICKAG: In our current library, there is just the plain element without wrapper and without label. As we do not offer real form support, this is not that big of a deal currently, but we want to make it better ;)

packages/tokens/src/figma-tokens/_docs.json
mariohamann: Yes, Shoelace does it providing the `name` attribute in the group: https://shoelace.style/components/radio-group

So it would be:

```
<form>
  <syn-radio-group name="test">
    <syn-radio value="item1">Item 1</syn-radio>
    <syn-radio value="item2" checked>Item 2</syn-radio>
    <syn-radio value="item2">Item 3</syn-radio>
  </syn-radio-group>
  <syn-button type="submit">Submit form</syn-button>
</form>
```

schilchSICKAG: Ok, that not quite as standard compliant as I was hoping for, but I see where this comes from and it would be more of a pain to eventually adopt it to make it work like it should.

mariohamann (approved):

PR #118, opened 2023-11-06
feat: ✨ add syn-button

Gwenevere:
# Pull Request

## 📖 Description

The syn-button should also work as form element to submit form-data. Furthermore it should work like a link if desired.
This PR serves as a way to add the syn-button component with the new styling to the component library.

### 🎫 Issues

* https://github.com/SickDesignSystem/synergy/issues/52

## 👩‍💻 Reviewer Notes

* I recommend that this is looked at by @LeaLeRose and @eileenwegner to make sure the design and functionalities are correct and work as expected.

## ✅ Checklist


packages/docs/stories/components/button.stories.ts
```diff
@@ -16,12 +10,23 @@ const { args, argTypes } = storybookDefaults('syn-button');
 const { overrideArgs } = storybookHelpers('syn-button');
 const { generateTemplate } = storybookTemplate('syn-button');
 
+const generateStoryParameters = (attributeName: string) => {
```
mariohamann, line 13: Please reduce this function on only the description, not the rest of the parameters. :)

packages/components/src/components/button/button.custom.styles.ts
```diff
@@ -1,5 +1,100 @@
...
+  
+  .button--text:hover:not(.button--disabled) {
+    color: var(--syn-color-primary-900);
+  }
+
+  .button:focus {
+    outline: var(--syn-focus-ring-color) solid 2px;
+    outline-offset: 2px;
+  }
+
```
reichst1SICKAG, line 65: button:active seems to be missing here for the 'pressed' state.
And also active gets the keyboard-focus ring.

reichst1SICKAG (commented):
Pressed state gets the focus ring

packages/components/src/components/button/button.custom.styles.ts
reichst1SICKAG: Hi @Gwenevere ,
I'm sorry but my comment was completely misunderstanding. The 'pressed' state should NOT get the keyboard focus-ring and also the bg-color is darker: 
button-pressed-state

Gwenevere: Ahhh, yeah, I misunderstood

packages/docs/stories/components/button.stories.ts
```diff
@@ -1,10 +1,4 @@
-// ---------------------------------------------------------------------
-// 🔒 AUTOGENERATED BY VENDORISM
-// Removing this comment will prevent it from being managed by it.
-// ---------------------------------------------------------------------
-
 /* eslint-disable */
-
 /* eslint-disable import/no-relative-packages */
```
mariohamann, line 2: Please remove eslint-disable here and for all other stories in the other PRs. As this code is under our control after ejecting, it should fit to our coding guidelines.

packages/components/.stylelintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "@synergy-design-system/stylelint-config-syn"
+  "extends": "@synergy-design-system/stylelint-config-syn",
+  "rules": {
```
schilchSICKAG, line 3: As this will also happen when developing components outside of @synergy scope and we would like to publish `@synergy-design-system/stylelint-config-syn`, I think it would be nice if you could add this to the ruleset of the stylelint-package in `_private` instead of the package directly. With that, we are more future proof.

packages/components/src/components/button/button.custom.styles.ts
```diff
@@ -1,5 +1,133 @@
 import { css } from 'lit';
 
 export default css`
-  /* Write custom CSS here */
-`;
+.button:focus-visible {
+    outline: var(--syn-focus-ring-color) solid 2px;
```
schilchSICKAG, line 5: I think the outline-width should also be a a design token so we can share it across other components, too. Example (naming to be defined!):

```css
outline: var(--syn-focus-ring-color) solid var(--syn-focus-ring-width);
outline-offset: var(--syn-focus-ring-offset);
```

packages/components/src/components/button/button.custom.styles.ts
```diff
@@ -1,5 +1,133 @@
...
+
+  .button.button--large.button--has-label .button__label {
+    font-size: var(--syn-font-size-large);
+  }
+
+  /*
+ * Standard buttons
+ */
+
+  .button--filled.button--primary.button--disabled {
```
schilchSICKAG, line 25: Why did the other button variants (e.g. accent, secondary, danger, ...) disappear? Was this intentional? Our current Design system ships applications that make use of those, so it would be pretty breaking change to just have text/outline/default available instead of different colors.

packages/docs/package.json
```diff
@@ -31,7 +32,6 @@
     "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-storybook": "^0.6.15",
     "html-format": "^1.1.2",
-    "lit": "^3.0.0",
```
schilchSICKAG, line 34: Not a big deal, but why was this moved to `dependencies`? As we will never ship @synergy-design-system/docs to npm, its not that big of a problem, but still I am not sure why this was added here?

packages/docs/stories/components/button.stories.ts
```diff
@@ -1,268 +1,226 @@
...
+  parameters: {
+    docs: {
+      description: generateStoryDescription('link'),
+    },
+  },
+  render: () => html`
+  <syn-button href="https://example.com/">Link</syn-button>
+  <syn-button href="https://example.com/" target="_blank">New Window</syn-button>
+  <syn-button href="/assets/images/wordmark.svg" download="synergy.svg">Download</syn-button>
+  <syn-button href="https://example.com/" disabled>Disabled</syn-button>
```
schilchSICKAG, line 97: Error: Disabled links with a href should not only be displayed as disabled, but also not change the location. Currently, when clicking on the Disabled Link, it will still open example.com.

schilchSICKAG (changes requested):
Hey @Gwenevere,

just had a chance to review the changes needed for the button. There are some smaller issues and adjustments I would like to make, as detailed in this Review.

Also, I wondered why the various button variants have disappeared?

packages/docs/stories/components/button.stories.ts
mariohamann: `<a>` doesn't have a disabled attribute in HTML. Disabled links usually just don't get an `href` to be correctly (non-)accessible. We should just remove this example here completely. If people don't want that a link is clickable, they should remove the link as well, otherwise we would have to build a lot more on top just for the link, e. g. making it not focusable etc.

packages/components/src/components/button/button.custom.styles.ts
mariohamann: We plan to provide more colors in future, but currently don't provide them in design as well. We have to clarify stuff like a11y of those colors, if provide error and danger colors as well.

I personally would provide every color as well for filled, outline and text.

packages/components/src/components/button/button.custom.styles.ts
mariohamann: Good point. Let's do that in a follow-up ticket, so we can merge this soon and design can provide it then.

packages/docs/stories/components/button.stories.ts
schilchSICKAG: That is the reason why I think it may be better to have two different components for that, which could also be done by sharing a render template. But we already talked about this once I think. Point is, people will eventually run into this and wonder why it does not work.

A very simple fix for this would be something like:

```css
syn-button[href][disabled] {
  pointer-events: none;
}
```

mariohamann: Then it would be still accessible via keyboard which is not correct from a11y perspective.

mariohamann: It might make more sense to provide a PR for that to only set `href` if the button is not disabled. What do you think about that?

schilchSICKAG: Thats also a good point. However, there seem to different arguments about this too:

> If disabled buttons are not focusable with TAB, our user would occasionally miss that the button ever exists, and they will never find it.

We should at least be thinking about this. Like I said, we will get this as an error eventually more sooner than later.

packages/components/src/components/button/button.custom.styles.ts
schilchSICKAG: Thats fine for me. Will you create the tickets for it?

packages/docs/stories/components/button.stories.ts
schilchSICKAG: > more

Sorry, we seem to have been typing at the same time :). Sure, we can do that. Mind to create the ticket for it?

packages/components/src/components/button/button.custom.styles.ts
mariohamann: Yes. You have a danger button on your side? That's interesting, because it's not on brand portal AFAIK.

packages/docs/stories/components/button.stories.ts
mariohamann: > If disabled buttons are not focusable with TAB, our user would occasionally miss that the button ever exists, and they will never find it.

This is how every button in HTML works. If a button is disabled it's not focusable. If a link doesn't have a href, it's not focusable as well. The problem is here more setting a href and making it disabled at the same time. (To be honest, I'm not sure, if this will ever be a use case. :D)

schilchSICKAG: Right. But as you have pointed out, this is not a button, but a link. Which is kind of the problem :)

Usecase:

Wizard Form (Step 1, Step 2, Step 3)
You are on Step 2 (part of a bigger Form) and Form validation does not permit you to move to /account/create/step3

So seen in various open source ecommerce platforms like Magento, or parts of Shopify. Both platforms would try to carry the Session-Data to the next page to prevent issues when reloading.

packages/components/src/components/button/button.custom.styles.ts
schilchSICKAG: I think that this is a great idea, giving much more freedom for forming out what you want to tell the developer, instead of giving him some 12 or 14 different button shapes. That said, we did not even multiply those out in our current system, so people added feature requests for primary-outline-without-text-but-only-on-hover (you name it) ;)

packages/docs/package.json
Gwenevere: the linter flagged the previous location as wrong. That's why I changed it. But if not needed, I can change it back.
Screenshot 2023-11-15 at 12 28 44

packages/docs/stories/components/button.stories.ts
Gwenevere: Sooooo.... should I change something still? 😆

schilchSICKAG: No, as @mariohamann has pointed out, we will tackle this issue in a new ticket.

packages/components/src/components/button/button.custom.styles.ts
reichst1SICKAG: Mario explainded it very good, thanks.

We also wanted to remove the orange button for a11y reasons.

Red and green buttons are not a requiremet I've heared, but easy to add if the applications needs it somewhere eg. some dangerous deletion thing.

packages/docs/stories/components/button.stories.ts
Gwenevere: thank you for clarifying :)

mariohamann: @Gwenevere  But you should remove the example from the docs, as they are misleading.

@schilchSICKAG What do you think about improving our docs on dev side for the disabled state to make that more clear? (`If you need disabled links, set `disabled` and do not set the `href`.) But yeah, let's tackle this in an iteration and let's write the ticket open enough, because there are several solutions (e. g. docs, PR, adding pointer-events to prevent at least mouse clicks etc.). :)

schilchSICKAG: Seems like a good idea. We can define the ticket tomorrow together, makes it easier :)

schilchSICKAG (approved):
With the comments and tickets created like discussed with @mariohamann, I think we are good to go.

PR #130, opened 2023-11-13
chore: 🔧 enable integration tests for components in CI/CD

### LINKED ISSUE #48: feat: ✨ enable integration tests for components in CI/CD
---

# Description

As a developer, I want to be able to end-to-end check the element.

# Rationale

Changes in visualization must be verified before a release for various browsers. Also verification when browser versions change is necessary.

# Acceptance criteria

- (Shoelace) tests based on playwright are running locally and in CI pipeline

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds support for automatically running tests for all packages. Currently, only `@synergy-design-system/components` has any tests attached to it.

### 🎫 Issues

Closes #48 

## 👩‍💻 Reviewer Notes

Make sure tests are running locally using `pnpm i -r && pnpm test -r`. Browsers should be downloaded automatically.

## ✅ Checklist


.github/workflows/quality-check.yml
```diff
@@ -3,31 +3,66 @@ name: Quality Gate
...
+
+    - name: Run Playwright tests
+      run: pnpm test -r
+    
+    # - uses: actions/upload-artifact@v3
+    #   if: always()
+    #   with:
+    #     name: playwright-report
+    #     path: playwright-report/
+    #     retention-days: 30
```
mariohamann, line 68: Needed?

packages/components/package.json
```diff
@@ -23,10 +23,14 @@
...
   },
-  "files": ["dist"],
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "build": "node scripts/build.js",
-    "test": "pnpm build && web-test-runner --group default",
+    "test.setup": "mkdir -p dist/themes && pnpm run -C ../tokens build && cp ../tokens/dist/themes/light.css dist/themes/light.css",
+    "test.download": "pnpm playwright install-deps && pnpm playwright install",
```
mariohamann, line 32: This is something one doesn't in day-to-day work. It looks like this is espeically for CI/CD and would therefore suggest to use tha playwright images as done in Solid: https://github.com/solid-design-system/solid/blob/c45e606ae9f6e8ada422e773dba24b4bf4101224/.github/workflows/check-pr.yml#L10

.github/workflows/quality-check.yml
schilchSICKAG: Will be needed in the future for screenshot regression tests and is the recommended way when using playwright later on, so I would opt to let it stay in for the time being.

packages/components/package.json
schilchSICKAG: This was altered to align it better. However, during testing, I ran into an issue described at https://github.com/microsoft/playwright/issues/6500. After adding the fix, everything worked. However, I don´t know why this does not seem to be an issue in Solid. Would recommend to have an eye on that if you will run into this issue there.

packages/components/package.json
```diff
@@ -23,10 +23,12 @@
...
     "./translations/*": "./dist/translations/*"
   },
-  "files": ["dist"],
+  "files": [
+    "dist"
+  ],
   "scripts": {
     "build": "node scripts/build.js",
-    "test": "pnpm build && web-test-runner --group default",
+    "test": "web-test-runner --group default",
```
mariohamann, line 31: I fear that this could lead to people testing on old versions if we don't do a fresh build here. What do you think?

schilchSICKAG: Good point. This is also done in Synergy in this way. When using something like `test:watch` (which we currently are not implementing), an automatic build would be also feasable I think. I would like to create a new ticket for this, maybe we can wire it into `packages/components/web-test-runner.config.js` to make sure that it also builds?

It would also be nice in this case when we use a stripped down building process (e.g. ignoring the angular and react wrappers), as they are causing more strain on the system in the first place. What do you think?

mariohamann: Definitely. The latter slows the process a lot.

mariohamann: Let's do a new ticket for that.

mariohamann (approved):

packages/components/package.json
schilchSICKAG: Done at https://github.com/SickDesignSystem/synergy/issues/141

PR #154, opened 2023-11-17
chore: 🔧 Adjust components build so it exports all components automatically

### LINKED ISSUE #146: chore: 🔧 Adjust components build so it exports all components automatically
# Description

As a developer of the Synergy Design System, I do not want to think about manual file exports when vendoring new components. Therefore, the exports from vendorism should be automatically included in our main exports file (`packages/components/src/synergy.ts`).

# Rationale

When using this as a manual step, the risk is high to break the generation of wrappers, even if it does not have any sideeffects on the created components bundle by itself.

schilchSICKAG:
# Pull Request

## 📖 Description

As we already forgot multiple times to export the created components in `src/synergy.ts` and our framework wrappers rely on the exports in this file, we have to make sure the exports are added without manual resolution. This PR adds the changes in the build that makes this possible.

### 🎫 Issues

Closes #146 

## 👩‍💻 Reviewer Notes

The main changes where done in `packages/components/scripts/jobs/createExports.js`.

## 📑 Test Plan

1. Check out this branch
2. Make sure dependencies are up to date (via `pnpm i -r`)
3. Go to `packages/components` and add a component to `scripts/config.js`
4. Vendor the new component
5. Issue a new build via `pnpm build`
6. Expection: The file `src/synergy.ts` should include a new export that is named after the newly included component.

## ✅ Checklist


mariohamann:
I think we have to be careful to not make too much stuff automatically and create weird dependencies between the scripts. This change would lead us to create a build before we can test new components in Storybook.

I would propose a different way: Let's make synergy.ts the single source of truth and let vendorism.js parse it for components. If a same-named component exists in Shoelace: Vendor it, if not: Do nothing.

mariohamann (changes requested):
See comment above.

schilchSICKAG:
> I think we have to be careful to not make too much stuff automatically

We have two new components that did not export in `synergy.ts`, which leads to errors when creating wrappers and stuff.

> and create weird dependencies between the scripts, which could get out of control.
> This change would lead us to create a build before we can test new components in Storybook.

It would not, as the components used in Storybook are directly referenced.

But the build has a different purpose.
> 
> I would propose a different way: Let's make synergy.ts the single source of truth and let vendorism.js parse it for components. If a same-named component exists in Shoelace: Vendor it, if not: Do nothing.

In this case, you would need to have to parse the whole file server side (which may be a problem as it uses ts and I would not want to write a regex which is very error prone). If you add the export to the file in the first place, it will break the file because it will not be able to find the components you just want to add in the first place, which may be more error prone than generating it as part of the build alltogether.

mariohamann:
Understand, thanks for clarification. Let's give it a try how it feels.

I believe this will become even more important then: 
https://github.com/SickDesignSystem/synergy/issues/149

mariohamann:
Still, I have the feeling, that there is to much stuff hidden. The config.js in scripts influences which elements are finally available, which should be vendored etc.

Maybe this is part of a new ticket, but I would love to omit this file completely to have clearer purposes and would like to have more something like this:

1. vendorism.js decides which files and components have to be vendored
2. build.js graps all files from `/events` and all folder names from `/components` and creates the correct output files from it.

What do you think?

mariohamann:
Otherwise people could manually add a components folder with a new component... and it still isn't exported, but shows up in Storybook.

schilchSICKAG:
> Still, I have the feeling, that there is to much stuff hidden. The config.js in scripts influences which elements are finally available, which should be vendored etc.
> 
> Maybe this is part of a new ticket, but I would love to omit this file completely to have clearer purposes and would like to have more something like this:
> 
> 1. vendorism.js decides which files and components have to be vendored
> 2. build.js graps all files from `/events` and all folder names from `/components` and creates the correct output files from it.
> 
> What do you think?

We will definitively have to adjust this. However I think it is more tricky than we initially tought:

1. Vendorism is just there to vendor the SL components
2. When vendoring a component, it will draw the dependencies in, too
3. We agreed at not exporting the dependencies and it has to be a manual step to add them to the external visible bundle, even when they are available internally.
4. Because of this, I am not able to just read the component contents currently:
5. There where dependencies (like `syn-spinner`) that we chose to not export. So, should I create the wrappers for them reguardless?
6. The current decision was to use the config file as single source of truth to make sure vendorism and wrapper builds stay in sync.
7. However: This will not work anymore when we create our first own component. So we have to adjust it somehow.

### Proposal:

- We let vendorism do its thing alone for components and events.
- We always export all components in the components package `synergy.ts`
- We always export all components in angular/react.

This would definitively have to be a new ticket for sure.

mariohamann:
Sounds good. Could you create a ticket? :)

mariohamann (approved):

schilchSICKAG:
> Sounds good. Could you create a ticket? :)

Done at #156. Please have a look and change accordingly if something is unclear in the new ticket.

PR #155, opened 2023-11-17
feat: ✨ syn-checkbox

### LINKED ISSUE #72: feat: ✨ syn-checkbox
**Description**:

The syn-checkbox is a form field used to toggle a boolean state (+ indeterminate state).

## DoD

- [ ] The component looks like the Figma design and reflects all its states
- [ ] User toggle a state: on / off
- [ ] use Form support, Slots, Events, Methods, Parts from Shoelace
- [ ] create a proper documentation which contains at least everything from the Shoelace component
- [ ] Text in the documentation can be synced

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds support for the `<syn-checkbox />` component. It includes:

- The vendored component
- Storybook stories
- Integration Tests
- The generated angular and react wrappers
- Adjustments for the angular and react demo apps

### 🎫 Issues

Closes #72

## ✅ Checklist


mariohamann (approved):

packages/docs/stories/components/checkbox.stories.ts
```diff
@@ -0,0 +1,140 @@
...
+import { waitUntil } from '@open-wc/testing-helpers';
+import docsTokens from '../../../tokens/src/figma-tokens/_docs.json';
+import { storybookDefaults, storybookHelpers, storybookTemplate } from '../../src/helpers/component.js';
+const { args, argTypes } = storybookDefaults('syn-checkbox');
+const { overrideArgs } = storybookHelpers('syn-checkbox');
+const { generateTemplate } = storybookTemplate('syn-checkbox');
+
+const meta: Meta = {
+  component: 'checkbox',
+  args,
```
mariohamann, line 18:
```suggestion
    args: overrideArgs(
    [
      { name: 'default', type: 'slot', value: 'Default' }
    ],
    args
  ),
```

It would be good to show an example.

mariohamann (changes requested):

packages/docs/stories/components/checkbox.stories.ts
```diff
@@ -0,0 +1,140 @@
...
+      }
+    } catch (error) {
+      console.error('Error in play function:', error);
+    }
+  },
+  render: () => html`
+    <form class="custom-validity">
+      <syn-checkbox>Check me</syn-checkbox>
+      <br />
+      <button size="medium" type="submit">Submit</button>
```
mariohamann, line 101:
```suggestion
      <syn-button size="medium" type="submit">Submit</syn-button>
```

I would use the syn-button here.

packages/docs/stories/components/checkbox.stories.ts
```diff
@@ -0,0 +1,140 @@
...
+    </form>
+    <style>
+    form {
+      display: flex;
+      flex-direction: column;
+    }
+
+    button {
+      margin-top: 1rem;
+      align-self: flex-end;
```
mariohamann, line 111:
```suggestion
      align-self: flex-start;
```

As the storybook frame could become reeeeelly big, it makes sense to move this to the beginning.

packages/docs/stories/components/checkbox.stories.ts
```diff
@@ -0,0 +1,140 @@
...
+        story: docsTokens?.components?.['checkbox']?.default?.description?.value ?? 'No Description',
+      }
+    }
+  }
+} as Story;
+
+
+/**
+ * Use the checked attribute to activate the checkbox.
+ */
```
mariohamann, line 49: Please replace this with the stuff coming from Figma: `docsTokens?.components?.['checkbox']?.checked?.description?.value ?? 'No Description',`

etc.

Same for all other stories where the content is declared.

--

In other components (I believe e. g. in the button) we created a single function at the beginning of the story, to make this a  bit more fluent to write.

packages/docs/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "target": "ESNext",
     "useDefineForClassFields": false,
     "experimentalDecorators": true,
-    "module": "ESNext",
+    "module": "NodeNext",
```
mariohamann, line 10: Does this make sense? Storybook should work in the browser in the end, Node might be the wrong target?

packages/docs/stories/components/checkbox.stories.ts
schilchSICKAG: This seems to be harder to do. I would like to create a new ticket out of this as it will also have to be adjusted for the input stories. I did not get it to work correctly in the first place. I have a small hint where this is coming from, but would also need to update some internal packages for it to work (e.g. dom-testing-library, openwc-testing and others). Seems like a bigger problem to me.

packages/docs/stories/components/checkbox.stories.ts
schilchSICKAG: There is currently only the default, it it is not namespaced like when using button, see https://github.com/SickDesignSystem/synergy/blob/feat/72-syn-checkbox/packages/tokens/src/figma-tokens/_docs.json#L215. I think @LeaLeRose would have to add more doc tokens to make this work.

packages/docs/stories/components/checkbox.stories.ts
```diff
@@ -0,0 +1,146 @@
...
+export const CustomValidity: Story = {
+  play: async ({ canvasElement }) => {
+    try {
+      const button = canvasElement.querySelector('button');
+
+      await waitUntil(() => customElements.whenDefined('syn-checkbox'));
+
+      if (button) {
+        await userEvent.click(button);
+      }
```
mariohamann, line 98: I think it's a ShadowDOM problem. I would either try:
- `button.click()` CleanShot 2023-11-22 at 11 08 14@2x
- or jump into the ShadowDOM to get the button inside and make `userEvent.click(shadowedButton)`

schilchSICKAG: I tried it before reverting to the native button solution, did not work. I think I had seen something like this as a shoelace issue, but did not find it yet.

mariohamann (dismissed):

packages/docs/stories/components/checkbox.stories.ts
schilchSICKAG: Done in latest commit.

mariohamann (dismissed):

reichst1SICKAG (approved):
Hoover state for the complete component now looks good.

PR #177, opened 2023-11-23
docs: 📚 add and adjust missing docs for icon-button

LeaLeRose:


schilchSICKAG (changes requested):
As we have talked about this yesterday, maybe we should also add documentation for the `label` prop. This is the way it is currently defined in code:

```typescript
  /**
   * A description that gets read by assistive devices. For optimal accessibility, you should always include a label
   * that describes what the icon button does.
   */
  @property() label = '';
```

Don´t know if this is helpful for you on design side, but for dev we will need this token, too.

mariohamann:
@schilchSICKAG Let's do that when we develop the component then dev could make a proposal. Fine for you?

schilchSICKAG:
> @schilchSICKAG Let's do that when we develop the component then dev could make a proposal. Fine for you?

Sure, no problem from my point of view

schilchSICKAG (approved):

PR #179, opened 2023-11-24
docs: add syn-icon variant

LeaLeRose:


Gwenevere (dismissed):

packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -434,6 +434,16 @@
...
+      "variant": {
+        "title": {
+          "value": "Variants",
+          "type": "text"
+        },
+        "description": {
+          "value": "The Icon button can have two color variants, Primary or Neutral Color.",
+          "type": "text"
+        }
       }
```
mariohamann, line 447: Oh wait a second. I thought you're just showing how to override colors in regard to these two colors. So the plan is to provide those two colors and we have to reflect them in dev as a property?

Then please name it `color` not variant, so we can have the same as in `syn-button`.

LeaLeRose: I changed it quickly (Y)

mariohamann (dismissed):

mariohamann (approved):

LeaLeRose:
This Branch we should delete. It is somehow destroying our syn-radio docs tokens. I will make a new branch and hope this is not happening again @mariohamann

PR #180, opened 2023-11-24
fix: use correct GitHub link

mariohamann:
# Pull Request

## 📖 Description

This adresses needs raised in https://github.com/SickDesignSystem/Internal/issues/45#issuecomment-1825512354 to make the issue closable. This should resolve all hard coded links.

### DIFF
--- a/packages/angular/README.md
+++ b/packages/angular/README.md
@@ -31,7 +31,9 @@ Include either light or dark theme in your application, for example in a newly i
       "architect": {
         "build": {
           "options": {
-            "styles": ["@synergy-design-system/tokens/themes/light.css"]
+            "styles": [
+              "@synergy-design-system/tokens/themes/light.css"
+            ]
           }
         }
       }
@@ -57,8 +59,13 @@ import { AppComponent } from "./app.component";
 import { SynergyComponentsModule } from "@synergy-design-system/angular";
 
 @NgModule({
-  declarations: [AppComponent],
-  imports: [BrowserModule, SynergyComponentsModule],
+  declarations: [
+    AppComponent,
+  ],
+  imports: [
+    BrowserModule,
+    SynergyComponentsModule,
+  ],
   providers: [],
   bootstrap: [AppComponent],
 })
--- a/packages/angular/README.md
+++ b/packages/angular/README.md
@@ -1,6 +1,6 @@
 # @synergy-design-system/angular
 
-This package provides [Angular](https://angular.io/) wrappers for [Synergy Web Components](https://github.com/SickDesignSystem/synergy/tree/main/packages/components).
+This package provides [Angular](https://angular.io/) wrappers for [Synergy Web Components](https://github.com/synergy-design-system/synergy-design-system/tree/main/packages/components).
 
 > We are currently supporting Angular version 17.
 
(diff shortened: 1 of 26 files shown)


packages/angular/README.md
```diff
@@ -31,9 +31,7 @@ Include either light or dark theme in your application, for example in a newly i
       "architect": {
         "build": {
           "options": {
-            "styles": [
-              "@synergy-design-system/tokens/themes/light.css"
-            ]
+            "styles": ["@synergy-design-system/tokens/themes/light.css"]
```
schilchSICKAG, line 34: I would leave it in the more verbose way in the docs, as it is easier for copy&paste when the setting is already there.

packages/angular/README.md
```diff
@@ -53,19 +51,14 @@ This library is providing an `NgModule` named `SynergyModule`, which takes care
...
-import { BrowserModule } from '@angular/platform-browser';
-import { AppComponent } from './app.component';
-import { SynergyComponentsModule } from '@synergy-design-system/angular';
+import { NgModule } from "@angular/core";
+import { BrowserModule } from "@angular/platform-browser";
+import { AppComponent } from "./app.component";
+import { SynergyComponentsModule } from "@synergy-design-system/angular";
 
 @NgModule({
-  declarations: [
```
schilchSICKAG, line 62: I would leave the formatting here alone, as it makes it easier to copy lines and see the structure of the config.

packages/angular/README.md
```diff
@@ -53,19 +51,14 @@ This library is providing an `NgModule` named `SynergyModule`, which takes care
...
-    AppComponent,
-  ],
-  imports: [
-    BrowserModule,
-    SynergyComponentsModule,
-  ],
+  declarations: [AppComponent],
+  imports: [BrowserModule, SynergyComponentsModule],
```
schilchSICKAG, line 61: Same for here, all Angular Projects come with an imports section. The multi line syntax makes it easier to see and copy the needed string, so I was opting for this to be used.

schilchSICKAG (changes requested):

packages/angular/README.md
mariohamann: We need some global linting rules for Markdown, YAML, JSON etc. 

Apparently our local rules clash sometimes with each other. I don't see this as part of this ticket, but as something coming up several times now.

Suggestion: Prettier for all those files from root and done.

schilchSICKAG: I personally do not use code-formatting /beautifying at all because of such issues. Also, it may make some sense for denser syntax when really working with code, but when working with documentation I want extensive and easy to read (if not to say exhaustive) documentation for the parts that matter. When using automatic formatting, you take this choice away to the tool that you are using.

We have to think about how to handle this later on.

mariohamann: I don't think we should lose any time on discussing code formatting. I'm fine with any kind of code formatting as long as it's in any way doable automatically. Take Prettier or whatever. But we need coded rules. As long as there aren't any in place, we can't expect any one in the team to handle it according to personal preferences. And discussing linebreaks, spacings etc. is cumbersome and always personal – that's why I use automatic tools for that.

I will open a ticket for that.

packages/angular/README.md
mariohamann: Done.

packages/angular/README.md
mariohamann: Done: https://github.com/SickDesignSystem/synergy/issues/181

I resolved your issues to get this PR merged, but please let's address this ASAP.

schilchSICKAG (approved):
Seems good to me

PR #186, opened 2023-11-28
fix: 🤔 Storybooks play function does not work with custom component in forms

### LINKED ISSUE #169: fix: 🤔 Storybooks play function does not work with custom component in forms
# Checklist

# Version

The bug occurred with the following version of the component library: all

# Component

The issue arises when using the following component(s): checkbox, input

# Environment

# Description

As a developer of the Synergy Design System, I want to write interactive tests using the components. However, forms do not submit when there is no native `<button />` tag in the form. `<syn-button />` is not able to submit a form when used in a `play` function.

# Steps to reproduce

- Create a new branch from main
- Run pnpm i -r && pnpm build
- Start Storybook at packages/docs
- Open Storybook and move to the checkbox stories
- Open the story "Custom Validity"
- It will show an error message for the checkbox
- Exchange the `<button />` in the story with a `<syn-button />`
- Adjust the querySelector-Statement to target the `<syn-button />`.
- Notice that nothing is happening in the interactive test, but you will get the validation message when clicking the submit button.

Expected outcome:

It should still call the interactive test no matter if we use a native button or a `<syn-button />`

Actual outcome:

Nothing happens in the interactive test anymore.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue that `syn-button` could not be used to submit forms when using Storybooks play functionality.

### 🎫 Issues

Closes #169

## 👩‍💻 Reviewer Notes

- The event utilities of `@testing-library` do not play well with `customElements`. To make it work, I had to rely on the native DOM events (e.g. `button.click()` instead of `userEvent.click(button)`.
- For the sake of portability, I have added both functionalities to make Storybook recognize that a Play-Interaction was available at all.

## 📑 Test Plan

- Run `pnpm i -r && pnpm build` in the root of this branch
- Move to `packages/docs` and start Storybook
- Open http://localhost:6006/?path=/story/components-syn-input--invalid and check if the Required-Error-Message is displayed
- The form should include a `syn-button` with the content **Submit**.
- `Interactions` should show a **1** marker, as there was one automatic story (the `syn-button` was clicked)
- When resetting the interaction, it should show the form as valid
- When running the interaction again, the Error-Message should be shown again.

## ✅ DoD

### DIFF
--- a/packages/docs/stories/components/checkbox.stories.ts
+++ b/packages/docs/stories/components/checkbox.stories.ts
@@ -1,7 +1,5 @@
-/* eslint-disable */
-
-/* eslint-disable import/no-relative-packages */
-
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import type { SynButton, SynCheckbox } from '@synergy-design-system/components';
 import '../../../components/src/components/checkbox/checkbox';
 import type { Meta, StoryObj } from '@storybook/web-components';
 import { html } from 'lit';
@@ -9,6 +7,7 @@ import { userEvent } from '@storybook/testing-library';
 import { waitUntil } from '@open-wc/testing-helpers';
 import docsTokens from '../../../tokens/src/figma-tokens/_docs.json';
 import { storybookDefaults, storybookHelpers, storybookTemplate } from '../../src/helpers/component.js';
+
 const { args, argTypes } = storybookDefaults('syn-checkbox');
 const { overrideArgs } = storybookHelpers('syn-checkbox');
 const { generateTemplate } = storybookTemplate('syn-checkbox');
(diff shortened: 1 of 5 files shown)


kirchsuSICKAG (changes requested):

kirchsuSICKAG (approved):

packages/docs/stories/components/checkbox.stories.ts
```diff
@@ -17,88 +16,98 @@ const generateStoryDescription = (attributeName: string) => {
...
+      await waitUntil(() => Promise.allSettled([
+        customElements.whenDefined('syn-checkbox'),
+        customElements.whenDefined('syn-button'),
+      ]));
 
-      if (button) {
+      if (form && button && checkbox) {
+        checkbox.setCustomValidity(errorMessage);
         await userEvent.click(button);
+        button.click();
```
mariohamann, line 110: You're doing two clicks here, one via userEvent and one via button.click() – is this needed?

packages/docs/stories/components/input.stories.ts
```diff
@@ -156,42 +155,51 @@ export const Invalid: Story = {
...
+      }
+
+      const input = form.querySelector('syn-input') as SynInput;
+      const button = form.querySelector('syn-button') as SynButton;
 
-      await waitUntil(() => input?.shadowRoot?.querySelector('input'));
+      await waitUntil(() => Promise.allSettled([
+        customElements.whenDefined('syn-input'),
+        customElements.whenDefined('syn-button'),
+      ]));
```
mariohamann, line 170: We had some issues in Solid with whenDefined or similar because sometimes the element wasn't completely rendered. But yah, maybe it works now, let's see.

packages/docs/stories/components/checkbox.stories.ts
schilchSICKAG: Yes, this is intended. The first click will make sure you are able to use Storybooks "Interactions", e.g. replay and stuff. The second one is really doing the work.

Problem is that a) `userEvent.click` does not work on web-components in play functions, but is needed for Storybook to work and b) `Element.click` does what it should, but does not notify Storybook :(

packages/docs/stories/components/input.stories.ts
schilchSICKAG: Interesting, I can see this in Chromatic, but not on Storybook itself. Will have to check why the Screenshots look scrambled.

packages/docs/stories/components/checkbox.stories.ts
mariohamann: Understand. Maybe you could write some inline docs do make that more understandable.

schilchSICKAG: Maybe I could also create a shared wrapper function so we don not have to repeat it all over? Semantics would be the original ones from TestingUtils, would look like:

```javascript
const triggerClick = async (elm: HTMLElement, ...args?: unknown[]) => {
  await testUtils.click(elm, ...args);
  elm.click();
}
```

What do you think?

mariohamann: Hm. I think that sometimes abstractions could lead to confusion. ;)

schilchSICKAG: Right, I also don´t know how much steps we will have where this becomes a problem. Seems testUtils fire events differently and lit is not able to see them as there are missing infos. Regular html elements do not seem to have this problem. Maybe this goes away when updating the tooling. Did not find an issue for this, through :(

packages/docs/stories/components/input.stories.ts
mariohamann: What's your plan on this? As this ticket would be important to show our real buttons (design already complained) I think we should stick to the initial task and remove these changes to get it done.

schilchSICKAG: Just pushed a new version of this branch where I also updated all components to use syn-button in the play functions. The calls to waitUntil are removed there as well, seems to work fine (at least locally)

Issue #150, opened 2023-11-16
docs: 📚 add new Synergy branding to Storybook and GitHub and everywhere


mariohamann:
# Description

Every place where we communicate Synergy outside, we should present the new logo, e. g. as favicon, logo etc.

# Rationale

We want to make our brand Synergy stronger. :)

# DoD

- [x] Create Artboards with safe space around the logo to export them appropriately
- [x] Find a place in Figma where the logo is maintained (maybe the Helper Library?)
- [x] Export svg, png of logo + icon into #internal repo
- [x] Github icon
- [x] Storybook icon
- [x] Figma workspace / project icon
- [x] Figma file cover template
- [x] Intranet pages
- [x] Comm Channels

eileenwegner:
**Favicon**
![Image](https://github.com/SickDesignSystem/synergy/assets/148873173/84cdd52a-5902-4538-b8a1-a13318958057)


**Profile Picture**
![Image](https://github.com/SickDesignSystem/synergy/assets/148873173/33a426a0-4aec-4107-9c72-68c90c77c51f)

reichst1SICKAG:
Thanks @eileenwegner 
I'll export some versions into #internal to have a 'stable' released version there to use, download or reference.

PR #201, opened 2023-12-01
docs: 📚revise general documentation and licensing

klaarseSICKAG:
# Pull Request

## 📖 Description

- revised doc
- revised licensing
- added acknoledgement to main 3rd party dependencies
- added Synergy logo to readme and storybook

### 🎫 Issues

- https://github.com/SickDesignSystem/Internal/issues/37
- #150


packages/components/package.json
schilchSICKAG, line 1: Use MIT here

packages/docs/NOTICE
schilchSICKAG, line 1: Why is LICENSE removed but NOTICE added? Wouldn´t it make more sense to juse use MIT like everywhere else?

packages/docs/package.json
schilchSICKAG, line 1: As above, does not make sense here. Suggestion: Use the SEE LICENSE IN LICENSE statement where it differs from MIT and use MIT on all other occurences.

packages/react/README.md
schilchSICKAG, line 1: This package does offer components. There is nothing like form-support in React and surely not two way data-binding (React is all about unidirectional dataflow).

Suggestion: Leave this out all together or reduce to:

- Autocompletion
- Event handling
- Typescript types (which would fit better to Auto-Completion imho, as ts will provide type hints and autocomplete).

packages/react/package.json
schilchSICKAG, line 1: See comment defails above. Noone will install a package that has such a license. And the packages don´t need it in the first place.

packages/tokens/README.md
```diff
@@ -1,7 +1,7 @@
 # @synergy-design-system/tokens
 
 This package provides the design tokens that form the base for all components of the synergy design system.
-It provides multiple exports that can be consumed in applications.
+It provides multiple exports for colors, spacings, shadows, sizings, etc. that can be also consumed in other applications.
```
schilchSICKAG, line 4: We do not offer applications, so the other may be misleading here.

packages/tokens/package.json
schilchSICKAG, line 1: See comments on licenses below

burktiSICKAG (approved):

PR #211, opened 2023-12-05
feat: ✨ add syn-icon-button

### LINKED ISSUE #167: feat: ✨ create icon-button
# Description

The icon button is an icon which behaves like a button.

# Acceptance criteria

- [ ] The component looks like the Figma design and reflects all its states
- [ ] Create a story which contains at least everything from the Shoelace component
- [ ] Add documentation for the label (a11y)
- [ ] Enforce setting the label (via css)
- [ ] Add a size property

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds support for the <syn-icon-button /> component. It includes:

- The vendored component
- Storybook stories
- The generated angular, react and vue wrappers

### 🎫 Issues

Closes #167 

## ✅ DoD


schilchSICKAG (approved):

schilchSICKAG (dismissed):

schilchSICKAG (dismissed):

mariohamann (changes requested):
Would really favor having the size inheritance as a default, so that per default the size is set via current font size (as it's done with the icons).

1. One principle we follow is "prepare for the unknown". By forcing to the sizes, we violate this principle.
2. We introduced the sizes in design to make the usage more convenient and systemized there. Still in Figma it's super easy to just change the size manually – and this is fine. This change in Figma was not meant to force to these sizes, but to make the usage easier. My understanding of this adoption would be: We offer the sizes, without forcing to them. If this now leads to problems, I would more go the route of removing the size in design and instead point to the spacing scale and show how to resize the element. Maybe that really would make everything easier. @LeaLeRose
3. Or proposed compromise would be: Provide the sizes, make clear on both sides, that's it's resizable to individual needs. And stick to size inheritance per default on dev side. ;)

reichst1SICKAG:
> Would really favor having the size inheritance as a default, so that per default the size is set via current font size (as it's done with the icons).
> 
>     1. One principle we follow is "prepare for the unknown". By forcing to the sizes, we violate this principle.
> 
>     2. We introduced the sizes in design to make the usage more convenient and systemized there. Still in Figma it's super easy to just change the size manually – and this is fine. This change in Figma was not meant to force to these sizes, but to make the usage easier. My understanding of this adoption would be: We offer the sizes, without forcing to them. If this now leads to problems, I would more go the route of removing the size in design and instead point to the spacing scale and show how to resize the element. Maybe that really would make everything easier. @LeaLeRose
> 
>     3. Or proposed compromise would be: Provide the sizes, make clear on both sides, that's it's resizable to individual needs. And stick to size inheritance per default on dev side. ;)

I assume you only refer to the syn-icon-button (and the syn-icon) with this flexible sizing. 

Nevertheless I would prefer option 3, so that the sizes initially match the other elements and they can be easily inserted with them, but can be adapted "if the icon just doesn't look good". 

Most of the users will just use the provided sizes and be happy if the can sometimes adapt them.

kirchsuSICKAG:
@mariohamann  @reichst1SICKAG  @LeaLeRose 
But one question will then be open: 
Which padding should the inherited icon-button get? Because every size has it`s own padding:
small -> 12px
medium -> 8px
large -> 16px
inherit -> ??

mariohamann:
Oh. The icon-buton has several paddings now? That could be a problem. @LeaLeRose Do we need different paddings in design? The shoelace one just has everywhere the same padding:

mariohamann (dismissed):

packages/components/package.json
```diff
@@ -169,9 +169,9 @@
...
   "meta": {
     "checksums": {
-      "angular": "9760e508e700fb38c86cc04d2a27fd0687b9afbd",
-      "react": "4564414376bb48f54c847326d862303f28366540",
-      "vue": "260b2e7d31d22a7f19c43ce42af51769a03fdf9f"
+      "angular": "c401af9982a050f49e6e83af0be303b4300d2ad8",
+      "react": "50a2cf23a2410d11a95b38eebf2d00ed1a4b4dc0",
+      "vue": "a91c253967859e9d215cbc255834feba738bde6c"
     }
   }
```
mariohamann, line 176: Remember to remove this. :)

mariohamann (approved):

Issue #181, opened 2023-11-24
chore: 🔧 setup code formatting rules


mariohamann:
# Rationale
As a maintainer of Synergy Design System I want to be able to write code without approaching the team how to format it.

# Description
Apparently we have different formatting settings and different settings on auto save which leads to problems in PRs: https://github.com/SickDesignSystem/synergy/pull/180#discussion_r1404268440

We need neutral ways to format code, to not lose any time on discussing line breaks, spacings etc.

Code formatting has therefore be able to be done automatically, so everyone can decide to do a `pnpm format.fix` or use `auto-save` including formatting or whatever.

A check has to be in place which checks every PR whether the formatting is correct.

mariohamann:
cc @schilchSICKAG

PR #241, opened 2024-01-09
chore: fix release process

mariohamann:
# Pull Request

## 📖 Description

Instead of building workarounds, we included a function in `semantic-release-monorepo` that allows it to check other folders to be relevant for a release. The change was included as a patch via `pnpm`.

## ✅ DoD


schilchSICKAG:
Unfortunately, I get an error when running `pnpm install -r`. The CI is also complaining that checksums are not matching. Any Idea where this may come from:

packages/components/package.json
```diff
@@ -23,9 +23,7 @@
     "./utilities/*": "./dist/utilities/*",
     "./translations/*": "./dist/translations/*"
   },
-  "files": [
-    "dist"
-  ],
+  "files": ["dist"],
```
schilchSICKAG, line 26: The beautifier again :). Will make sure to tackle #181 in the next couple of days :)

packages/components/package.json
```diff
@@ -40,7 +38,8 @@
     "release": "semantic-release  --tagFormat 'components/${version}' -e semantic-release-monorepo",
     "release.angular": "git add ../angular && git commit -m \"chore(release/angular) via component release [skip actions]\"",
     "release.react": "git add ../react && git commit -m \"chore(release/react) via component release [skip actions]\"",
-    "release.vue": "git add ../vue && git commit -m \"chore(release/vue) via component release [skip actions]\""
+    "release.vue": "git add ../vue && git commit -m \"chore(release/vue) via component release [skip actions]\"",
+    "release.dry": "semantic-release --dry-run --tagFormat 'components/${version}' -e semantic-release-monorepo"
```
schilchSICKAG, line 42: Do you think it would make sense to add this as sort of a "semi verify-step" to `CONTRIBUTION.md`?

packages/components/package.json
```diff
@@ -145,19 +144,13 @@
...
-        "@semantic-release/github",
-        {
-          "successComment": false
+          "assets": ["CHANGELOG.md", "package.json"]
         }
       ]
-    ]
+    ],
+    "monorepo": {
+      "dependencies": ["packages/vue", "packages/react", "packages/angular"]
```
schilchSICKAG, line 152: Really like the ease of configuration here!

patches/semantic-release-monorepo@7.0.5.patch
schilchSICKAG, line 1: Would you mind to also commit your patch source file, not only the diff? Also big thumbs up for taking care of communication for https://github.com/pmowrer/semantic-release-monorepo/issues/142

schilchSICKAG (dismissed):
Nice work!

packages/components/package.json
mariohamann: Thanks!

patches/semantic-release-monorepo@7.0.5.patch
mariohamann: You can see the file when you do `pnpm patch semantic-release-monorepo` again. I hesitate on having two sources of truth. I know, that it's uglier to read, but I fear that things divide.

packages/components/package.json
mariohamann: Locally it breaks because of missing GitHub tokens, but on GitHub we could implement it into our Quality Gate. Will do that directly.

patches/semantic-release-monorepo@7.0.5.patch
schilchSICKAG: Just tried it out, perfect :)

schilchSICKAG (approved):

PR #250, opened 2024-01-12
docs: 📚 Add token README and Color Swatches to storybooks package

### LINKED ISSUE #218: docs: 📚 Add token README and Color Swatches to storybooks package
# Description

We are currently missing the token documentation in our storybook instance. We should add at least:

- [x] README.md and CHANGELOG.md as for the other packages.
- [x] A story that displays the use of colors. This can be archived by reviving [the old stories originally created for the first iteration](https://github.com/synergy-design-system/synergy-design-system/tree/main/packages/docs/stories/foundations/Colors) of the tokens package.

# Rationale

As a user of the Synergy Design System, I want to have easy access to the `@synergy-design-system/tokens` documentation to have a reference of all available colors.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds back available color swatches in storybook. It includes all tokens in dark and light mode and the ability to display and copy the values for the following variants of the tokens:

- Raw (e.g. css hex value)
- CSS Variable
- SCSS Variable
- JS Variable

### 🎫 Issues

Closes #218 

## 👩‍💻 Reviewer Notes

- Check out this branch
- After updating/building, move to `packages/docs` and start storybook
- Navigate to Tokens -> Colors and play around with it.
- Click on one of the table cells to copy the value.
- The dark mode is currently not working, as the branch for dark mode is not merged yet. Classes and definitions are fine however, so it should be easily be reintegrated after dark mode support has landed.

> The copy functionality is provided as a preview. I am aware that it does not give any status information about the obtained copy. Maybe we can later adapt it to use Storybooks messaging system.

## ✅ DoD


kirchsuSICKAG (commented):
Just one thing that I noticed:
For every color we have hex values except for **--syn-color-neutral-0: white;** 
and **--syn-color-neutral-1000: black;**
Do we want it like this or should we use hex values for everything (therefore we would need to update those tokens)?
Was just struggling over this on the first glance, because I though there is some error somewhere in how the raw value was created.
image

schilchSICKAG:
> Just one thing that I noticed: For every color we have hex values except for **--syn-color-neutral-0: white;** and **--syn-color-neutral-1000: black;** Do we want it like this or should we use hex values for everything (therefore we would need to update those tokens)?

Its drawn from the tokens where the value is set as string. It also appears like this in the Figma Documentation, so for me it is fine, through I find it kind of inconsequent, too. Maybe we should create a follow up ticket for @LeaLeRose?

As it will automatically update when design updates it, I do not think that this is a blocker.

kirchsuSICKAG:
> > Just one thing that I noticed: For every color we have hex values except for **--syn-color-neutral-0: white;** and **--syn-color-neutral-1000: black;** Do we want it like this or should we use hex values for everything (therefore we would need to update those tokens)?
> 
> Its drawn from the tokens where the value is set as string. It also appears like this in the Figma Documentation, so for me it is fine, through I find it kind of inconsequent, too. Maybe we should create a follow up ticket for @LeaLeRose?
> 
> As it will automatically update when design updates it, I do not think that this is a blocker.

Yes, saw that it comes from our tokens / figma. I don`t know if we directly need a ticket. Maybe it is enough for now pointing it out and discuss it in the group?

But I agree with you, that this PR can be merged for now

kirchsuSICKAG (approved):

PR #251, opened 2024-01-12
docs: 📚 Create showcase for form field usage in Storybook

### LINKED ISSUE #124: docs: 📚 Create showcase for form field usage in Storybook
# Stories

As a user of the Synergy Design System I want to have a demo which shows how to build html forms using Synergy components.

As a developer of the Synergy Design System I want to have a demo to make sure that the created form field behaves as expected.

As a developer of the Synergy Design System I want to have a demo in a separate package to validate that the SDS can be flawlessly used as an external library. 

# Acceptance critera

 - status of the form field is submitted via the form data when triggering the submit action
 - data binding is working
 - reference to SDS package is working
 - Figma example are used as reference

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the first version of a static form element, as seen on the corresponding Figma branch. It is intended to show how to build simple forms. The provided markup will be used for the framework demos to provide a coherent form demo for all supported frameworks.

### 🎫 Issues

Closes #124.

## 👩‍💻 Reviewer Notes

I have created two samples for the form, one is done in `mdx` and one in storybooks story format. Both have a set of advantages and drawbacks which should be discussed. I would opt for the regular story approach.

#### Advantages:

- You write (nearly) plain HTML in a single file, using react syntax.
- You can write your JS functionality inline

#### Disadvantages:

- Newlines will automatically insert a `<p />`, which is unwanted and leads to problems during formatting the code.
- There are no screenshots here.
- No possibilities to see the code in Storybook.
- There will be no screenshots captured currently (this may be configurable, I did not have a look at it).
- Code Formatting, AutoComplete and Linting not possible.

---

#### Advantages:

- Code Preview available via `Docs`.
- Formatting is up to you completely, no messy custom tags inserted.
- Proper syntax highlighting and type checking.

#### Disadvantages:

- Displays unneeded information (e.g. argTypes in Docs page)
- Unable to hide the "default" export from sidebar, which leads to duplicated content (Docs and Default Story)

---

Also, as this is the recommended base we want others to adhere to, we should double check accessibility and formatting issues (e.g. css adjustments). Would look forward to get some feedback, especially from @mariohamann.

I have currently opted to add the text in a static manner, this should be done via docs tokens (and maybe also exchanged for english when we are at it) later on. This can be done when @LeaLeRose starts working on the Figma template.

## 📑 Test Plan

- Check out the branch
- Issue `pnpm i -r && pnpm build`
- Start up Storybook in `packages/docs`
- Inspect the two versions at the new `Templates` navigation at the bottom.

## ✅ DoD


klaarseSICKAG:
Nice work! Just a few remarks:
- I think we should provide it better in english although the template is in german
- Do we need 3 pages with the same contents? (Docs, Default, Forms) ?

schilchSICKAG:
> Nice work! Just a few remarks:
> 
> * I think we should provide it better in english although the template is in german

You are right. However, as stated in the PR description, this will have to be done via docs tokens when the Figma implementation gets done. I do not think that this is a thing that we should do twice (e.g. now with static strings in Storybook, then again with exchanging the static strings from the docs tokens). I would like to wait until the docs tokens are ready and implement it in one go. However, I could also translate it for myself and create a follow up ticket. You decide ;)

> * Do we need 3 pages with the same contents? (Docs, Default, Forms) ? Lets have a discussion about the pros and cons of mdx vs CSF

No, as stated in the description I have shown both ways of creating the story. As I have started with MDX story format and migrated it to a "regular" story, you will have all three of them currently, until we decide which one to use. I think the CSF (a.k.a "regular story) wins here, even when duplicating Docs and the Story. However, I want to have feedbacks from other devs about this before deciding which one we will use.

kirchsuSICKAG (dismissed):

packages/docs/stories/templates/form.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+    .synergy-form-demo {
+      background: var(--syn-color-neutral-0);
+      border: 1px dashed var(--syn-color-neutral-300);
+      margin: 0 auto;
+      padding: 10px 20px;
+      max-width: 750px;
+      min-width: 360px;
+    }
+
+    .synergy-form-demo h1 {
```
mariohamann, line 35: 1. What do you think about wrapping all templates in iFrames on the docs page, so you can get rid of all classes to scope e. g. `h1` –  here an example for implementation https://github.com/storybookjs/storybook/issues/13765#issuecomment-1538517811.
2. We implemented PostCSS to look into html`` template literals: https://github.com/solid-design-system/solid/commit/452ac1c4ba1f37f304d056c033541779227ca406#diff-56a2517c7d913d792b087e343de1d7e6e9e23e493608be111aa8ff69dd4bafe8 This made it possible to do SCSS nesting like e. g. here https://github.com/solid-design-system/solid/commit/452ac1c4ba1f37f304d056c033541779227ca406#diff-06a0efe585bedb897c7a0062a74ce5d5aa6a058e3406ecb973b2eddef1bf952aR145-R161

I think this could be interesting for us?

packages/docs/stories/templates/form.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+    /*
+     * The form control inputs are using a css grid to be displayed.
+     * We just let them flow automatically mobile and arrange them
+     * in two columns when space is wide enough
+     */
+    syn-radio-group::part(form-control-input) {
+      display: grid;
+    }
+
+    @media screen and (min-width: 641px) {
```
mariohamann, line 92: As one principle of us is `Progressive` I would suggest to go the Container Query Route, instead of looking on screen sizes – this makes especially sense for multi column layouts. Support ist quite fine and I think we should lead by example for Progressive Enhancement: https://caniuse.com/css-container-queries

In fact that's something where we have to align with Design on.

packages/docs/stories/templates/form.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+            <syn-input label="PLZ" name="zip" required></syn-input>
+            <syn-input label="Ort" name="city" required></syn-input>
+            <syn-input label="Land" name="country" required></syn-input>
+            <syn-input label="Ansprechpartner" name="salesPerson" required></syn-input>
+            <syn-input type="tel" label="Telefonnummer" name="phone"></syn-input>
+            <syn-input type="tel" label="Fax" name="fax"></syn-input>
+            <syn-input type="email" label="E-Mail-Adresse" name="mail" required></syn-input>
+          </div>
+          <syn-checkbox name="subscribeNewsletter">
+            Ja, ich möchte regelmäßig aktuelle und interessanteInformationen zu Lösungen mit Produkten, Systemenund Services von SICK per E-Mail erhalten.
```
mariohamann, line 151: Typos. But I'm not sure if this should already be fixed now or if we wait for the data from design. :)

packages/docs/stories/templates/form.stories.tsx
```diff
@@ -0,0 +1,182 @@
+import React from 'react';
```
mariohamann, line 1: 🎉

schilchSICKAG: :) Needed for the docs renderer unfortunately, as Storybooks config does not respect the  "Import not Needed" functionality of react@18 :(

packages/docs/stories/templates/form.stories.tsx
schilchSICKAG: I will have a look at the iframe solution. What do you think about using css nesting directly? We would not need to have to use a toolchain for that, as support seems pretty good nowadays: https://caniuse.com/css-nesting. Safari would be supported from MacOS Sonoma and i(pad) OS 17 onwards.

I just do not want to integrate more tooling than necessary.

packages/docs/stories/templates/form.stories.tsx
schilchSICKAG: I will try it out and adapt the story.

packages/docs/stories/templates/form.stories.tsx
schilchSICKAG: iFrame encapsulation works great, thanks for the tip! Unfortunately, you are only allowed to set the iframe to a fixed size, no auto sizing is supported :(. I have opted to set the height to the current forms height manually. Not an elegant solution, but I think this is fine for now.

mariohamann: > I just do not want to integrate more tooling than necessary.

That sounds very good.

> CSS Nesting

There are quite some drawbacks ATM. I fear that this might be tooooo progressive. 😆

I think the iFrame solution already improves readability a lot, therefore I think that's totally fine like that.

packages/docs/stories/templates/form.stories.tsx
```diff
@@ -0,0 +1,206 @@
...
+    </style>
+    <div class="synergy-form-demo">
+      <h1>${getTranslation('contactForm.headline')}</h1>
+      <p>${getTranslation('contactForm.subHeadline')}</p>
+
+      <form method="post" id="syn-form-demo">
+        
+        <fieldset>
+          <syn-radio-group
+            label="${getLabel('contactForm.topicLabel')}"
```
kirchsuSICKAG, line 133: Is there a specific reason why sometimes the `getTranslation` method and sometimes the `getLabel` method is used? 
I think it would be easier for others, if we keep the code as simple as possible and therefore only use one of them for translations.

schilchSICKAG: Reason for this was Typescript. When using `getTranslation`, you may either receive a string or an array of strings (for options). In this case, TS will yell at you as lits properties only accept strings. In this case you would always have to cast, e.g.

`label="${getLabel('contactForm.topicLabel')}"` will become:
`label="${(getLabel('contactForm.topicLabel') as string)}"` in all lines, makeing it kind of repetitive. This was the sole reason for this function.

schilchSICKAG: Update: Just had another look at it. Seems I was smart enough to circumvent this by updating `getTranslations`, but I did not remove `getLabel`. Will do that now :)

kirchsuSICKAG (dismissed):

PR #272, opened 2024-01-19
docs: 📚 add select divider option docs

LeaLeRose:
# Pull Request

## ✅ DoD


packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -851,13 +915,125 @@
...
+          "type": "text"
+        }
+      },
+      "initialvalue": {
+        "title": {
+          "value": "Setting Initial Values",
+          "type": "text"
+        },
+        "description": {
+          "value": "Use the value attribute to set the initial selection.\n\nWhen using multiple, the value attribute uses space-delimited values to select more than one option. Because of this, <sl-option> values cannot contain spaces. If you’re accessing the value property through Javascript, it will be an array.",
```
kirchsuSICKAG, line 994: `<sl-option>` should be exchanged with `<syn-option>`

```suggestion
          "value": "Use the value attribute to set the initial selection.\n\nWhen using multiple, the value attribute uses space-delimited values to select more than one option. Because of this, <syn-option> values cannot contain spaces. If you’re accessing the value property through Javascript, it will be an array.",
```

packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -851,13 +915,125 @@
...
+          "type": "text"
+        }
+      },
+      "gettag": {
+        "title": {
+          "value": "Custom Tags",
+          "type": "text"
+        },
+        "description": {
+          "value": "When multiple options can be selected, you can provide custom tags by passing a function to the getTag property. Your function can return a string of HTML, a Lit Template, or an HTMLElement. The getTag() function will be called for each option. The first argument is an <sl-option> element and the second argument is the tag’s index (its position in the tag list).\n\nRemember that custom tags are rendered in a shadow root. To style them, you can use the style attribute in your template or you can add your own parts and target them with the ::part() selector.",
```
kirchsuSICKAG, line 1034: `<sl-option>` should be exchanged with `<syn-option>`

```suggestion
          "value": "When multiple options can be selected, you can provide custom tags by passing a function to the getTag property. Your function can return a string of HTML, a Lit Template, or an HTMLElement. The getTag() function will be called for each option. The first argument is an <syn-option> element and the second argument is the tag’s index (its position in the tag list).\n\nRemember that custom tags are rendered in a shadow root. To style them, you can use the style attribute in your template or you can add your own parts and target them with the ::part() selector.",
```

kirchsuSICKAG (changes requested):

kirchsuSICKAG (dismissed):

kirchsuSICKAG (dismissed):

schilchSICKAG (approved):

PR #280, opened 2024-01-24
Add ability to run test.watch with automatic builds

### LINKED ISSUE #141: chore: 🔧 Adjust test runs so they use the latest version of the bundle for components
# Description

As discussed in https://github.com/SickDesignSystem/synergy/pull/130#discussion_r1393812705, we should provide ways for creating the distribution files automatically when running tests in `@synergy-design-system/components`. There may also be issues when using `web-test-runner` in watch mode, because the new bundle would need be created on another shell.

# Proposed solution:

- Find a way to automatically create the bundle when spinning up the web-test-runner
- When running in watch mode, web-test-runner should rebuild the bundle
- Make sure to NOT create a complete build (e.g. including angular and react wrappers), but just the components.

# Rationale

The current process is error prone and not documented, which might lead to problems (e.g. forgetting to build before testing leads to false positives).

# Acceptance criteria

- [ ] tests are running locally on latest components
- [ ] test watch mode should trigger rebuilds

schilchSICKAG:
# Pull Request

This PR adds the ability to use the tests for components watch mode. A new command `pnpm test.watch` was implemented for this to work.

## 📖 Description

Until now, we did not provide watch mode for tests, making it necessary to build the components and run the tests again. This is now automatically included with the new watch mode.

### 🎫 Issues

Closes #141

## 👩‍💻 Reviewer Notes

- I had to create a simple plugin for the test server to make our changes available via chokidar.
- As I do not have access to the original bound listener, tests will run twice when changing a component. There was no simple way to avoid that unfortunately.
- Changes to the test files themselves will NOT trigger a rebuild.

## 📑 Test Plan

- Check out the branch and prepare it via `pnpm i -r && pnpm build`
- Move to `packages/components` and start the test server via `pnpm test.watch`.
- Change a test and validate that it breaks (e.g. changing an assertion in `components/button/button.custom.tests.ts`
- Change a default attribute of the button to some string or just hit save in `button.component.ts` (Make sure to remove it from the readonly files first!).
- The build will now recreate the bundle and run the tests again. The tests will echo out one to multiple 404 errors, which happen because the new listener and the original one run concurrently. After another run, the system should be back to normal.

## ✅ DoD

### DIFF
--- a/packages/components/scripts/tests/index.js
+++ b/packages/components/scripts/tests/index.js
@@ -1,0 +1,64 @@
+import { execSync } from 'child_process';
+import * as jobs from '../jobs/index.js';
+import { getPackageJSONAsObject, getPath } from '../jobs/shared.js';
+
+const componentDir = getPath('../');
+const outDir = getPath('../dist');
+const temporaryOutDir = getPath('../dist.testrunner');
+
+const { version } = getPackageJSONAsObject();
+
+// eslint-disable-next-line no-underscore-dangle
+const __PACKAGE_VERSION__ = JSON.stringify(version.toString());
+
+const synRefreshComponentBundle = () => {
+  const name = 'syn-refresh-component-bundle';
+
+  const rebuildSynergy = async () => {
+    // Create the build in its own directory
+    await jobs.runPrepare(temporaryOutDir);
+    await jobs.runCreateEvents(componentDir);
+    await jobs.runCreateExports(componentDir);
+    await jobs.runTypeScript(temporaryOutDir, './tsconfig.prod.json');
+    await jobs.runEsBuildComponents(temporaryOutDir, __PACKAGE_VERSION__);
+
+    // Finally, remove the dist folder completely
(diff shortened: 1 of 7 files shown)


packages/components/scripts/tests/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+  return {
+    name,
+    serverStart: async ({ fileWatcher }) => {
+      // Make sure we always have correct initial data by running the build
+      await rebuildSynergy();
+
+      // Whenever one of our component files has changed,
+      // make sure to recreate our bundle
+      fileWatcher.add('./src/components/**/!(*.(style|test)).ts');
```
mariohamann, line 39: I would include styles for the tests. If stuff like the visibility has to be tested, this would be part of CSS, shouldn't it?

schilchSICKAG: Fixed in latest version

mariohamann (approved):

PR #286, opened 2024-01-26
fix: 🤔 SCSS tokens cannot be loaded via postcss because of defunct exports map

### LINKED ISSUE #282: fix: 🤔 SCSS tokens cannot be loaded via postcss because of defunct exports map

# Version

The bug occurred with the following version of the component library: latest

# Component

The issue arises when using the following component(s): all

# Environment

# Description

When trying to include the scss tokens in a postcss build pipeline, postcss breaks with an error that states the tokens.scss file cannot be found. This is because the exports map does not state the correct path to `dist/scss/_tokens.scss`.

# Acceptance criteria

# Reproduction

- [ ] Create an empty test project
- [ ] Install postcss, @synergy-design-system/tokens and scss
- [ ] Create an empty css file and try to load the scss tokens.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue with the exports map that prevents scss files from being loaded correctly. It also adjusts the documentation for sass especially to clarify the import paths for scss files.

### 🎫 Issues

Closes #282 

## 👩‍💻 Reviewer Notes

Please have a look at the updated documentation.

## ✅ DoD


packages/tokens/README.md
```diff
@@ -135,14 +135,17 @@ console.log(div.style.backgroundColor);
...
 
+> Note that the SASS compiler does not ship with built in support for `node_modules`. You will either have to use relative paths or use a loader like [node-sass-package-importer](https://www.npmjs.com/package/node-sass-package-importer) or [sass-module-importer](https://www.npmjs.com/package/sass-module-importer).
+> Configuration for this is NOT part of this guide.
+
 ```scss
 // Import the design tokens first.
 // This can be done in a sass file or in any other way described above.
 // Make sure to NOT add the .css file suffix, this will not work in sass
-@import "~@synergy-design-system/tokens/themes/light";
+@import "./node_modules/@synergy-design-system/tokens/themes/light";
```
kirchsuSICKAG, line 145: I tested it in a vite project. 
image

So it seems on relative imports we need the **dist** in the import path, otherwise not.
Therefore we should update the import here to also have the **dist** in the path

packages/tokens/README.md
```diff
@@ -135,14 +135,17 @@ console.log(div.style.backgroundColor);
...
 
 // Import the scss variables
-@import "@synergy-design-system/tokens/scss/tokens";
+@import "./node_modules/@synergy-design-system/tokens/dist/scss/tokens";
```
kirchsuSICKAG, line 148: Maybe we should add documentation for both import ways. For relative import and the other one. Because here again on relative import the **dist** is needed and on non-relative imports there must be **no dist**.

image

kirchsuSICKAG (changes requested):

packages/tokens/README.md
schilchSICKAG: Yes, I think this is caused by the fact vite and its plugins honor the `exports` statement in `package.json`, but other bundlers ignore it (like webpack). On relative paths it ignores `package.json` completely, so maybe it is more save to use the `dist` folder always for scss before messing different paths up for each separate bundler. What do you think?

schilchSICKAG: Resolved in latest commit

packages/tokens/README.md
schilchSICKAG: I have opted to remove the references to the non relative ways of importing completely. This was done as there are multiple loaders for sass in the various build pipelines with different configurations. There seems to be no "one fits it all" approach there. Please have a look at the updated documentation and provide feedback about it.

kirchsuSICKAG: I like the updated documentation

kirchsuSICKAG (approved):

PR #307, opened 2024-02-05
feat: ✨ Improve events export in framework wrapper

### LINKED ISSUE #303: feat: ✨ improve events export of framework wrapper

# Component

I suggest to add the feature to the following component: all three framework wrapper

# Story

As a maintainer of the Synergy Design System, I would like to export the events over the event.ts file of the `@synergy-design-system/components` directly in each wrapper instead of exporting every used event in each component so that the maintainability of the wrapper builder is easier.

# Acceptance criteria

- [ ] no event is exported anymore in a component itself

kirchsuSICKAG:
# Pull Request

## 📖 Description

To make the framework wrapper better maintainable, the exports of the components events will be unified and done in the index.ts 

### 🎫 Issues

Closes #303 

## ✅ DoD


packages/_private/angular-demo/src/demoform/demoform.component.ts
```diff
@@ -1,4 +1,4 @@
-import type { SynChangeEvent } from '@synergy-design-system/components';
+import type { SynChangeEvent } from '@synergy-design-system/angular'
```
schilchSICKAG, line 1: Much cleaner, like this!

packages/_private/angular-demo/src/themeswitch/themeswitch.component.ts
```diff
@@ -1,8 +1,6 @@
 import { Component } from '@angular/core';
-import type {
-  SynChangeEvent,
-  SynSwitch as SynSwitchElement,
-} from '@synergy-design-system/components';
+import type { SynChangeEvent } from '@synergy-design-system/angular';
+import type { SynSwitch as SynSwitchElement } from '@synergy-design-system/components';
```
schilchSICKAG, line 3: Here a problem becomes apparent: We are exporting all types from `@synergy-design-system/components` via `export type *...`.

packages/_private/react-demo/src/ThemeSwitch.tsx
```diff
@@ -1,9 +1,9 @@
 import { useState } from 'react';
-import { SynSwitch } from '@synergy-design-system/react';
-import type {
```
schilchSICKAG, line 3: And here we have a big problem: `@react` exports the react wrappers, which are called EXACTLY like their native counterparts, e.g. `export const SynSwitch:React.FC` and in `@components` `export class SynSwitch: LitElement`. This crashes and makes the usage unintuitive. Suggestion: Only reexport the events for all wrappers, not the other types from `@components`.

packages/_private/react-demo/src/ThemeSwitch.tsx
```diff
@@ -1,9 +1,9 @@
...
-  SynChangeEvent,
-  SynSwitch as SynSwitchElement,
-} from '@synergy-design-system/components';
+import {
+  type SynChangeEvent,
+  SynSwitch,
```
schilchSICKAG, line 4: Am I the react component or LitElement?

packages/_private/react-demo/src/ThemeSwitch.tsx
```diff
@@ -1,9 +1,9 @@
...
+} from '@synergy-design-system/react';
+import type { SynSwitch as SynSwitchElement } from '@synergy-design-system/components';
```
schilchSICKAG, line 6: Am I the react component or LitElement?

schilchSICKAG (changes requested):
I think we have a problem with our exports, details see in comments. We should get this clear before releasing something that may be broken. Main problem stems from the fact that we reexport all types from `@components`, but packages may also export things with the same name, which may lead to confusion (e.g. type SynInput vs SynInput React component).

kirchsuSICKAG:
> I think we have a problem with our exports, details see in comments. We should get this clear before releasing something that may be broken. Main problem stems from the fact that we reexport all types from `@components`, but packages may also export things with the same name, which may lead to confusion (e.g. type SynInput vs SynInput React component).

We decided to not export all events in the index.ts file. It would be much cleaner to not export the events at all in the wrappers, so the types are only used / exported from the components package.
Both of this variants (exporting in index.ts and not exporting the events at all in the wrappers) would lead to a major release.
Because of that we have decided to do neither of those two possibilities.

The only thing which will be done in this PR now, is to unify the handling of the events exports by exporting all used events in the component itself (was missing in the vue wrapper).

packages/components/scripts/jobs/vue/createWrappers.js
```diff
@@ -306,6 +311,10 @@ defineEmits<{
     ${slots}
   </${component.tagName}>
 </template>
+
+<script lang="ts">
```
schilchSICKAG, line 315: Would it be possible to prevent empty `<script lang="ts" />` by checking if the exports map has values, something like?

```typescript
const eventExports = getEventExports(component.events);
const exports = eventExports.length > 0 ? `<script lang="ts">${eventExports}</script>` : '';
```

packages/components/scripts/jobs/vue/createWrappers.js
```diff
@@ -306,6 +311,10 @@ defineEmits<{
...
+  ${eventExports}
+</script>
```
schilchSICKAG, line 317: When using SFCs, it seems that the agreed practice is something like:

```typescript
<script setup />
<script />
<template />
<style />
```

Don´t know if we enforce this currently.

packages/vue/src/components/SynVueButtonGroup.vue
```diff
@@ -64,3 +64,7 @@ defineEmits<{
     <slot />
   </syn-button-group>
 </template>
+
+<script lang="ts">
```
schilchSICKAG, line 68: This should not be on this position. If it is empty, it should be not there at all.

schilchSICKAG (commented):
The only think I am not sure of is the empty `<script />` tags, everything else I just want to annotate, there is nothing breaking

schilchSICKAG (approved):
Perfect, thank you :)

PR #324, opened 2024-02-15
feat: ✨ adjust the size tokens for syn-Input and syn-Button

LeaLeRose:
# Pull Request

### 🎫 Issues
https://github.com/synergy-design-system/design/issues/74

## ✅ DoD

### DIFF
--- a/packages/tokens/src/figma-tokens/semantic/input.json
+++ b/packages/tokens/src/figma-tokens/semantic/input.json
@@ -218,7 +218,7 @@
     },
     "height": {
       "small": {
-        "value": "{spacing.x-large}",
+        "value": "36px",
         "type": "spacing"
       },
       "medium": {
--- a/packages/tokens/src/figma-tokens/semantic/button.json
+++ b/packages/tokens/src/figma-tokens/semantic/button.json
@@ -2,7 +2,7 @@
   "button": {
     "font-size": {
       "small": {
-        "value": "{font-size.x-small}",
+        "value": "{font-size.small}",
         "type": "fontSizes"
       },
       "medium": {


packages/tokens/src/figma-tokens/semantic/input.json
```diff
@@ -218,7 +218,7 @@
     },
     "height": {
       "small": {
-        "value": "{spacing.x-large}",
+        "value": "36px",
```
schilchSICKAG, line 221: I am not sure if I think this is a good way to do this. I understand the size is needed to adjust the input to the buttons height, but is there no other definition that is matching here? When we adjust the value of the small button again, it may well be `42px` or `24px`, which feels kind of planless to me.

Idea: We could more easily couple the two values by using a `min-height` of `calc(--syn-font-size-x-small)` on both elements, which will also align to `36px` and make sure it stays fluid (and computed).

@mariohamann: What do you think? I am torn between: Keep it simple (like the fixed solution in the PR here) or making it more flexible and scalable.

reichst1SICKAG: We don't want to include the 36px into our spacings.
I like to use a token for this, maybe we keep it this way for now and think of some new token for size which is available in the Tokens Studio as type: https://docs.tokens.studio/available-tokens/sizing-tokens

schilchSICKAG: > We don't want to include the 36px into our spacings. I like to use a token for this, maybe we keep it this way for now and think of some new token for size which is available in the Tokens Studio as type: https://docs.tokens.studio/available-tokens/sizing-tokens

You will not need to include it as a new token, you would basically set the tokens value to `calc(3 * {spacing.x-small})`. In this way it would inherit from the small size which it would still honor if it is changed.

mariohamann: I would personally leave it like this – shoelace does the same and I think that's fine: https://shoelace.style/tokens/more#form-inputs

If you really want to redesign things, you will have a look on the tokens anywasy, so that should be okay.

mariohamann (approved):

Issue #326, opened 2024-02-16
feat: ✨ syn-header: Allow automatic interactions with syn-side-nav


schilchSICKAG:
# Checklist (for the issue creator)

- [x] I have looked at the documentation site and was not able to find the required feature.
- [ ] I have looked into the Figma boards and was not able to find the required feature.
- [x] I would be willing to provide a PR for this feature.

# Component

I suggest to add the feature to the following component: `<syn-header />`

# Story

- As a [User] of an application built using the Synergy Design System, I would like to [have the ability to toggle a `<syn-side-nav />`] via a button in the applications `<syn-header />`].
- As a [Developer] using the Synergy Design System, I would like to [have a out of the box solution for toggling a `<syn-side-nav />` from my applications `<syn-header />`] so that [the user is able to toggle the navigation].

# Acceptance criteria

- [ ] All DoD items successfully completed
- [ ] SidebarNav can be toggled via burger menu
- [ ] Burger menu is focusable
- [ ] Burger menu is only shown if sidenav is slotted into header
- [ ] Burger menu should have a hidden label

PR #331, opened 2024-02-19
feat: ✨ Create syn-header

### LINKED ISSUE #299: feat: ✨ Create syn-header

- [ ] I have looked at the documentation site and was not able to find the component.
- [ ] I have looked into the Figma boards and was not able to find the component.

# Story

As a user of the Synergy Design System, I would like to have a syn-header so that I can brand and control my application

# Acceptance criteria

- [ ] Create demo for Framework wrappers
- [ ] The `syn-header` does not interact with slotted navigations initially.
- [ ] Slots are prepared for future use (e.g. for navigations).
- [ ] When developers use the `syn-header`, the SICK logo should be visible per default.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the first iteration of the new `<syn-header />` component, as well as its according test cases, unittests and framework wrappers.

> Be aware that this is not shipping with support for automatic `<syn-side-nav />` handling. This feature will get shipped at a later date in #326.
> As it is not ready yet, the `<syn-horizontal-nav />` is also not part of this ticket. The slot for this is prepared, but no stories are provided.

### 🎫 Issues

Closes #299 

## 👩‍💻 Reviewer Notes

There are currently some known issues that have to be tackled before the release:

- [x] The SICK-Logo is currently baked into the component. It should instead added to the system library by @LeaLeRose or @mariohamann.
- [ ] ~~The icons **close** and **menu** **should** also be added to the system library for future interactive use with `<syn-sidebar-navigation />`.~~ (Will be tackled with #326.
- [ ] ~~The top-navigation story currently uses a `<nav />` element with multiple `<syn-buttons />` as children. It works quite well, however not as well as intended in the Figma boards. Creating the horizontal navigation is not part of this ticket, so I think it still is good enough™.~~ Was not good enough for design, so shelved.
- [x] We are currently not using documentation tokens.
- [ ] ~~We have to define when the **application name** should hide. According to the docs this happens on mobile, but we have not defined what this means yet. Currently, I have not yet added this behavior because of this.~~ We opted to not ship this behavior at all, see comments below

Other than that, the component is pretty functional.

## 📑 Test Plan

Just have a look at the provided stories or the updated framework demos.

## ✅ DoD


packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @dependency syn-icon-button
+ *
+ * @slot - The label for the header. Will automatically be hidden on mobile.
+ * @slot logo - The logo that should be displayed. Will fall back to the SICK logo if not applied.
+ * @slot option-menu - Used to add various application toolbar icons.
+ *                     Best used with `<syn-icon-button />` and `<syn-drop-down />`
+ * @slot top-navigation - Used to add an optional horizontal navigation
+ *
+ * @csspart base - The component's base wrapper.
+ * @csspart side-navigation-button - Can be used to apply styles to the side-navigation menu icon
```
mariohamann, line 29: From my feeling we should skip this part completely in this PR. I would wait with anything like styling or whatever, until we build the side-nav as well. What do you think? From my feeling this lays too much foundation for things that may change when building the side-nav. (And I thought that we explicitly excluded everything related to that in the ticket – I just saw, that it jsut handles "interaractivity", but from my perspective it should skip the side-navigation part completely ATM.

mariohamann: I would propose to copy the parts for the side nav into the other ticket, so that this clearly stated as not shipping it.

mariohamann (changes requested):

packages/components/src/components/header/header.component.ts
mariohamann: My expectation was that it just handles the foundation for the horizontal-navigation (which is not connected for an interactivity perspective) and the top bar things.

schilchSICKAG: > My expectation was that it just handles the foundation for the horizontal-navigation (which is not connected for an interactivity perspective) and the top bar things.

This is now true again. I initially added the menu item as it has its own layout requirements, but I also agree it is saver to not ship it yet.

packages/_private/angular-demo/src/app.component.css
```diff
@@ -1,21 +1,10 @@
-header {
-  align-items: center;
-  background: var(--syn-color-neutral-50);
-  border-bottom: 1px solid var(--syn-color-neutral-400);
+syn-header {
   box-shadow: var(--syn-shadow-large);
```
mariohamann, line 2: @eileenwegner I'm not sure but we don't add shadows to the `syn-header`, right?

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  render() {
+    const hasTopNavigation = this.hasSlotController.test('top-navigation');
+
+    return html`
+      <header
+        class=${classMap({
+          header: true,
+
+          // Is the top navigation available?
```
mariohamann, line 57: I think you can remove this comment. :)

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,99 @@
...
+          'header--has-top-navigation': hasTopNavigation,
+        })}
+        part="header"
+      >
+        <!-- .primary-content-area -->
+        <div part="primary-content-area" class="primary-content-area">
+
+          <div part="logo" class="header__logo">
```
mariohamann, line 65: Following BEM this is not really correct, as you added a class in between, is it?

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,99 @@
...
+            </slot>
+          </div>
+
+          <div part="option-menu" class="header__option-menu">
+            <slot name="option-menu"></slot>
+          </div>
+        </div>
+        <!-- /.primary-content-area -->
+
+        <div part="top-navigation" class="header__top-navigation">
```
mariohamann, line 93: I don't like the structure `primary-content-area` vs. `top-navigation`. Could we iprove this somehow?

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @slot option-menu - Used to add various application toolbar icons.
+ *                     Best used with `<syn-icon-button />` and `<syn-drop-down />`
+ * @slot top-navigation - Used to add an optional horizontal navigation
+ *
+ * @csspart base - The component's base wrapper.
+ * @csspart side-navigation-button - Can be used to apply styles to the side-navigation menu icon
+ * @csspart logo - The wrapper where the application logo resides in
+ * @csspart label - Wrapper of the application name label
+ * @csspart option-menu - Item that wraps the optional application menu
+ * @csspart top-navigation - Wrapper that holds the optional top navigation section
```
mariohamann, line 31: Not sure on this one. In the whole concept of the shell, this makes sense – but calling it "top-navigation" although it is something at the bottom is irritating when having an isolated view on the component.

What do you think about calling it just `navigation`.


Furthermore: Should this be `nav` and `side-nav`? ;)

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @slot logo - The logo that should be displayed. Will fall back to the SICK logo if not applied.
+ * @slot option-menu - Used to add various application toolbar icons.
+ *                     Best used with `<syn-icon-button />` and `<syn-drop-down />`
+ * @slot top-navigation - Used to add an optional horizontal navigation
+ *
+ * @csspart base - The component's base wrapper.
+ * @csspart side-navigation-button - Can be used to apply styles to the side-navigation menu icon
+ * @csspart logo - The wrapper where the application logo resides in
+ * @csspart label - Wrapper of the application name label
+ * @csspart option-menu - Item that wraps the optional application menu
```
mariohamann, line 30: I know that this comes from design – but reading this, I'm not really happy with the wording. We have an `syn-option` and maybe a `syn-menu` and `syn-menu-item` (see https://shoelace.style/components/menu).

Both have nothing to do with this area. It's not a menu.

What do you think about calling it `toolbar`? Or `tools` or similar? Would love to read some opinions on that @kirchsuSICKAG @reichst1SICKAG @LeaLeRose

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ *
+ * @slot - The label for the header. Will automatically be hidden on mobile.
+ * @slot logo - The logo that should be displayed. Will fall back to the SICK logo if not applied.
+ * @slot option-menu - Used to add various application toolbar icons.
+ *                     Best used with `<syn-icon-button />` and `<syn-drop-down />`
+ * @slot top-navigation - Used to add an optional horizontal navigation
+ *
+ * @csspart base - The component's base wrapper.
+ * @csspart side-navigation-button - Can be used to apply styles to the side-navigation menu icon
```
mariohamann, line 27: Leftover.

packages/components/src/components/header/header.styles.ts
```diff
@@ -0,0 +1,83 @@
...
+    justify-content: end;
+  }
+
+  .header__option-menu ::slotted(*) {
+    display: contents;
+    font-size: var(--syn-font-size-x-large);
+  }
+
+  /**
+   * The top navigation is displayed underneath the primary content area
```
mariohamann, line 78: Not sure regarding those comments – they feel more like comments for code and slots, bot not for CSS. I would just mark the areas here, but wouldn't write whole descriptions about thte purpose?

mariohamann (changes requested):

packages/_private/angular-demo/src/app.component.css
reichst1SICKAG: There is no shadow defined, but we have a ticket for shadows: https://github.com/synergy-design-system/design/issues/79

packages/components/src/components/header/header.styles.ts
```diff
@@ -0,0 +1,83 @@
+import { css } from 'lit';
+
+export default css`
+  :host {
+    display: block;
+  }
+
+  .header {
+    background: var(--syn-color-neutral-0);
+    box-shadow: inset 0 -1px 0 0 var(--syn-color-neutral-400);
```
mariohamann, line 10: Hrhr. :D Is this the border-bottom inside the element?

packages/components/src/components/header/header.component.ts
reichst1SICKAG: On the website we have those icons also on the top-right and we are calling it `meta-navigation`

`toolbar` could be also anywhere in the main application.

packages/_private/angular-demo/src/app.component.css
schilchSICKAG: In this case we should make this clear in the docs. There is no documentation that I am aware of where we say the navigation is supposed or even forced to be fixed, alas we could also add it directly to the component.

packages/components/src/components/header/header.styles.ts
mariohamann: TBH I'm not sure if is correct, as this changes the paddings, so that we have 1px less to the bottom. Do you see a problem with border-bottom?

packages/_private/angular-demo/src/app.component.css
mariohamann: For the shown apps we just expected it and for the rail-nav there's no other possibility than having it fixed. But you're absolutetly right,this isn't clearly documented and should defintietely be handled when doing the templates.

packages/components/src/components/header/header.component.ts
mariohamann: `meta-navigation` is nice, especially as this could contain some kind of nav items as well (e. g. to switch apps). What does the others think?

schilchSICKAG: Think so, too. Lets stick with `meta-navigation`.

packages/_private/angular-demo/src/app.component.css
schilchSICKAG: I have removed the box shadow from all demos in the meantime, should be fine now :)

packages/components/src/components/header/header.component.ts
schilchSICKAG: I opted to use `meta-navigation` as the slots title. Will be available on next push.

packages/components/src/components/header/header.styles.ts
schilchSICKAG: Removed this as it really makes no sense

packages/components/src/components/header/header.styles.ts
schilchSICKAG: Just had a talk with @mariohamann, we leave it like this because it will make styling easier in the end (e.g. when using the to be created horizontal navigation, we will be able to slightly override the shadow, but it would be harder when using a border).

packages/components/src/components/header/header.component.ts
schilchSICKAG: Removed in next version

packages/components/src/components/header/header.component.ts
schilchSICKAG: Went with `navigation` initially. However, this leaves us with:

- `meta-navigation` for Toolbars
- `navigation` for the horizontal nav and
- `sidebar-navigation` (later on) for the side navigation.

packages/components/src/components/header/header.component.ts
schilchSICKAG: I opted for the name `header__content` (as generic as it gets...)

packages/components/src/components/header/header.component.ts
schilchSICKAG: We are now using

```html
<div part="content" class="header__content"></div>
<div part="navigation" class="header__navigation"></div>
```

Hopefully this will make the intent more clear.

mariohamann (commented):
I'm fine in general. Just some little the docs things, but besides that it looks super nice!

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,97 @@
...
+  @property() label = '';
+
+  /**
+   * The logo-label attribute can be used as fallback description text
+   * in cases where the logo cannot be loaded.
+   * This only works for the `syn-icon` that gets slotted into the `navigation` as fallback.
+   * If custom content is provided in the `logo` slot,
+   * please take care that it is accessible by yourself!
+   */
+  @property({ attribute: 'logo-label' }) logoLabel = 'SICK Sensor Intelligence';
```
mariohamann, line 57: This feels over-engineered to me and I don't see a reason why people should be able to override the label at this point if they don't change the log. Lots of "ifes". We have a syn-icon and the icon is a system icon, what should go wrong? If there is an issue it's a bug – but pre-fixing this with a prop feels useless.

Currently we have a little problem here: ATM it's only possible to add a link to these elements inside a slot? Therefore we definitely need the logo in the system icons, so people can slot it in combination with an `a href` or `button` whatever they need.

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,97 @@
...
+        })}
+        part="base"
+      >
+        <!-- .header__content -->
+        <div part="content" class="header__content">
+
+          <div part="logo" class="header__logo">
+            <slot name="logo">
+              <syn-icon name="logo-color" library="system" label=${this.logoLabel}></syn-icon>
+            </slot>
```
mariohamann, line 76: We should show a story, how people can set an `<a href="/"><syn-icon name="logo-color" library="system" label=${this.logoLabel}></syn-icon></a>` or similiar if they want to have the logo clickable as this is a function the header currently provides.

Not sure, if there should be a simple effect, like e. g. opacity: 80% automatically on links and buttons inside this slot + correct focus style? Maybe you could clarify with design. I think it might make even sense to show this in design, that this can be focusable and hoverable if correctly set.

This could be a further ticket, so no blocker (but could be helpful from the very start.)

mariohamann (changes requested):

packages/components/src/components/header/header.component.ts
mariohamann: Please discuss with @kirchsuSICKAG, but I don't see any reason for this overridable prop, as this even is ALWAYS the SICK logo.

packages/components/src/components/header/header.component.ts
reichst1SICKAG: To make the SICK logo as a link is definetely needed.
Focus style, yes. But I've rarely seen a hover effect on a linked logo and would not go for that.

packages/components/src/components/header/header.component.ts
schilchSICKAG: Thanks for the input. I see where you are coming from with this. When I created it I just thought about using the `label` prop as default, but it may also very well be the user is providing a `<slot />` for this, making it harder to get to the content. As we only need this when NOT providing a custom logo, I too see no reason for this to exist anymore, so opted to remove it.

packages/components/src/components/header/header.component.ts
schilchSICKAG: I have implemented this in the latest version. Please note that you will have to apply some styling by yourself when using slots. I am able to add the `:focus` outline without a problem, but developers will have to define the dimensions of logos by themselfs as it is not that easy to override.

schilchSICKAG:
> Will do the review after the comments in Design are resoved, as they might trigger some changes.

I just pushed a new version where the following changes where applied:

- The left padding of the container is now using `--syn-spacing-large`, so we have the same paddings applied to all sides.
- The focus state is now only visible when using keyboard interaction.

packages/components/src/components/header/header.component.ts
```diff
@@ -0,0 +1,88 @@
...
+import styles from './header.styles.js';
+import SynIcon from '../icon/icon.js';
+
+/**
+ * @summary The <syn-header /> element provides a generic application header
+ * that can be used to add applications name, toolbar and primary navigation.
+ *
+ * @documentation https://synergy-design-system.github.io/?path=/docs/components-syn-header--docs
+ * @status stable
+ * @since 1.8.0
```
kirchsuSICKAG, line 18: The version needs to be adapted to 1.10.0

```suggestion
 * @since 1.10.0
```

kirchsuSICKAG:
Looks nice :) 
The only thing I was asking myself is: do we need to handle this , is it something the user has to handle or is it event the right behavior that we want: 
if the app name is long and the header is getting smaller (e.g. browser is getting smaller and so on), the height of the header changes because the app name is getting multi line (see gif)

syn-header

kirchsuSICKAG (changes requested):

schilchSICKAG:
> Looks nice :) The only thing I was asking myself is: do we need to handle this , is it something the user has to handle or is it event the right behavior that we want: if the app name is long and the header is getting smaller (e.g. browser is getting smaller and so on), the height of the header changes because the app name is getting multi line (see gif)
> 
> syn-header [ syn-header ](https://private-user-images.githubusercontent.com/115139104/308563513-ce671bfe-8da3-4f29-883e-8d3aaadce041.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDkxMjcwNjIsIm5iZiI6MTcwOTEyNjc2MiwicGF0aCI6Ii8xMTUxMzkxMDQvMzA4NTYzNTEzLWNlNjcxYmZlLThkYTMtNGYyOS04ODNlLThkM2FhYWRjZTA0MS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMjI4JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDIyOFQxMzI2MDJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1mMWM4ZTgwYjU1NTAzOTlmNWU2NDZiMDZkNzI5YWM0MDM0MGZkNWNmMzczMjk3MmU5MTI5NTY2MjI4NjMyYjc2JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.bg8uMP2OqrJ8ivBNVxAPCLlRI2tcEhMZxi34iHzfuXI) [ ](https://private-user-images.githubusercontent.com/115139104/308563513-ce671bfe-8da3-4f29-883e-8d3aaadce041.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDkxMjcwNjIsIm5iZiI6MTcwOTEyNjc2MiwicGF0aCI6Ii8xMTUxMzkxMDQvMzA4NTYzNTEzLWNlNjcxYmZlLThkYTMtNGYyOS04ODNlLThkM2FhYWRjZTA0MS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwMjI4JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDIyOFQxMzI2MDJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1mMWM4ZTgwYjU1NTAzOTlmNWU2NDZiMDZkNzI5YWM0MDM0MGZkNWNmMzczMjk3MmU5MTI5NTY2MjI4NjMyYjc2JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.bg8uMP2OqrJ8ivBNVxAPCLlRI2tcEhMZxi34iHzfuXI)

We discussed this feature [here](https://github.com/synergy-design-system/synergy-design-system/pull/331#discussion_r1498769948) already and opted to just not include it per default, as we did not add it to the menu bar.

kirchsuSICKAG (approved):

Issue #304, opened 2024-02-05
docs: 📚 improve top level storybook documentation


klaarseSICKAG:
# Description

As a user of synergy I want to have a quick overview of what to expect from Synergy and what I can do with it.

# Rationale

Currently the user is confronted with a very short introduction and a huge presentation. Most users do not want to click through a presentation (which might also contain unneeded information, long loading times, password needed, no search support) to get the information they like.
On the other hand, the current introduction doesn't contain enough information to get the full picture.

# Acceptance criteria
The Storybook Start page ...
- [x]  ... includes an easy to use navigation that points users to the following entry points: Wrapper Packages (Angular, React, Vue), Vanilla Package. Each of those navigation items should show the Icon of the framework to make it easier to find.
- [x] ... includes a **more prominent link** to GitHub (this was often requested when talking with stakeholders)
- [x] ... includes a link to the figma board

PR #334, opened 2024-02-21
feat: ✨ Add utility classes for typography

### LINKED ISSUE #269: feat: ✨ Add utility classes for typography

# Story

As a user of the Synergy Design System, I would like to use typographic styles for headline and body text in a easy way so that my product uses the generally valid styles.

Figma branch: https://www.figma.com/file/bZFqk9urD3NlghGUKrkKCR/branch/4oqUjJbgNpMsX7fqfgOMyl/Synergy-Digital-Design-System?type=design&node-id=104%3A235&mode=design&t=LdKsFz2sUGWhh0kP-1

# Acceptance criteria

- [x] check the generated tokens from branch feat/add-text-styles from Figma
- [x] Create a new package with the name that fit future usecases
- [x] Create classes instead of tokens synchronization
- [x] Create documentation in Storybook (e.g. Readme, Changelog, Story for all typographic items) 
- [x] Use typography for Storybook stories
- [x] Offer a css file for basic styling
- [ ] Make sure to offer a singular css token for each font size that does **NOT** include font-weight, but makes sure that current variables are utilized correctly. See below for details:

## Token structure

Given the following JSON:

```json
  "body": {
    "x-small": {
      "regular": {
        "value": {
          "fontFamily": "{font.sans}",
          "fontWeight": "{font-weight.normal}",
          "fontSize": "{font-size.x-small}",
          "lineHeight": "{line-height.normal}"
        },
        "type": "typography"
      }
    }
  }
```

the following token should be created:

```css
:root {
  --syn-body-x-small: var(--syn-font-size-x-small)/var(--syn-line-height-normal) var(--syn-font-sans);
}
```

schilchSICKAG:
# Pull Request

## 📖 Description

This PR provides a new utility package `@synergy-design-system/styles` based on the ideas and feedback gathered at https://github.com/orgs/synergy-design-system/discussions/278.

The new package provides the following features:

- High-Level CSS Outputs (see Storybook for when to use which, Feedback for the README is highly anticipated)
- Low level modules (e.g. for forms only) 

### 🎫 Issues

Closes #269.

## 👩‍💻 Reviewer Notes

Please have a look at storybook in this branch. The version in this branch should show multiple new items:

- `Packages/Tokens` shows the CHANGELOG and README files, as with our other packages
- `Tokens/Typography` Shows the underlying typographic styles for body copy and headlines
- `Styles/Forms` Shows the usage of form based helper classes
- `Styles/Typography` shows how to use the typographic utilities
- Storybook has been updated to NOT use the low level tokens anymore, but uses the "all.css" theme.

Please also have a look at the modules code and how the modules are structured. Are there any more changes needed? Are the output targets aptly named? Is the documentation satisfying? For example, I also would like to add a decision page in Storybook somewhere that shows when to use the tokens and when to use the styles. Do you have any suggestions for this (e.g. where to place?). Also, what does that mean for the READMEs of Wrappers and Components (currently, we are telling people to install `@synergy-design-system/tokens`, but this now has the possibility to use `@synergy-design-system/styles` instead. How do we make it clear when to use which?

## ✅ DoD


.vscode/settings.json
```diff
@@ -12,6 +12,7 @@
     "packages/_private/stylelint-config-syn",
     "packages/assets",
     "packages/components",
+    "packages/css",
```
kirchsuSICKAG, line 15: should be `packages/styles`

```suggestion
    "packages/styles",
```

packages/styles/src/modules/index.css
```diff
@@ -0,0 +1,11 @@
+/**
+ * This file holds all import for utility functions
```
kirchsuSICKAG, line 2: typo

```suggestion
 * This file holds all imports for utility functions
```

packages/styles/src/modules/typography.css
```diff
@@ -0,0 +1,90 @@
...
+ * May be combined with weight classes defined above
+ */
+.syn-body-small {
+  font: var(--syn-body-small-regular);
+}
+
+/**
+ * Utility class for setting a medium body copy font style
+ * This is the default style for elements for synergy applications
+ * May be combined with weight classes defined above
```
kirchsuSICKAG, line 24: The weight classes are defined below:

```suggestion
 * May be combined with weight classes defined below
```

packages/styles/src/modules/typography.css
```diff
@@ -0,0 +1,90 @@
...
+ */
+.syn-body-medium {
+  font: var(--syn-body-medium-regular);
+}
+
+/**
+ * Utility class for setting a large body copy font style
+ * May be combined with weight classes defined above
```
kirchsuSICKAG, line 32: The weight classes are defined below:

```suggestion
 * May be combined with weight classes defined below
```

packages/styles/README.md
```diff
@@ -0,0 +1,37 @@
+# @synergy-design-system/styles
+
+This packages provides utility classes and basic document stylings intended for **new applications and websites** using `@synergy-design-system/components` or one of its derived wrapper packages.
```
kirchsuSICKAG, line 3: Typo

```suggestion
This package provides utility classes and basic document stylings intended for **new applications and websites** using `@synergy-design-system/components` or one of its derived wrapper packages.
```

kirchsuSICKAG:
For me I can't see any difference between our semi-bold and bold token (see pic).
Semi-bold is 600 and bold 700. Is this right?

image

packages/docs/stories/styles/typography.stories.ts
kirchsuSICKAG, line 1: Stories for the weight classes are missing

packages/docs/stories/tokens/Typography.stories.mdx
```diff
@@ -0,0 +1,100 @@
...
+      <th>SCSS Token</th>
+      <th>JS Token</th>
+    </tr>
+  </thead>
+  <tbody>
+    {[
+      'SynBody-X-SmallRegular',
+      'SynBody-Small-Regular',
+      'SynBody-Medium-Regular',
+      'SynBody-Large-Regular',
```
kirchsuSICKAG, line 84: I would not only add the regular body tokens here but also the semibold (e.g. --syn-body-x-small-semibold) and bold ones. Otherwise its hard for the users to know how they are called if they want to use them.

kirchsuSICKAG (changes requested):

packages/docs/stories/tokens/Typography.stories.mdx
schilchSICKAG: I already had a discussion with @mariohamann on this. We opted to just provide the default ones and also add support for utility classes (`syn-weight-*`) for this to use. This is also made clearer in the story for the typography utility classes. Unfortunately, we are not able to strip the semibold / bold things out of the stylesheet in an easy way, so technically they will be there, but not documented.

kirchsuSICKAG:
> Please also have a look at the modules code and how the modules are structured. Are there any more changes needed? Are the output targets aptly named?

I ask myself if we really want to separate outputs and modules. I don`t really like, that the user has to do something like 
`import @synergy-design-system/styles/src/modules/body.css`
Maybe it would be enough to also let the modules folder be generated into the `dist`. So that we have an `outputs` and `modules` folder. With this we would get rid of the **src** at least in the import statement

Everything else (naming, etc seems fine for me)

kirchsuSICKAG:
> Is the documentation satisfying? For example, I also would like to add a decision page in Storybook somewhere that shows when to use the tokens and when to use the styles. Do you have any suggestions for this (e.g. where to place?)

Maybe we could either do a new file on the same level like the `Welcome` page called something like `Recommendations` or if there will be much more in the future maybe a whole new title with `Recommendations` and multiple files under it. 
Or maybe a naming like `Best practices`

schilchSICKAG:
> > Please also have a look at the modules code and how the modules are structured. Are there any more changes needed? Are the output targets aptly named?
> 
> I ask myself if we really want to separate outputs and modules. I don`t really like, that the user has to do something like `import @synergy-design-system/styles/src/modules/body.css`Maybe it would be enough to also let the modules folder be generated into the`dist`. So that we have an `outputs`and`modules` folder. With this we would get rid of the **src** at least in the import statement
> 
> Everything else (naming, etc seems fine for me)

The user will have the following possibilities:

### Include one of the output targets directly. This can be done by loading, e.g in vite:

```typescript
// Full support with all output modules.
// This would be the recommended import way as you get everything
// Only working for systems that support package.json exports map!
import '@synergy-design-system/styles'; // Will load styles/dist/all.css
import '@synergy-design-system/styles/css/all.css'; // Explicit, with dark/light
import '@synergy-design-system/styles/css/all-with-auto-theme.css'; // Explicit, with auto theme
```

### Mix and match what you want

This can be used if you for example just want the fouc styles, but not typo

```typescript
import '@synergy-design-system/src/modules/fouc.css'; // I only want fouc prevention
```

The `src` folder is part of the bundle, so basically we are letting the user decide, with a strong opinion on our side to use the generated output bundles.

schilchSICKAG:
> For me I can't see any difference between our semi-bold and bold token (see pic). Semi-bold is 600 and bold 700 (also in Figma). Is this right?
> 
> image
> 
> In Figma a difference can be seen: image

This seems to be an issue with your locally loaded font. If it does not support different font-weights, it will fall back to the commonly used one. On MacOS based systems, you will see the difference.

kirchsuSICKAG:
> The src folder is part of the bundle, so basically we are letting the user decide, with a strong opinion on our side to use the generated output bundles

yes, i just don't like the `src` in the import statement there. But we want to try to influence the user by this it's fine for me ^^

schilchSICKAG:
> > Is the documentation satisfying? For example, I also would like to add a decision page in Storybook somewhere that shows when to use the tokens and when to use the styles. Do you have any suggestions for this (e.g. where to place?)
> 
> Maybe we could either do a new file on the same level like the `Welcome` page called something like `Recommendations` or if there will be much more in the future maybe a whole new title with `Recommendations` and multiple files under it. Or maybe a naming like `Best practices`

Yes, I also thought about something like adding something like "basic setup instructions" for new project. This would also come in handy when we adjust the start page as described in #304? Giving some basic examples would really be good. Basically, you just don´t need to `@synergy-design-system/tokens` package at all for most projects. It is only needed when creating custom themes for frameworks (e.g. when using material ui react wrapper). In most cases, you should be fine using the variables, that in turn come from `@synergy-design-system/styles`.

> There we can also add an info where we say something like: "you can set the .syn-heading-3x-large onto a div, but please try to stick to semantic html" (or something in this direction) Because this info is missing for me right know somewhere

I started this with an inline documentation in the Styles section, e.g. if you check out the source after setting the style to a headline, you will get a `<div>` that is styles as a headline, but also includes the correct aria roles.

packages/docs/stories/styles/typography.stories.ts
schilchSICKAG: As described, this will be handeled via weight utility classes. Those can be toggled in the Styles/Typography story

packages/styles/src/modules/fouc.css
```diff
@@ -0,0 +1,52 @@
...
+syn-radio-group:not(:defined),
+syn-select:not(:defined),
+syn-spinner:not(:defined),
+syn-switch:not(:defined),
+syn-tag:not(:defined),
+syn-textarea:not(:defined) {
+  display: inline-block;
+  opacity: 0;
+  transition: opacity 0.3s ease-in-out;
+ }
```
mariohamann, line 32: Is styleling working here? It feels like there is a wrong space here. :)

packages/styles/src/modules/typography.css
```diff
@@ -0,0 +1,90 @@
...
+.syn-weight-semibold {
+  font-weight: var(--syn-font-weight-semibold);
+}
+
+/**
+ * Defines a font weight of bold, according to synergy specs
+ */
+.syn-weight-bold {
+  font-weight: var(--syn-font-weight-bold);
+}
```
mariohamann, line 89: Do you think we need `weight` here in the name? Wouldn't it be more straightforward for a utility class to abandon it? I know, that syn-normal looks a bit strange and should be more `syn-regular`. (Maybe that's a problem in general.)

No blocker – just a thought.

reichst1SICKAG: I also like `syn-bold` and `syn-semibold`. It's straightforward.

schilchSICKAG:
> > The src folder is part of the bundle, so basically we are letting the user decide, with a strong opinion on our side to use the generated output bundles
> 
> yes, i just don't like the `src` in the import statement there. But we want to try to influence the user by this it's fine for me ^^

I have added an optional export target, so you are now able to include the modules via:

```javascript
import '@synergy-design-system/styles/modules/body.css`
```

packages/styles/src/outputs/utilities.css
```diff
@@ -0,0 +1,5 @@
+/**
+ * This bundle only includes the utility classes
+ */
+ @import url("../modules/index.css");
```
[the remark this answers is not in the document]
schilchSICKAG: This one is meant to only include the helpers for cases you do not want the variables from tokens. With this, you are able to do one of the following:

```javascript
// Example 1: Load everything, including tokens (first dark, then light)
import '@synergy-design-system/styles/css/all.css';

// Example 2: Load everything, including tokens. Applies the tokens via prefers-color-scheme
import '@synergy-design-system/styles/css/all-with-auto-theme.css';

// Example 3: Only load the utilities. Loading of tokens is done somewhere else (e.g. for custom styling)
import '@synergy-design-system/styles/css/utilities.css';
```

packages/styles/src/modules/fouc.css
schilchSICKAG: It already works in the adjusted storybook instance. Of course it is opinionated, but better to ship nothing. Maybe we can create a better version for this in the future. I already have some ideas to prevent jumps for the elements.

packages/styles/src/modules/typography.css
schilchSICKAG: I don´t have any problems with this. Classnames will get rather long if we prefix everything. When using `bold` or `semibold` it is indeed shorter and descriptive enough. I am not so sure that I would detect the meaning of `<div class="syn-regular">Hello World</div>` on my first glimpse of a piece of code.

schilchSICKAG: Yes it it. However you may run into something like:

```html
<p>
  <strong>Most of this sentence is bold, <span class="syn-regular">but this part is not</span></strong>
</p>
```

In this case, you would have to use the "regular" class.

schilchSICKAG:
Closed, handled via https://github.com/synergy-design-system/synergy-design-system/pull/495 now.

PR #343, opened 2024-02-26
docs: 📚 optimize documentation (close #297)

### LINKED ISSUE #297: docs: 📚 always provide meaningful input fields for all attributes in storybook
# Description

In some cases confusing options and controls are used in storybook.
e.g. https://synergy-design-system.github.io/?path=/docs/components-syn-input--docs
--> for attribute "step" a select is used which allows setting "number" or "any"
   --> expected control is a number-input for configuring the step width.
--> for attribute "min"/"max" a radio is used with options "number" or "string"
  --> expected control is a number-input for configuring the minimal or maximum values

## Acceptance criteria

- [ ] DoD Items
- [ ] Show a default text input when the type can not be determined

mariohamann:
# Pull Request

## ✅ DoD


mariohamann:
Currently there seem to be errors on Chromatic:

CleanShot 2024-02-26 at 07 45 25@2x

I can't reproduce them locally or in the published Storybook. I expect this to be an error on Chromatic side – please try to re-run the tests soon-ish and check with other PRs. Besides that maybe someone else has time to investigate?

klaarseSICKAG:
Looks definitely better than before! Just a minor thing to the input fields: I'd prefer an input type text instead of a textarea. Intuitively I press enter after entering a value. I expect that the value is then set and not a CR.

schilchSICKAG:
> Currently there seem to be errors on Chromatic:
> 
> CleanShot 2024-02-26 at 07 45 25@2x
> 
> I can't reproduce them locally or in the published Storybook. I expect this to be an error on Chromatic side – please try to re-run the tests soon-ish and check with other PRs. Besides that maybe someone else has time to investigate?

This seems to be an issue with Safari which is currently very unhappy with our bundle. We already created a ticket for this and disabled Safari tests for the time being.

mariohamann:
> Looks definitely better than before! Just a minor thing to the input fields: I'd prefer an input type text instead of a textarea. Intuitively I press enter after entering a value. I expect that the value is then set and not a CR.

Understand that, but apparently this is the default, and other design systems have the same (couldn't find something different):
- GitHub Primer React: https://main--61a90feace7802003a4d9c45.chromatic.com/?path=/story/components-avatartoken--default-token
- GitLab: https://gitlab-org.gitlab.io/gitlab-ui/?path=/docs/base-icon--docs
- Storybook: https://master--5ccbc373887ca40020446347.chromatic.com/?path=/docs/button--docs

I suppose this comes from the fact that for everything with the type `string` `textarea` is the best compromise. Therefore I would leave this like is.

package.json
```diff
@@ -32,7 +32,8 @@
   "pnpm": {
     "patchedDependencies": {
       "semantic-release-monorepo@7.0.5": "patches/semantic-release-monorepo@7.0.5.patch",
-      "@open-wc/lit-helpers@0.5.1": "patches/@open-wc__lit-helpers@0.5.1.patch"
+      "@open-wc/lit-helpers@0.5.1": "patches/@open-wc__lit-helpers@0.5.1.patch",
+      "wc-storybook-helpers@1.3.1": "patches/wc-storybook-helpers@1.3.1.patch"
```
schilchSICKAG, line 36: This is the only thing that I am not a 100% sure with. I know pnpm patches sound like a great idea, however, you could also add the changed code to `packages/docs/src/helpers/component.ts` into the storybookDefaults function. This would have made it possible to easily upgrade `@open-wc/lit-helpers` when there is a new release, which unfurtunately is not possible currently.

schilchSICKAG (commented):
If the others are fine with it, I am too. Just seeing the risk of non (easily) updateable packages again.

package.json
mariohamann: I understand the perspective. I tried to have the following approach:
1. Things that are buggy are patched. (https://github.com/synergy-design-system/synergy-design-system/pull/343/files#diff-766d360edf31dd65886b2efed53ffdc8cad96fb8577d6c2913d1f5af7847b779R69-R84)
2. Things that would have changes in the helper AND would have to be patched are only patched (e. g. https://github.com/synergy-design-system/synergy-design-system/pull/343/files#diff-766d360edf31dd65886b2efed53ffdc8cad96fb8577d6c2913d1f5af7847b779R101-R103)
3. Things that are easily addable/overridable on our side are added in the helper, e. g. https://github.com/synergy-design-system/synergy-design-system/pull/343/files#diff-b0ab8fca684831c354518633163bf08d0188724b7b25440a39f92fd37b104421R100-R104

For the first two issues would try to make a PR for the original package. As the author already took some other things from my fork, I believe this will be quite lightweight to bring in. I would make a follow up issue for that.

Fine for you?

PR #364, opened 2024-03-13
feat: ✨ add syn-side-nav, syn-nav-item, syn-prio-nav

### LINKED ISSUE #312: feat: ✨ syn-side-nav
# Story

As a user of the Synergy Design System, I would like to provide my users a way to navigation through the application.

# Acceptance criteria

- [ ] A11Y - Support
- [ ] Use drawer as sub element
- [ ] Should be based on Figma designs
- [ ] Uses a list of nav-items
- [ ] Provides footer (e.g. for pin or nav-items) and main navigation area (for nav-items)
- [ ] Can display full or rail
- [ ] Support for fixed and shrinking content (content displacement)

### LINKED ISSUE #313: feat: ✨ syn-nav-item
# Story

As a user of the Synergy Design System, I would like to navigate through my application

# Acceptance criteria

- [x] Should work as link or button
- [x] Accordion functionality
- [x] Use Solid as template (https://github.com/solid-design-system/solid/blob/main/packages/components/src/components/navigation-item/navigation-item.ts)

### LINKED ISSUE #317: feat: ✨ prio-nav
# Story

As a user of the Synergy Design System, I would like to use a header navigation to navigate through my application.

# Acceptance criteria

- [x] Should be able to display an arbitary number of navitation items (e.g. `<syn-nav-item />`, `<a[href] />`, `<button />`
- [x] Collapse items into a `<syn-drop-down />` element if space isn't sufficient as outlined in Figma.
- [x] Should follow a11y best practices
- [x] Should be able to navigate via keyboard

### LINKED ISSUE #326: feat: ✨ syn-header: Allow automatic interactions with syn-side-nav

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds the `<syn-nav-item />`, `<syn-side-nav />` and `<syn-horizontal-nav />` components as well as their according test cases, unit tests and framework wrappers.

### 🎫 Issues

Closes #313
Closes #312
Closes #317
Closes #326

## 👩‍💻 Reviewer Notes

We made one branch for all 3 components, because the nav-item is the basis for the other two components.

### *Newly added components:*
- syn-nav-item
- syn-prio-nav
- syn-side-nav

### *Changed component with added feature:*
- syn-header

## ✅ DoD


packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
mariohamann, line 1: General question: Should we maybe rename this to `priority-nav`? Before ask that design, what's your opinion on that @kirchsuSICKAG @schilchSICKAG?

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    ) as HTMLElement[];
+
+    return slottedItems.filter(i => i instanceof SynNavItem) as SynNavItem[];
+  }
+
+  /**
+   * Adjust the visual placement of the priority menu
+   * @param startPos The start position to calculate from
+   */
+  private adjustPriorityMenuPosition(startPos: number) {
```
mariohamann, line 91:
```suggestion
  private adjustPriorityMenuPosition(startPosition: number) {
```

We barely use abbreviations in code.

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
```diff
@@ -0,0 +1,196 @@
...
+    return html`
+      <div
+        class=${classMap({
+          'priority-menu': true,
+          'priority-menu--hidden': items.length === 0,
+        })}
+        part="priority-menu"
+      >
+        <syn-nav-item>
+          <syn-icon name="more_horiz" label="More" slot="prefix"></syn-icon>
```
mariohamann, line 156: `more_horiz`? :D Is this the original name?  OMG. This is really the original material name...

This should be a system icon, right? And maybe we can name it nicer then...

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
```diff
@@ -0,0 +1,196 @@
...
+          <span class="priority-menu__label" part="priority-menu-label">${this.priorityMenuLabel}</span>
+          <nav class="priority-menu__list" slot="children">
+            ${items.map((item, index) => html`
+              <syn-nav-item
+                ?divider=${index !== 0}
+                vertical
+              >
+                ${item.textContent}
+              </syn-nav-item>
```
mariohamann, line 165: I have very big concerns in re-rendering elements here. We don't know, what people do with their nav items, maybe they use them with event listeners, maybe with links, maybe with whatever. This prevents to do anything individual with them.

I would strongly suggest to MOVE items and when moving only toggle the `vertical` attribute on syn-nav-items on the first level, but do nothing else here. I think moving between slots could be the best thing here, so we just toggle a `slot=more` attribute. 

To be honest: I think it would be even fine, if they don't use this with a `syn-nav-item` at the top – I believe as  a developer I maybe would do more something like:

```html
<syn-horizontal-nav>
<div class="nav-item-wrapper">
   <syn-nav-item>I am horizontal</syn-nav-item>
   <syn-nav-item vertical>I am vertical and do things differently</syn-nav-item>
</div>
</syn-horizontal-nav>

<styles>
.nav-item-wrapper[slot="more"] syn-nav-item:not([vertical]){display: none;}
.nav-item-wrapper:not([slot="more"]) syn-nav-item[vertical]{display: none;}
</styles>
```

Initially here the vertical nav item is hidden.

When the item doesn't have enough space, it would move: 
```html
<div class="nav-item-wrapper" slot="more">
   <syn-nav-item>I am horizontal</syn-nav-item>
   <syn-nav-item vertical>I am vertical and do things differently</syn-nav-item>
</div>
```

This would hide the horizontal element and show the vertical one.

So my proposal summarized:
1. Toggle a `slot=more` attribute to indicate if elements move into a menu or not.
2. Toggle `vertical` if those elements are `syn-nav-items`. If REALLY needed, you could even show the divider programmatically.

packages/components/src/components/side-nav/side-nav.component.ts
```diff
@@ -0,0 +1,291 @@
...
+   * This function handles the rail mode.
+   * Rail mode is only valid if every nav-item has a prefix.
+   */
+  // eslint-disable-next-line complexity
+  private handleRailMode() {
+    const navItems = this.getAllNavItems();
+
+    const itemsWithPrefix = navItems.filter((navItem) => !!navItem.querySelector(':scope > [slot="prefix"]'));
+
+    this.hasPrefixIcons = navItems.length !== 0 && itemsWithPrefix.length === navItems.length;
```
mariohamann, line 140: I'm curious: What would happen if someone doesn't set a prefix and we wouldn't have this check?

mariohamann: And if we really need this: How can we indicate users, that they have to set the prefixes? Console.log?

packages/docs/stories/components/side-nav.stories.ts
mariohamann, line 1: We should show those stories with the syn-header, shouldn't we? They are so closely connected and as developer I want to know how to use them together already in the side-nav docs.

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
schilchSICKAG: Seems like a nice idea. However, we will have to make sure there are clones that occupy the space, because otherwise the navigations size will change and cause infinite rerenders. This was the reason I created copies. However as you have pointed out, even when using `cloneNode`, event listeners will not get propagated. So maybe we could do both:

1. Let the current size check stay as it is to make sure to position the dropdown correctly.
2. Create dummies in the dom with the same relative positions than the original items and set an absolute position on them with the same base coordinates (will have to be synced with the dummies then...)
3. Move the items to the dropdown completely like you suggested.

I suggest we should have a talk about this tomorrow.

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
schilchSICKAG: If you mean the slot `priority-menu`, I am fine with this and the needed change for `priority-menu-label` to use `nav` instead 👍

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
schilchSICKAG: Yeah, it unfortunately is the "currect" name :). I am unsure if we may get problems with the renaming. Currently, it is pretty consistent. However, I personally have no problems giving it another name. We could do the same as with the `caret`, which we use to transform between left/right/top/bottom and also exists in the material library. So giving it a name of `more` and just toggling it between horizontal and vertical mode would be fine for me (at least if the original icons are the same with different orientation)

packages/docs/stories/components/side-nav.stories.ts
schilchSICKAG: I am unsure about this, as we try to decouple stories currently and there are also the templates in the works. There will eventually be stories when #326 is done. @kirchsuSICKAG: What do you think about this?

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
kirchsuSICKAG: would be fine for me too

packages/docs/stories/components/side-nav.stories.ts
kirchsuSICKAG: I was also thinking about adding the header.
Right now we only have documentation in design, where the side-nav itself with the content part on the right is shown.
But for some cases the header would be definitely helpful. E.g. for showing how to handle shrinking the main content. Right now I have a button as workaround for me to toggle the open / close state of the side-nav, so I don't have to add / remove the open attribute on the dom tree all the time..

packages/components/src/components/horizontal-nav/horizontal-nav.component.ts
mariohamann: This sounds reasonable. So if I get right:
1. Invisible copies for calculating sizes.
2. Slot switching for movement.

If this somehow works, that sounds good. As an alternative: We could just store the "initial" sizes in the elements and check, if there is enough space? That's how we handled it at uni-freiburg.de

packages/docs/stories/components/side-nav.stories.ts
mariohamann: Hm. To me it feels like e.g. `select` and `option` – both components are connected to each other. And they highly depend on each other. I especially would expect that e. g. there's a hamburger menu if there's no `rail`, but not a hamburger menu if there is a `rail`. I don't have the feeling, that showing this in a template is enough as this is functional dependency, which maybe even deserves screenshots or interactive stories. Hmm...

packages/components/scripts/vendorism/custom/translations.vendorism.js
schilchSICKAG, line 1: Would you mind if I refactor this to the version in https://github.com/synergy-design-system/synergy-design-system/pull/390/files#diff-ba292985634a78a2680369ac2044ccab1eb8566ecc6a80ccde2a66f927a8aadd? This would have the benefint that it would be a) mergeable and b) does not require special treatment for each component.

packages/docs/src/DisableFocusTrap.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Custom DocsContainer that disables focus trapping for all syn-side-nav and syn-drawer elements.
+ */
+export const DisableFocusTrap: FC<DisableFocusTrapProps> = ({ children, ...props }) => {
+  const ref = React.useRef<HTMLDivElement>(null);
+  React.useEffect(() => {
+    // Needed otherwise the side-nav / drawer can`t be found
+    setTimeout(() => {
+      // Disable focus trapping for all syn-side-nav`s
```
schilchSICKAG, line 17: Maybe this is because the element is not yet defined when this is called? How about:

```typescript
useEffect(() => {
  const waitForDefined = async () => {
    await customElements.whenDefined('syn-side-nav');
    await customElements.whenDefined('syn-drawer');
    
    // Further focus trapping logic...
  }

  waitForDefined();
}, []);
```

kirchsuSICKAG:
@mariohamann a question related to the syn-nav-item:
The syn-nav-item can either be under the hood
- a `button`
- a `link` ( if the href prop is added)
- or a `details` ( if the nav-item gets other nav-items in it`s children slot)

What should happen if someone sets the href attribute, but also does nav-items in the children slot. Should it then be a `link` or a `details`.
Currently it would be a `details` (the documentation needs to be aligned, because it says the `link` has prio) but before I will do that, I would like to hear you opinion

mariohamann:
> @mariohamann a question related to the syn-nav-item: The syn-nav-item can either be under the hood
> 
> * a `button`
> * a `link` ( if the href prop is added)
> * or a `details` ( if the nav-item gets other nav-items in it`s children slot)
> 
> What should happen if someone sets the href attribute, but also does nav-items in the children slot. Should it then be a `link` or a `details`. Currently it would be a `details` (the documentation needs to be aligned, because it says the `link` has prio) but before I will do that, I would like to hear you opinion

Technically it can't be a link and a button at the same time. AFAIK it should be no "Index" pages, but should only contain Subpages.

See e. g. here:
https://github.com/synergy-design-system/synergy-design-system/assets/26542182/739e06ab-e4d0-4717-bb63-404af0d2f160

https://m3.material.io/foundations

If a combination should be possible, we would have to seperate the Chevron and the Text or similar.

packages/components/src/components/prio-nav/prio-nav.component.ts
```diff
@@ -0,0 +1,266 @@
...
+      <syn-dropdown
+        class=${classMap({
+          'priority-menu': true,
+          'priority-menu--hidden': !this.hasItemsInDropdown,
+        })}
+        part="priority-menu"
+        placement="bottom-end"
+      >
+        <syn-nav-item slot="trigger" horizontal>
+          <syn-icon name="more" library="system" label="More" slot="prefix"></syn-icon>
```
kirchsuSICKAG, line 207: The `part`'s are missing here.
And don`t we also want to make the prio-menu-icon exchangeable via slot like in all other components? 
What do you think @mariohamann ?

kirchsuSICKAG: I talked about this with Mario. We came to the point, that this is a slot which is used for internal functionality and therefore should not be exposed with the official documentation

packages/components/src/components/prio-nav/prio-nav.component.ts
```diff
@@ -0,0 +1,264 @@
...
+ * @documentation https://synergy-design-system.github.io/?path=/docs/components-syn-prio-nav--docs
+ * @status stable
+ * @since 1.10
+ *
+ * @dependency syn-dropdown
+ * @dependency syn-icon
+ * @dependency syn-menu
+ * @dependency syn-nav-item
+ *
+ * @slot - The given navigation items. Must be `<syn-nav-item>`s or have a role of "menuitem"
```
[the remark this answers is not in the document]
kirchsuSICKAG: I talked about this with Mario. We came to the point, that this is a slot which is used for internal functionality and therefore should not be exposed with the official documentation

packages/components/src/components/prio-nav/prio-nav.component.ts
kirchsuSICKAG: i added `part`'s to all elements

mariohamann (approved):

PR #390, opened 2024-03-25
feat: ✨ syn-badge

### LINKED ISSUE #378: feat: ✨ syn-badge
# Story

As a user of the Synergy Design System, I would like to use a badges to draw attention and display statuses or counts.

# Acceptance criteria

- [x] Incorporates slots, Events, Methods, Parts from Shoelace
- [x] Tested: Playwright + Chromatic
- [x] Create a proper documentation which contains at least everything from the Shoelace component
- [x] Text in the documentation can be synced
- [x] non-pill variant und pulse are removed

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds support for `<syn-badge />`, as well as placement helpers to use badges in `<syn-button />` and `<syn-menu />`.

### 🎫 Issues

Closes #378.

## 👩‍💻 Reviewer Notes

- Just follow the usual procedure
- The story "With Empty Content" should be checked for naming and the used design tokens (`badge.withEmptyContent`).
- ~~We use a static `padding` value of **`7px`** in the component. This is done to make it look round when only one character is placed in the badge (e.g. `width` and `height` are both `22px` in this case). I opted to not use calc for this.~~ This is now adjusted in design and dev to use our tokens.
- The `warning` state in dark mode is currently not accessible. We know this and there are plans of fixing this issue, but doing it **may** not be part of this PR.

## 📑 Test Plan

- Check if everything looks right in the new Screenshots. Other Stories should not be affected at all.

## ✅ DoD


packages/components/src/components/badge/badge.component.ts
```diff
@@ -0,0 +1,54 @@
...
+          'badge--success': this.variant === 'success',
+          'badge--neutral': this.variant === 'neutral',
+          'badge--warning': this.variant === 'warning',
+          'badge--danger': this.variant === 'danger',
+        })}
+        role="status"
+        aria-label="${this.localize.term(`badge_${this.variant}`)}"
+      >
+        <slot></slot>
+      </span>
```
mariohamann, line 51:
```suggestion
      >
        <slot>
            ${this.localize.term(`badge_${this.variant}`)}</slot>
        </span>
```

The aria-label should only be there if no other content is provided, otherwise it overrides correct or adds unneeded information. Maybe it therefore is easier to provide default slots.

schilchSICKAG: This is actually a very nice idea. Seems to work fine

packages/components/src/translations/de.ts
```diff
@@ -37,7 +37,12 @@ const translation: Translation = {
   selectAColorFromTheScreen: 'Farbe vom Bildschirm auswählen',
   showPassword: 'Passwort anzeigen',
   slideNum: slide => `Folie ${slide}`,
-  toggleColorFormat: 'Farbformat umschalten'
+  toggleColorFormat: 'Farbformat umschalten',
+  badge_danger: 'Gefahr',
+badge_neutral: 'Neutral',
+badge_primary: 'Primär',
+badge_success: 'Erfolg',
+badge_warning: 'Warnung',
```
mariohamann, line 45: Translation keys shouldn't be component specific, as long as there translated term doesn't contain the component (e. g. `Close menu` could lead to the key `closeMenu` or similar).

Furthermore the texts should inform the user approriately, "Primär" doesn't give any information to screenreaders. Therefore some mapping has to be done on component side.

Here's my proposal, feel free to merge into your branch: https://github.com/synergy-design-system/synergy-design-system/pull/394

See especially https://github.com/synergy-design-system/synergy-design-system/pull/394/files#diff-0899acde5808332ea69599190c82ad6208b1704765348446d27c885f5f33ae01R52 and https://github.com/synergy-design-system/synergy-design-system/pull/394/files#diff-6e75b3b7229331821ff3bc44f3240f7385fad9d7d38c30f7094a92e1f8168890R42

mariohamann (approved):
Technically fine, there might be a little change regarding the spacings from design, but code-wise totally approved!

mariohamann (approved):

PR #406, opened 2024-04-05
docs: 📚 add syn-alert

LeaLeRose:
# Pull Request

## ✅ DoD

### DIFF
--- a/packages/tokens/src/figma-tokens/_docs.json
+++ b/packages/tokens/src/figma-tokens/_docs.json
@@ -77,7 +77,7 @@
           "type": "text"
         },
         "description": {
-          "value": "The toast stack is a fixed position singleton element created and managed internally by the alert component. It will be added and removed from the DOM as needed when toasts are shown. When more than one toast is visible, they will stack vertically in the toast stack.\n\nBy default, the toast stack is positioned at the top-right of the viewport. You can change its position by targeting .sl-toast-stack in your stylesheet. To make toasts appear at the top-left of the viewport, for example, use the following styles.",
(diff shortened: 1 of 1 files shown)


packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -1,13 +1,85 @@
...
+          "type": "text"
+        }
+      },
+      "toast-stack": {
+        "title": {
+          "value": "The Toast Stack",
+          "type": "text"
+        },
+        "description": {
+          "value": "The toast stack is a fixed position singleton element created and managed internally by the alert component. It will be added and removed from the DOM as needed when toasts are shown. When more than one toast is visible, they will stack vertically in the toast stack.\n\nBy default, the toast stack is positioned at the top-right of the viewport. You can change its position by targeting .sl-toast-stack in your stylesheet. To make toasts appear at the top-left of the viewport, for example, use the following styles.",
```
kirchsuSICKAG, line 80: `sl` should be changed to `syn`

```suggestion
          "value": "The toast stack is a fixed position singleton element created and managed internally by the alert component. It will be added and removed from the DOM as needed when toasts are shown. When more than one toast is visible, they will stack vertically in the toast stack.\n\nBy default, the toast stack is positioned at the top-right of the viewport. You can change its position by targeting .syn-toast-stack in your stylesheet. To make toasts appear at the top-left of the viewport, for example, use the following styles.",
```

kirchsuSICKAG (dismissed):

kirchsuSICKAG (approved):

PR #413, opened 2024-04-12
docs: 📚 add syn-dialog

LeaLeRose:
# Pull Request

## ✅ DoD

### DIFF
--- a/packages/tokens/src/figma-tokens/_docs.json
+++ b/packages/tokens/src/figma-tokens/_docs.json
@@ -367,13 +367,65 @@
       }
     },
     "dialog": {
-      "description": {
-        "value": "Dialogs, sometimes called \"modals\", appear above the page and require the user's immediate attention.",
-        "type": "text"
+      "default": {
+        "description": {
+          "value": "Dialogs, sometimes called \"modals\", appear above the page and require the user's immediate attention.",
+          "type": "text"
+        },
+        "title": {
+          "value": "Dialog",
+          "type": "text"
+        }
       },
-      "title": {
-        "value": "Dialog",
-        "type": "text"
+      "custom-width": {
+        "title": {
+          "value": "Custom Width",
+          "type": "text"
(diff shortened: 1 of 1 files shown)


packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -367,13 +367,65 @@
...
+          "type": "text"
+        }
+      },
+      "prevent-closing": {
+        "title": {
+          "value": "Preventing the Dialog from Closing",
+          "type": "text"
+        },
+        "description": {
+          "value": "By default, dialogs will close when the user clicks the close button, clicks the overlay, or presses the Escape key. In most cases, the default behavior is the best behavior in terms of UX. However, there are situations where this may be undesirable, such as when data loss will occur.\n\nTo keep the dialog open in such cases, you can cancel the sl-request-close event. When canceled, the dialog will remain open and pulse briefly to draw the user’s attention to it.\n\nYou can use event.detail.source to determine what triggered the request to close. This example prevents the dialog from closing when the overlay is clicked, but allows the close button or Escape to dismiss it.",
```
kirchsuSICKAG, line 416: `sl` should by changed to `syn`

```suggestion
          "value": "By default, dialogs will close when the user clicks the close button, clicks the overlay, or presses the Escape key. In most cases, the default behavior is the best behavior in terms of UX. However, there are situations where this may be undesirable, such as when data loss will occur.\n\nTo keep the dialog open in such cases, you can cancel the syn-request-close event. When canceled, the dialog will remain open and pulse briefly to draw the user’s attention to it.\n\nYou can use event.detail.source to determine what triggered the request to close. This example prevents the dialog from closing when the overlay is clicked, but allows the close button or Escape to dismiss it.",
```

kirchsuSICKAG (changes requested):

kirchsuSICKAG (approved):

PR #437, opened 2024-04-19
fix: 🤔 syn-text-area may throw an error when unmounted too quick

### LINKED ISSUE #435: fix: 🤔 syn-text-area may throw an error when unmounted too quick

- [ ] ~~I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...).~~

# Version

The bug occurred with the following version of the component library: 1.17.0

# Component

The issue arises when using the following component(s): syn-textarea

# Environment

# Description

When unmounting the `<syn-textarea />` from the DOM, there may be an error while cleaning up the the component instance. This stems from the fact that during `disconnectedCallback`, the component tries to unobserve one of its children (`this.input`) from the created `resizeObserver`. As the element is not in the `DOM` anymore, the whole `disconnectedCallback` fails with an `resizeObserver` exception.

This can be prevented if checking for existence of the input field. As this is is critical (e.g. for Karma Tests), I will add the fix in Synergy and also provide a PR at Shoelace.

# Acceptance criteria

- [ ] Textarea tests run fine in Karma

schilchSICKAG:
# Pull Request

## 📖 Description

This PR prevents an error that may appear when removing the `<syn-textarea />` too fast from the DOM. This may especially happen in test frameworks (e.g. Karma), where the current behaviour lead to an error `Uncaught TypeError: Failed to execute 'unobserve' on 'ResizeObserver': parameter 1 is not of type 'Element'. thrown`.

### 🎫 Issues

Closes #435

## 👩‍💻 Reviewer Notes

- I have opted to rewrite the vendorism script in the new style and also add two new utility functions to make it easier in the future: `replaceSection` will replace a section with the same syntax as `removeSection`, whereas `replaceSections` will do the same given an array of replacements.
- There where some dangling css classes that where copied left overs from the input vendorism (prefix and suffix). Those are removed
- The fix is applied, too ;)

## 📑 Test Plan

Unfortunately, I was not able to reproduce this problem on my local machine. However, I have already tested the implementation on a project where the issue arose and can say for sure it does not break after applying the fix.

## ✅ DoD

### DIFF
--- a/packages/components/scripts/vendorism/custom/textarea.vendorism.js
+++ b/packages/components/scripts/vendorism/custom/textarea.vendorism.js
@@ -15,7 +15,6 @@ const FILES_TO_TRANSFORM = [
  */
 const transformComponent = (path, originalContent) => {
   let content = removeSection(originalContent, '/** Draws a filled', 'filled = false;');
-  content = replaceSection(['filled', 'readonly'], content);
 
   content = replaceSections([
     ['filled', 'readonly'],
--- a/packages/components/scripts/vendorism/custom/textarea.vendorism.js
+++ b/packages/components/scripts/vendorism/custom/textarea.vendorism.js
@@ -1,40 +1,80 @@
 import { removeSection } from '../remove-section.js';
+import { replaceSection, replaceSections } from '../replace-section.js';
+
+const FILES_TO_TRANSFORM = [
+  'textarea.component.ts',
+  'textarea.styles.ts',
+  'textarea.test.ts',
+];
+
+/**
+ * Transform the component code
+ * @param {String} path
+ * @param {String} originalContent
+ * @returns
+ */
+const transformComponent = (path, originalContent) => {
+  let content = removeSection(originalContent, '/** Draws a filled', 'filled = false;');
+  content = replaceSection(['filled', 'readonly'], content);
+  return {
+    content,
+    path,
+  };
+};
+
+/**
+ * Transform the components styles
+ * @param {String} path
+ * @param {String} originalContent
+ * @returns
+ */
+const transformStyles = (path, originalContent) => {
+  const content = replaceSections([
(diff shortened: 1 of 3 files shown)


packages/components/scripts/vendorism/replace-section.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Replace a section of content with a new section
+ * @param {string[]} pair Tuple of search and replace strings
+ * @param {*} content The content to run the replacement in
+ * @returns {string} New content with replacements applied
+ */
+export const replaceSection = ([search, replace], content) => content.replaceAll(search, replace);
+
+/**
+ * Takes an array of section options and applies them via removeSection one by one
```
kirchsuSICKAG, line 10:
```suggestion
 * Takes an array of section options and applies them via replaceSection one by one
```

packages/components/scripts/vendorism/custom/textarea.vendorism.js
```diff
@@ -1,40 +1,92 @@
...
+
+/**
+ * Transform the component code
+ * @param {String} path
+ * @param {String} originalContent
+ * @returns
+ */
+const transformComponent = (path, originalContent) => {
+  let content = removeSection(originalContent, '/** Draws a filled', 'filled = false;');
+  content = replaceSection(['filled', 'readonly'], content);
```
kirchsuSICKAG, line 18: This line can be removed. 
The string is replaced in line 21
```suggestion
```

kirchsuSICKAG (changes requested):

packages/components/scripts/vendorism/custom/textarea.vendorism.js
schilchSICKAG: Leftover from refactoring :(

kirchsuSICKAG (approved):

PR #438, opened 2024-04-22
fix: 🤔 update dependencies

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR upgrades (most) packages to their latest stable version. Exceptions are:

- semantic-release (does not work with semantic-release-monorepo@7.x)
- semantic-release-monorepo (Patch does not apply for current 7.0.8 or 8.0.0)
- wc-storybook-helpers (We patched it currently and need to have further investigation, if the patch is still needed etc.)
- @tokens-studio/sd-transforms (the new version depends on a prerelease version `^4.0.0-prerelease.22` of `style-dictionary`)
- eslint (got a major update, but some used eslint plugins do not yet support eslint 9 and also it would need more investigation, because the whole configuration of esliint changed)

### 🎫 Issues

Closes #258 

## 👩‍💻 Reviewer Notes

As we have substantial updates to nearly every package, everything has to be checked (e.g. linting, building, tests, framework wrappers, docs).

## 📑 Test Plan

- Check out this branch and run pnpm i -r && pnpm build && pnpm lint. There should be no errors
- Make sure icons are drawn via cd packages/assets && pnpm fetch-assets. There should be no diffs between icons!
- Check the components package via cd packages/components && pnpm vendor.get && pnpm vendor.set && pnpm build && pnpm test && pnpm lint. There should be no issues, no new files in components or the wrappers.
- Check if Docs work via cd packages/docs && pnpm start. Navigate around and see if there are any issues arising in the started application.
- Check if the framework wrappers are still working by starting the local demos and playing around with them.

## ✅ DoD


schilchSICKAG (commented):
Most of the changes are fine. I was able to find another issue that I already fixed in this branch (the tokens scss values where never shown correct as they use the JS notation, not the css variable notation). I also updated to the latest versions (vue and storybook respectively) and let all tests and demo projects run. Works fine :).

However why did we do not an update to eslint@9 in this step? We also have to check if wc-storybook-helpers can be patched again with its new major version, however this seems to be out of scope for this ticket.

PR #447, opened 2024-04-23
fix: 🤔 chevron of syn-nav-item using wrong symbol

### LINKED ISSUE #432: fix: 🤔 chevron of syn-nav-item using wrong symbol
# Version

The bug occurred with the following version of the component library: 1.16.0

# Component

The issue arises when using the following component(s): syn-nav-item

# Description

Currently the chevron is either ^ or ⌄ . Instead it should be  > when acting as a link

# Acceptance criteria

- [ ] Have a story in place
- [ ] Use the right arrow styles

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue that showed the wrong chevron position was used for `<syn-nav-items>` when having both a `href` and `chevron` attribute set.

Old behavior:

old

New behavior:

new

### 🎫 Issues

Closes #432 

## 👩‍💻 Reviewer Notes

The change does not have an associated story.

## 📑 Test Plan

- Check out this branch
- Build world and start storybook
- Move to the docs page of `<syn-nav-item>`
- Set a `href` attribute and set `chevron` to true

## ✅ DoD

### DIFF
--- a/packages/components/src/components/nav-item/nav-item.styles.ts
+++ b/packages/components/src/components/nav-item/nav-item.styles.ts
@@ -145,7 +145,7 @@ export default css`
   }
 
   /**
-   * The chevron indicates the use as a <details /> element
+   * The chevron indicates the use as a <details /> element OR a link
    */
   .nav-item__chevron {
     font-size: var(--syn-font-size-x-large);
@@ -156,6 +156,13 @@ export default css`
 
   .nav-item__chevron-open {
     rotate: -180deg;
+  }
+
+  /**
+   * Links should always show a chevron pointing to the right
+   */
+  .nav-item--is-link .nav-item__chevron {
+    rotate: -90deg;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/components/src/components/nav-item/nav-item.styles.ts
```diff
@@ -158,6 +158,13 @@ export default css`
     rotate: -180deg;
   }
 
+  /**
+   * Links should always show a chevron pointing to the right
+   */
+  .nav-item--is-link .nav-item__chevron {
+    rotate: -90deg;
+  }
```
kirchsuSICKAG, line 166: Don't we only want the chevron to show downwards / upwards in case of its an accordion? Right now, if not having a link and using the chevron attribute it would show upwards / downwards depending of the open attribute state.
image

For example if someone is using the Angular `routerLink` i would assume the chevron wpuld also show to the right.

schilchSICKAG: Fixed with latest push. Please verify :)

kirchsuSICKAG (approved):

PR #451, opened 2024-04-24
feat: ✨ tooltip

### LINKED ISSUE #382: feat: ✨ tooltip
# Story
As a user of the Synergy Design System, I would like to see additional information (hidden until action is executed) based on a specific action (e.g. hover, click, etc.)

# Acceptance criteria

- [ ] Incorporates slots, Events, Methods, Parts from Shoelace
- [ ] Create a proper documentation which contains at least everything from the Shoelace component
- [ ] Text in the documentation can be synced

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-tooltip />` element.

### 🎫 Issues

Closes #382.

## 👩‍💻 Reviewer Notes

- I removed all unused shoelace fallback tokens from `/packages/tokens/src/shoelace-fallbacks`.
Please have a look, that nothing changed at the output of the build `light.css` and `dark.css` files.

- the `placement` story does show up a11y violations and the other stories have incompletes. These can be ignored, because they don't belong to the tooltip. but to the used buttons

## 📑 Test Plan

- have a look at the `light.css` and `dark.css` output of the tokens package (for more information see the Reviewer notes section)
- Have a look at the new Components -> syn-tooltip stories

## ✅ DoD


packages/components/src/components/tooltip/tooltip.custom.styles.ts
```diff
@@ -0,0 +1,8 @@
+import { css } from 'lit';
+
+export default css`
```
schilchSICKAG, line 3: We have new tokens exactly for the tooltip. Shouldn`t we override the current style configuration to use them?

packages/docs/stories/components/tooltip.stories.ts
```diff
@@ -0,0 +1,312 @@
...
+      position: relative;
+      border: solid 2px var(--syn-panel-border-color);
+      overflow: hidden;
+      padding: var(--syn-spacing-medium);
+    }
+  </style>
+  `,
+};
+
+// Bundled screenshot story
```
schilchSICKAG, line 302: I tend to silence the warnings like this:


```suggestion
/* eslint-disable sort-keys */
```

packages/components/src/components/tooltip/tooltip.custom.styles.ts
kirchsuSICKAG: They are already used in the tooltip.styles.ts

packages/docs/stories/components/tooltip.stories.ts
```diff
@@ -0,0 +1,320 @@
...
+      overflow: hidden;
+      padding: var(--syn-spacing-large);
+    }
+  </style>
+  `,
+};
+
+/* eslint-disable sort-keys */
+export const Screenshot: Story = generateScreenshotStory({
+  Default,
```
schilchSICKAG, line 304: Does it make sense to show the default story here? It just triggers on hover, so it shows a button only

packages/docs/stories/components/tooltip.stories.ts
```diff
@@ -0,0 +1,320 @@
...
+        </syn-tooltip>
+
+        <syn-tooltip content="top-end" placement="top-end" open>
+          <syn-button></syn-button>
+        </syn-tooltip>
+      </div>
+
+      <div class="tooltip-placement-example-row">
+        <syn-tooltip content="left-start" placement="left-start" open>
+          <syn-button></syn-button>
```
schilchSICKAG, line 103: When adding trigger="manual", the items don´t close when hovering. I think this would be a fine addition for viewing in Storybook, what do you think?


```suggestion
        <syn-tooltip content="left-start" placement="left-start" trigger="manual" open>
```

packages/docs/stories/components/tooltip.stories.ts
```diff
@@ -0,0 +1,320 @@
...
+      },
+      story: {
+        height: '450px',
+      },
+    },
+    layout: 'padded',
+  },
+  render: () => html`
+    <div class="tooltip-placement-example">
+      <div class="tooltip-placement-example-row">
```
schilchSICKAG, line 87: The issue seems to go away when using `overflow: hidden` on the placement example. However, in this case the container is not wide enough anymore to show the left-start and left-end examples in their "natural" placement. Mind to play around with this a bit?


```suggestion
    <!-- Not those are just values to play with! -->
    <div class="tooltip-placement-example" style="height: 400px; overflow: hidden;">
```

packages/docs/stories/components/tooltip.stories.ts
kirchsuSICKAG: I open all tooltips via the `open` prop. Therefore all tooltips are open on start for the screenshot

packages/docs/stories/components/tooltip.stories.ts
kirchsuSICKAG: This is a problem shoelace also has in its documentation. 
I think the problem appears because the trigger is set on `hover` per default (and I do not change it) but I open all tooltips via the `open` prop.
Therefore i would assume there is no problem with the tooltips themselves.
I spoke with Stefan and we decided we won`t investigate more in it, because it could only be a problem of our documentation.

If you`d like to investigate more, you can have a look at it.

packages/docs/stories/components/tooltip.stories.ts
kirchsuSICKAG: What you mean is  trigger="click"  so it will just close and open on click.  I think its okay like this. But I don`t mind if we change the behavior from `hover` to `click`

packages/docs/stories/components/tooltip.stories.ts
schilchSICKAG: I am actually fine with this :)

packages/docs/stories/components/tooltip.stories.ts
schilchSICKAG: Then I am fine, too :)

schilchSICKAG (approved):

PR #456, opened 2024-04-30
fix: 🤔 Syn-Checkbox uses wrong line-height when in multi-line mode

### LINKED ISSUE #298: fix: 🤔 Syn-Checkbox uses wrong line-height when in multi-line mode

# Version

The bug occurred with the following version of the component library: 1.6.0 (probably since 1.1.0)

# Component

The issue arises when using the following component(s): `<syn-checkbox />`, possibly also `<syn-switch />` (not tested)

# Environment

# Description

Our checkbox component uses a fixed `line-height` for the label of checkboxes (and possibly switch) that matches the `height` of the item. This leads to an error when using multi line labels, where the font-size and line-height are set the same per default (e.g. `font-size: 16px; line-height: 16px`. This results in broken typographic styles.
See attached image:

checkbox-multiline

# Acceptance criteria

- [ ] Single Line Checkboxes and Switches work like before
- [ ] Multi line checkboxes have spacing between the text lines. (To be defined by @LeaLeRose)

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue that lead to broken line-heights for the label of `<syn-checkboxes>` when the label was displayed in multiple lines.

### 🎫 Issues

Closes #298 

## 👩‍💻 Reviewer Notes

The issue stems from the fact that Shoelace uses a fixed line-height value for the label that is set to the size of the checkboxes checkmark icon. When using a checkbox that has a multi-line label, this lead to the problem that spacing between the lines was too narrow.

A word on tokens: I was not able to use our token values for this as the margins are too narrow to tackle them this way.

## 📑 Test Plan

1. Verify the bug exists in the current version of Synergy (e.g. in our own form-template at https://synergy-design-system.github.io/?path=/story/templates-forms--contact-form&globals=viewport:default)
2. Have a look at the storybook version of this PR. It should show a line-height of `1.4`.

## ✅ DoD

### DIFF
--- a/packages/components/src/components/checkbox/checkbox.custom.styles.ts
+++ b/packages/components/src/components/checkbox/checkbox.custom.styles.ts
@@ -29,7 +29,21 @@ export default css`
   }
 
   .checkbox__label {
+    line-height: var(--syn-line-height-normal);
     margin-inline-start: var(--syn-spacing-x-small);
+    position: relative;
+  }
+
+  .checkbox--small .checkbox__label {
+    top: -3px;
+  }
+
+  .checkbox--medium .checkbox__label {
+    top: -3px;
+  }
+
+  .checkbox--large .checkbox__label {
+    top: -2px;
   }
 
   /* Disabled */


kirchsuSICKAG:
We also have this problem with the radio component. If a multiline text is used there, we also get a compressed text with the same line-height as the font-size: 
image

Adding the same css changes there, it would look like following: 
Bildschirmfoto 2024-05-08 um 10 01 23

```css
  .radio__label {
    line-height: var(--syn-line-height-normal);
    position: relative;
  }

  .radio--small .radio__label {
    top: -3px;
  }

  .radio--medium .radio__label {
    top: -3px;
  }

  .radio--large .radio__label {
    top: -2px;
  }
  ```

kirchsuSICKAG (approved):

PR #458, opened 2024-04-30
feat: ✨ syn-dialog

### LINKED ISSUE #443: feat: ✨ syn-dialog
# Story

As a user of the Synergy Design System, I would like to display information to the end-user in a very focused way. The end-user has to confirm the message so he can continue working.

# Acceptance criteria

- [ ] All kind of information (text, images, etc.) can be displayed in a overflowing container
- [ ] The box consists of a header, a main section and a footer
- [ ] The footer can be used to confirm the message with a button
- [ ] Keyboard input can be used to close the dialog (ESC -> reject, Enter -> confirm)

### LINKED ISSUE #459: chore: 🔧 adapt spacing for syn-dialog header
# Description

To have a uniform design, we are adjusting the inner spacing of the syn-dialog header. previously: none, in future: spacing/large 24

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-dialog>` component, as well as tests and stories for how to use it. It also adjusts the spacing in `<syn-drawer>` to match the `<syn-dialog>`.

### 🎫 Issues

Closes #443 
Closes #459

## 👩‍💻 Reviewer Notes

Just follow the regular procedure. Also make sure that the component is working in our demo projects.

## 📑 Test Plan

- Verify the stories do not have any errors in storybook
- Verify the visual changes to `<syn-drawer>`
- Verify the component matches the provided figma component

## ✅ DoD


klaarseSICKAG:
sometimes the dialog won't close when pressing the "close" button. X-button, ESC and clicking outside always works.

putting another button in the footer also doesn't work.

schilchSICKAG:
> sometimes the dialog won't close when pressing the "close" button. X-button, ESC and clicking outside always works.
> 
> putting another button in the footer also doesn't work.

This is now fixed in the latest version. The reason for this was that updating the slot recreated only the rendered `<syn-dialog>` in the story, but not the button itself. Because of this, the reference to the original `<syn-dialog>` was missing. The listener is now always working when updating slots, even the footer slot that holds the listener.

packages/components/src/components/drawer/drawer.custom.styles.ts
```diff
@@ -67,7 +67,7 @@ export default css`
    */
   .drawer__title,
   .header-actions {
```
kirchsuSICKAG, line 69: I just noticed, that we have a wrong className here. We don't have any `header-actions` class in the drawer component. Therefore the `min-height` is never added somewhere. Should be `drawer__header-actions` instead

 ```suggestion
  .drawer__header-actions {
```

File: packages/components/src/components/dialog/dialog.component.ts [hunk 2]
```diff
@@ -0,0 +1,362 @@
...
+      this.originalTrigger = document.activeElement as HTMLElement;
+      this.modal.activate();
+
+      lockBodyScrolling(this);
+
+      // When the dialog is shown, Safari will attempt to set focus on whatever element has autofocus. This can cause
+      // the dialogs's animation to jitter (if it starts offscreen), so we'll temporarily remove the attribute, call
+      // `focus({ preventScroll: true })` ourselves, and add the attribute back afterwards.
+      //
+      // Related: https://github.com/synergy-design-system/synergy/issues/693
```
kirchsuSICKAG, line 179: Just a note on something we should look at at some point:
Here and on other components (e.g. select, input, ...) we have a link to an github issue. But this link is not correct, because we overwrite all shoelace occurrences. Should be something `https.//github.com/shoelace/shoelace....`

Maybe we need to overwrite it with vendorism again in all components, where we have this problem. Or we'll just ignore it.
What do you think?

packages/components/src/components/dialog/dialog.custom.styles.ts
```diff
@@ -0,0 +1,54 @@
+import { css } from 'lit';
+
+export default css`
+  :host {
+    /**
+     * Used as the default padding for the drawer title
```
kirchsuSICKAG, line 6: Copy paste

```suggestion
     * Used as the default padding for the dialog title
```

packages/components/src/components/dialog/dialog.custom.styles.ts
```diff
@@ -0,0 +1,54 @@
...
+    gap: var(--syn-spacing-x-small);
+    padding: var(--syn-spacing-large) var(--syn-spacing-small) var(--syn-spacing-large) 0;
+  }
+
+  /**
+   * We need this to make sure we are big enough as defined in the layout
+   * When omitting this statement, we will get rounding problems via line-height
+   */
+  .dialog__title,
+  .header-actions {
```
kirchsuSICKAG, line 38: Same as with the drawer. We don't have any `header-actions` class in the dialog component. Therefore the min-height is never added somewhere. Should be `dialog__header-actions` instead

```suggestion
  .dialog__header-actions {
```

packages/docs/stories/components/dialog.stories.ts
```diff
@@ -0,0 +1,261 @@
...
+  <script type="module">
+    const createOpener = (opener) => {
+      // Storybook only: When loading the docs page, all dialogs are applying a focus trap.
+      // Remove the initial trap and make sure to do the same when recreating the story.
+      const loadedDialog = opener.parentElement.querySelector('syn-dialog');
+      loadedDialog.modal.activateExternal();
+
+      opener.addEventListener('click', (e) => {
+        const currentDialog = e.target.parentElement.querySelector('syn-dialog');
+        console.log(currentDialog)
```
kirchsuSICKAG, line 51: the console.log should be removed

```suggestion
```

kirchsuSICKAG (changes requested):

packages/components/src/components/dialog/dialog.component.ts
schilchSICKAG: I have created a ticket for this at https://github.com/synergy-design-system/synergy-design-system/issues/471.

packages/components/src/components/drawer/drawer.custom.styles.ts
schilchSICKAG: Fixed && thanks :)

PR #466, opened 2024-05-03
docs: 📚 AppShell templates

### LINKED ISSUE #316: docs: 📚 AppShell template(s)
# Description

As a user of synergy I want to have an easy start with an application which looks like a typical SICK app without spending too much time.

# Rationale

Show how to orchestrate a typical application with app-shell components in Storybook.
If meaningful provide more than one template: e.g. for a fixed sidebar, dynamic sidebar, with and without top-bar navigation, with and without step-navigation

# Acceptance criteria
- [x] template for vanilla, vue, react, ng
- [x] 5 different variants
- [x] Have a storybook demos
- [x] template should contain header, top-bar navigation, side-bar, footer (text only)
- [x] SICK logo should be in the header
- [x] Shrinked contend

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds new templates for our new application shell components to show usage in applications.

### 🎫 Issues

Closes #316 

## 👩‍💻 Reviewer Notes

- The current version is now based on the current Figma dev variants as seen at https://www.figma.com/file/bZFqk9urD3NlghGUKrkKCR/branch/dwSXnI3fvvqOF9VGg4OpTZ/Synergy-Digital-Design-System?node-id=16081-38188.
- ~~I was not able to force a mobile mode in Storybook. It seems you can only adjust the viewport, but not enable touch support. Therefore, the story "Rail Navigation (On Touch-click)" could not be created :(.~~ The mobile rail mode is now available as a mobile story. It shows how to create a mobile version of the rail navigation. However, clicking is still not possible. Also, the width of the story does not match storybook. We would have to configure the storybook views addon to make this work. Maybe we should switch to some sensitive defaults for common device sizes here and in Figma?
- The structure for template documentation tokens was wrong, as it used `templates.default` for the tokens of the form demos. I have adjusted this into `templates.contact-form` and create `templates.application-shell`. We have to decide on a text for this. Also, `templates.templates` is not used in Storybook, I think this is only for Figma. I left it alone without changing it as I do not know what a change there might cause :). Please have a look at the provided `_docs.json` file in the tokens package and adjust the tokens on Figma side after merging.
- 🤔 I would also recommend to use english text for the footer, as the rest of the applications navigation items also use english terms. However the current version uses german translations for the footer, as seen in the current version in Figma.
- The template in Figma is called "App Shell Templates". I found this to be redundant in Storybook as it resides in "Templates". Therefore, I opted for the name "AppShell".
- ~~🤔 Do we want to create automatic tests for these templates? More or less, they are copies from the `syn-side-nav` stories, so we have to decide. If we create screenshots, should we show the side-nav in its opened state?~~ -> During talking about this with @klaarseSICKAG, we have decided to not screenshot the templates as most of the code is already taken from the `<syn-side-bar>` stories.

## 📑 Test Plan

- Check the new Storybook stories ad Templates -> AppShell.

## ✅ DoD


klaarseSICKAG (approved):

klaarseSICKAG (dismissed):

kirchsuSICKAG:
I just noticed something with our side-nav as I was playing around with the app shell template...
It is not a problem with this branch, but a general problem we have.
When using tab to focus the syn-nav-items in a syn-side-nav we are getting a horizontal scrollbar in the side-nav

And the left and right border of the focus is also not visible:
image

kirchsuSICKAG (changes requested):

packages/docs/stories/templates/appshell.stories.tsx
```diff
@@ -0,0 +1,502 @@
...
+ */
+const createSharedStyles = () => html`
+  <style>
+    body {
+      margin: 0 !important;
+      padding: 0 !important;
+    }
+
+    syn-side-nav {
+      --side-nav-open-width: 320px;
```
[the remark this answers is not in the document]
kirchsuSICKAG: I created a ticket for this: https://github.com/synergy-design-system/synergy-design-system/issues/479

packages/docs/stories/components/header.stories.ts
```diff
@@ -158,6 +158,27 @@ export const Navigation = {
       </syn-prio-nav>
       App Name
     </syn-header>
+    <script type="module">
+      // This emulates a click on the side-nav and updates the main content
```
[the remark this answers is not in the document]
schilchSICKAG: Fixed

packages/docs/stories/templates/appshell.stories.tsx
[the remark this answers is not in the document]
schilchSICKAG: Ok, get it. In this case I would remove the custom width here and wait for the bugfix in the ticket.

packages/docs/stories/templates/appshell.stories.tsx
```diff
@@ -0,0 +1,502 @@
...
+
+/**
+ * Create a header with optional prio-nav
+ */
+const createHeader = ({
+  withMetaNavigation = false,
+} = {}) => html`
+  <!-- header -->
+  <syn-header>
+  ${getTranslation('appShell.appName')}
```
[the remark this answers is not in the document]
schilchSICKAG: When using the label it will still fail as there is other content in the default slot (which is the reason for the whole 2.x release anyways ;)). I can only fix this when 2.x is available. If we get the template to be merged before the 2.x release it should not be that much of a problem as we can adjust before 2.x. If it comes after, we will have to merge it later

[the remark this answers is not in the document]
kirchsuSICKAG: Then I´m fine with it. We just need to keep in mind to change it as soon as we merge the 2x. branch

packages/docs/stories/components/side-nav.stories.tsx
```diff
@@ -82,24 +109,25 @@ export const Default: Story = {
...
+        const selector = 'story-loaded-'.concat(index);
+        if(!mainContent.classList.contains(selector)) {
+          const header = mainContent.previousElementSibling;
+          const sideNav = mainContent.querySelector('syn-side-nav');
+          header.connectSideNavigation(sideNav);
+          mainContent.classList.add(selector);
+        }
+      });
+    </script>
+    ${createDummyActiveNavItemListener()}
```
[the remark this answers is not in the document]
schilchSICKAG: Really only tested it on the docs page instead of the single stories ^^. It is built other ways as the other stories, so I was happy there was just a single entry to do and it worked. Should have looked better, its fixed now.

kirchsuSICKAG (approved):

kirchsuSICKAG (approved):

packages/docs/stories/templates/appshell.stories.tsx
```diff
@@ -0,0 +1,503 @@
...
+/**
+ * Create a side navigation with an optional rail mode
+ */
+const createSideNav = ({
+  open = false,
+  rail = false,
+} = {}) => html`
+  <!-- side-navigation -->
+  <syn-side-nav
+    no-focus-trapping
```
mariohamann, line 152: This shouldn't be added to every example! IT shouldn't be ther ee. g. for the first one.

packages/docs/stories/templates/appshell.stories.tsx
```diff
@@ -0,0 +1,503 @@
...
+          display: none;
+        }
+
+        .synergy-demo-content-inner {
+          /**
+           * Set the initial margin left. Needed because we
+           * create the page with the side-bar open
+           */
+          margin-left: var(--appshell-shrink-nav-open-width);
+          transition: margin-left 250ms;
```
mariohamann, line 372: This should only happen if reduced motion is not active. Please make a query for that.

mariohamann (changes requested):

packages/docs/stories/templates/appshell.stories.tsx
mariohamann: In addition @kirchsuSICKAG  we have a problem here: The header should be reachable, as it's not affected by the overlay. Otherweise you can't reach the close button anymore. We should provide a fix for that and write a new issue.

packages/docs/stories/templates/appshell.stories.tsx
schilchSICKAG: I just had a look as I was not able to find a call to `prefers-reduced-motion` in our other animations. It seems that shoelace handles this with its `animate` js function, but not when using inline transforms or transitions in css (e.g. in checkbox there are fixed transitions without `prefers-reduced-motion`, too).

I would add it here, but we also have to think about the other uses of inline transitions.

packages/docs/stories/templates/appshell.stories.tsx
schilchSICKAG: The reason for this is that we do not use iframes on the the docs page. When using the iFrame approach, dark mode would not work. Will have to check if it is possible in another way, but having "global" components as the shell or dialog without killing the focus trapping will lead to strange behavior when navigating the docs overview pages.

packages/docs/stories/templates/appshell.stories.tsx
mariohamann: It's about _reduced_ motion, not removing _every_ motion.

Currently the drawer doesn't have a motion effect, when you activate "Reduce motion" in the settings:

CleanShot 2024-05-16 at 09 50 09

The problem here is: The drawer skips the transition, while the content always does the transition, which doesn't fit together here. :) So we have to adapt the template, that it works together from a11y perspective.

schilchSICKAG: Thats exactly what I meant. I did a quick search and saw that shoelace does this in its own animate function, but not in its custom css transitions, which we will also have to check as our own components may also animate (which may be unwanted).

packages/docs/stories/templates/appshell.stories.tsx
schilchSICKAG: Easiest solution would be to remove the prop and set the items into iFrames. However, some stuff (like dark mode) may not work in this case. Like I said, I will check this. Its unfortunate that stuff like settings attributes or other plugins do not work in the docs overview when using "real" isolation (yet) :(. It also leads to bleeding, which is why I would love to have the iframe approach everywhere. This is also why I prefer the story approach instead of the overview, but this seems to be personal taste :)

packages/docs/stories/templates/appshell.stories.tsx
schilchSICKAG: Done in latest

packages/docs/stories/templates/appshell.stories.tsx
schilchSICKAG: You are right, the focus handling is kind of strange currently. There seems to be a hidden element that is focused when using the navigation. Also, it does not honor the close icon in the header. Will create a ticket for this. It does seem to be related to the sidebar/header components as it also appears to be a problem in the default sidebar stories.

The main problem (adding `no-focus-trapping` on everything) is now fixed.

schilchSICKAG: The problem is caused by [shoelaces drawer implementation](https://github.com/shoelace-style/shoelace/blob/next/src/components/drawer/drawer.component.ts#L342) that adds the panel as focusable element. The same can be seen at the Shoelace demos of `<sl-drawer>`. Will open a ticket and see what we can do there.

schilchSICKAG: https://github.com/shoelace-style/shoelace/issues/2018

packages/components/src/components/nav-item/nav-item.styles.ts
```diff
@@ -253,7 +253,7 @@ export default css`
   }
 
   .nav-item--horizontal:hover .current-indicator--visible,
-  .nav-item--horizontal:focus .current-indicator--visible {
```
mariohamann, line 256: Because of the commit message this change won't be promoted in the changelog. This should be a separate PR. (Sorry.)

mariohamann (dismissed):
Codewise I'm fine – besides my comment regarding the change in the styles, as this won't fit the release process.

packages/components/src/components/nav-item/nav-item.styles.ts
schilchSICKAG: Its now removed and will be handled in #486 instead.

mariohamann (approved):

PR #470, opened 2024-05-07
feat: ✨ syn-breadcrumb

### LINKED ISSUE #463: feat: ✨ syn-breadcrumb
# Story

As a user of the Synergy Design System, I would like to display the current navigation trail on the page.

# Acceptance criteria

- [x] Components `<syn-breadcrumb>` and `<syn-breadcrumb-item>` are provided
- [x] Allows to set a prefix or suffix on the `<syn-breadcrumb-item>`
- [ ] Add a template that shows how dropdowns can be used in the middle of a `<syn-breadcrumb>`
- [x] Allow to change the separator between breadcrumb items
- [ ] Adjust documentation to make sure parts (e.g. separator) should not be changed (TBD: "Do NOT change in SICK Applications")

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-breadcrumb>` and `<syn-breadcrumb-item>` components, including tests and stories.

### 🎫 Issues

Closes #463 

## 👩‍💻 Reviewer Notes

- I had to remove the accessibility aria checks in unit tests as they checked the color contrast of the last item (which defaults to a light grey). This was discussed with @LeaLeRose and is ok to do as the last item is drawn as an "inactive" one.

## 📑 Test Plan

Proceed as usual: Check the committed files, check Storybook, test in framework projects.

## ✅ DoD


klaarseSICKAG:
- The hover-state is not applied to the prefix suffix icons
- The already selected element is still interactive. It this right?

packages/components/src/components/breadcrumb-item/breadcrumb-item.custom.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+  :host(:not(:last-of-type)) .breadcrumb-item__label {
+    color: var(--syn-typography-color-text);
+  }
+
+  :host(:not(:last-of-type)) .breadcrumb-item__label:hover {
+    color: var(--syn-color-primary-600);
+  }
+
+  :host(:not(:last-of-type)) .breadcrumb-item__label:active {
+    color: var(--syn-color-primary-950);
```
kirchsuSICKAG, line 27: The `active` styling / color can't be differentiated that well vs the normal styling / color: 
See left normal and right active:
image

In the current figma version the `active` color is neutral-500

packages/components/src/components/breadcrumb-item/breadcrumb-item.custom.styles.ts
```diff
@@ -0,0 +1,70 @@
...
+    font-size: var(--syn-font-size-x-small);
+    font-weight: var(--syn-font-weight-normal);
+  }
+
+  /**
+   * Make sure to remove the border radius.
+   * Without this, focus will be rounded
+   */
+  .breadcrumb-item__label {
+    border-radius: var(--syn-border-radius-none);
```
kirchsuSICKAG, line 15: Focus from shoelace and the focus of the current figma version differ: 
Figma: 
image

Shoelace / current dev version:
image

packages/components/src/components/breadcrumb-item/breadcrumb-item.custom.styles.ts
schilchSICKAG: I think this is not the actually `:active` pseuso-state, but the active element (e.g. the last one). @LeaLeRose: Could you give some information here?

packages/components/src/components/breadcrumb-item/breadcrumb-item.custom.styles.ts
schilchSICKAG: This is unfortunately not possible to to, because Shoelace does not add the prefix- or suffix-slots to the surrounding link or button element. It would also not work when using `syn-dropdown` as slot content, as in this case everything would be need to be part of the dropdown.

schilchSICKAG:
Hi @klaarseSICKAG,

> * The hover-state is not applied to the prefix suffix icons
Yes, this is correct as the link/button is just the "middle" part of the component. There is currently no way to also add the link to the prefix or suffix slots. We could render the link tag around those tags, but this would lead to a lot of additinal problems:

- You would not be able to set custom links or components requiring focus in the slots as the outer link would have precedence.
- It would not work with the dropdown in the middle slot, as we would need to add the prefix/suffix slot to the dropdown instead, leading to different rendering.

> * The already selected element is still interactive. It this right?
This is the way shoelace does it. Basically, they just apply a different style to the last element, but there is nothing preventing you from clicking a link. However, the user does have the possibility to render the last item without link, for example by iterating over a navigation array like structure and just leaving the href attribute out.

packages/components/src/components/breadcrumb-item/breadcrumb-item.custom.styles.ts
kirchsuSICKAG: I think you are right.

But I think we should add an :active state for pressing down a breadcrumb-item, which is more visible. Especially after the statement of our stakeholder, who said he misses a visual pressed state for some components. Or we postpone this topic to a later time and we add the pressed / active state for all components simultaniously.

schilchSICKAG: The active one is the last item in the breadcrumb list, so it works as intended

kirchsuSICKAG (approved):

kirchsuSICKAG (approved):

PR #496, opened 2024-05-27
feat: ✨ syn-tab

### LINKED ISSUE #465: feat: ✨ syn-tab, syn-tab-group, syn-tab-panel
# Story

As a user of the Synergy Design System, I want to have the possibility to show an arbitary list of information items in a clear and concise way by showing only the active tabs content.

# Acceptance criteria

- [ ] A `<syn-tab>` always links to its corresponding `<syn-tab-panel>`
- [ ] All kind of information (text, images, etc.) can be displayed in a `<syn-tab-panel>`.
- [ ] Multiple `<syn-tab>` / `<syn-tab-panel>` components can be grouped into one `<syn-tab-group>`
- [ ] All possible items must be overridable (e.g. via part or slots)

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-tab>`, `<syn-tab-panel>` and `<syn-tab-group>` components.

### 🎫 Issues

Closes #465.

## 👩‍💻 Reviewer Notes

- I needed to add the `contained`, `sharp` and `placement` property of the tab-group also to the tab, so some stylings (like rounded borders for the focus) could be achieved.
- I added the css variable `--indicator-width` to the tab-group. Previously, the indicator (blue active tab line) had the same width as the track (gray border under the tabs). Due to the new design, both lines have different widths and therefore I added the new css variable

### Open question: 
- the track width of the tab-panel can be changed via exposed css variable. Should this track width be used to draw the border of the `contained` tab group, so something like this is possible for the user, or don't we want this to be changeable this easy?: 
image

## 📑 Test Plan

have a look at:
- the changes
- storybook
- the new components with the framework wrappers

## ✅ DoD


mariohamann:
@kirchsuSICKAG The border with should use https://shoelace.style/tokens/more#panels `--syn-panel-border-width`

packages/components/scripts/vendorism/custom/tab.vendorism.js
```diff
@@ -0,0 +1,118 @@
...
+ * @param {String} originalContent
+ * @returns
+ */
+const transformComponent = (path, originalContent) => {
+  let content = addSectionsAfter([
+    // Add prefix slot documentation
+    [
+      '* @slot - The tab\'s label.',
+      ' * @slot prefix - Used to prepend an icon or similar element to the tab.',
+    ],
```
mariohamann, line 22: Can we try to bring that to Shoelace? I'm realizing here, that we introduce this here – in fact this could be easily done inside the main slot + CSS as well, or what do you think?

packages/components/scripts/vendorism/replace-section.js
mariohamann, line 1: How this thing evolves... 😆

packages/components/src/components/icon/library.system.ts
```diff
@@ -18,6 +18,8 @@ const icons = {
 		<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" fill='currentColor'><path d="M8.79496 15.875L4.62496 11.705L3.20496 13.115L8.79496 18.705L20.795 6.70498L19.385 5.29498L8.79496 15.875Z"/></svg>`,
 	'chevron-down': `
 		<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" fill='currentColor'><path d="M16.59 8.29498L12 12.875L7.41 8.29498L6 9.70498L12 15.705L18 9.70498L16.59 8.29498Z"/></svg>`,
+	'chevron_right': `
```
mariohamann, line 21:
```suggestion
	'chevron-right': `
```

We have `chevron-down`

kirchsuSICKAG:
> --syn-panel-border-width



> @kirchsuSICKAG The border width should use https://shoelace.style/tokens/more#panels `--syn-panel-border-width`

I changed to all syn-panel tokens for the contained style (--syn-panel-border-width, --syn-panel-border-color, --syn-panel-background-color)

packages/components/scripts/vendorism/custom/tab.vendorism.js
mariohamann: Yah, please talk with Lea. I think we can easily just show it in a template.

packages/components/src/components/icon/library.system.ts
mariohamann: I did – as well for `more-vert` and `chevron-right`.

mariohamann:
> > --syn-panel-border-width
> 
> > @kirchsuSICKAG The border width should use https://shoelace.style/tokens/more#panels `--syn-panel-border-width`
> 
> I changed to all syn-panel tokens for the contained style (--syn-panel-border-width, --syn-panel-border-color, --syn-panel-background-color)

Perfekt. @LeaLeRose Do we use those in Figma as well?

packages/components/src/components/icon/library.system.ts
kirchsuSICKAG: I updated the system icons

packages/components/scripts/vendorism/custom/tab.vendorism.js
kirchsuSICKAG: Done

reichst1SICKAG:
I want to add my coment to the open question: 
It is not that easy to change the border-width in Figma and also for consistency and from a design perspective the track width should be removed from the css properties.

klaarseSICKAG:
toggling the active attribute doesn't seem to have an effect

klaarseSICKAG:
the documentation of syn-tab-show and syn-tab-hide doesn't say anything about what the payload contains. only what type it is (string)

klaarseSICKAG:
please review migration guide: https://gitlab.sick.com/platforms/davinci/synergymigrationguide/-/merge_requests/1

kirchsuSICKAG:
> toggling the active attribute doesn't seem to have an effect

@klaarseSICKAG 
Yes, this is right, but we can`t do much here about it. 

In example of the tabs, only "syn-tabs" are used and there is no "syn-tab-group" around in the story.

The problem here is, our own styling for the active state, is the blue border under the tab.

The implementation for the blue border is not in the "tab" element itself, but in the "tab-group" of shoelace (the reason for this is e.g. to make the animation possible of the border, when another tab is selected).

Shoelace has additionally another color for the text of the activated tab. We don`t have this.

I already discussed it with Lea and we decided to do nothing here, because no one should use the "syn-tab" on its own, but they will always have a "syn-tab-group" around it and then the "activated" blue border will be there.

kirchsuSICKAG:
> the documentation of syn-tab-show and syn-tab-hide doesn't say anything about what the payload contains. only what type it is (string)

@klaarseSICKAG 
This is the information, that shoelace gives. I could vendor it to something like:
"Emitted when a tab is shown. The payload of the event returns the "panel" attribute of the shown tab"

Would that be more helpful?

kirchsuSICKAG:
> I want to add my coment to the open question: It is not that easy to change the border-width in Figma and also for consistency and from a design perspective the track width should be removed from the css properties.

@mariohamann What do you think about it?
I could remove the documentation about the css variable, so it is not obvious, that there is one. 
But what would happen with the other three? We have `--track-color`, `--indicator-color` and `--track-width`, that come from shoelace and `--indicator-width`, which I added. Should for all four of them the documentation be removed?

I'm not sure if we really want to remove them, as we take the flexibility (otherwise, they could still achieve it, if they have a look at the code, because I would only remove the documentation)

klaarseSICKAG:
> > the documentation of syn-tab-show and syn-tab-hide doesn't say anything about what the payload contains. only what type it is (string)
> 
> @klaarseSICKAG This is the information, that shoelace gives. I could vendor it to something like: "Emitted when a tab is shown. The payload of the event returns the "panel" attribute of the shown tab"
> 
> Would that be more helpful?

Either that or a PR for shoelace. I wouldn't block this PR for that though! This text would be helpful ...

schilchSICKAG (approved):

PR #498, opened 2024-05-29
feat: ✨ icon-only button

### LINKED ISSUE #309: feat: ✨ icon-only button

# Component

I suggest to add the feature to the following component: syn-button

# Story

As a user of the Synergy Design System, I would like to use a button without text and just a single icon. The button should then be squared.

# Acceptance criteria

- [ ] Button is a square (width same as height) after an icon is placed into default slot as only child
- [ ] unittests
- [ ] story

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds the possibility to use the button as icon-only button which is squared, if only an icon is used in the default slot.

### 🎫 Issues

Closes #309 

## 👩‍💻 Reviewer Notes

When I vendored the button, I noticed, that until know we also always vendored the radio-button and the icon-button with the button.vendorism.js. Because we only checked for:
```js
if (path.endsWith('button.component.ts')) {
```
I fixed this, so they are no longer getting vendored with the button.vendorism. But because of that we got a change in the radio-button.components.ts.
I decided not to create a radio-button.vendorism.js because until know we do NOT officially ship the radio-button (it just came with the radio-group) and therefore other stuff is still not right in the radio-button. I would postpone this, until we officially add the radio-button as component.

## 📑 Test Plan

- have a look at the new icon-only story: http://localhost:6006/?path=/story/components-syn-button--icon-only
- check that nothing else was broken with this changes

## ✅ DoD

### DIFF
--- a/packages/docs/stories/components/button.stories.ts
+++ b/packages/docs/stories/components/button.stories.ts
@@ -142,6 +142,56 @@ export const SettingACustomWidth: Story = {
   <syn-button size="large" style="width: 100%;">Large</syn-button>`,
 };
 
+export const IconOnly: Story = {
+  parameters: {
+    docs: {
+      description: {
+        story: generateStoryDescription('button', 'icon-only'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-button size="small" variant="filled">
+      <syn-icon name="settings"></syn-icon>
+    </syn-button>
+    <syn-button size="small">
+      <syn-icon name="settings"></syn-icon>
+    </syn-button>
+    <syn-button size="small" variant="text">
+      <syn-icon name="settings"></syn-icon>
+    </syn-button>
+
+    <br> 
+
+    <syn-button size="medium" variant="filled">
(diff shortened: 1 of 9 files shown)


packages/docs/stories/components/button.stories.ts
```diff
@@ -142,6 +142,56 @@ export const SettingACustomWidth: Story = {
...
+  parameters: {
+    docs: {
+      description: {
+        story: generateStoryDescription('button', 'icon-only'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-button size="small" variant="filled">
+      <syn-icon name="settings"></syn-icon>
```
mariohamann, line 155: Please add a label on every `syn-icon` to make this accessible.

kirchsuSICKAG: Thanks! Added the labels

reichst1SICKAG (approved):
Design Review done

schilchSICKAG (approved):

PR #503, opened 2024-06-04
fix: 🤔  syn-nav-item getting too wide when in syn-side-nav with keyboard navigation

### LINKED ISSUE #477: fix: 🤔 syn-nav-item getting too wide when placed inside a syn-side-nav in keyboard navigation

# Version

The bug occurred with the following version of the component library: 1.23.1

# Component

The issue arises when using the following component(s): syn-nav-item

# Environment

The issue occurred in the following browser(s): All evergreen browsers

# Description

When visiting the [default side-nav story](https://synergy-design-system.github.io/?path=/story/components-syn-side-nav--default) and trying to navigate via keyboard, a vertical scrollbar is shown. This is caused by the outline of `<syn-nav-item>`, which draws a container around the original `<button>` that has a width that is larger than the element itself. This was initially done to prevent the left and right outlines to show, however causes overflowing.

This can easily be adjusted by setting `overflow: hidden;` on the surrounding `<button>` element of the `<syn-nav-item>`.

# Acceptance criteria

- [ ] There should be no scrollbar when navigating with the keyboard when a `<syn-nav-item>` is navigated on via keyboard interaction in `<syn-side-bar>`.

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR fixes the issue that a scrollbar appears in the syn-side-nav when using keyboard navigation.
Additionally it fixes a wrong focus state of the vertical syn-nav-item, which was not aligned with the design.

### 🎫 Issues

Closes #477 

## 👩‍💻 Reviewer Notes

- Got to a story of the syn-side-nav and check that no longer a scrollbar appears when using keyboard navigation
- Check out the focus states of the vertical and horizontal nav-items, that they are aligned to the design

## WIP
- currently I am waiting for design feedback, if the focus-border should be before or behind the current indicator
image

## ✅ DoD


kirchsuSICKAG:
We decided to move the current indicator to the foreground so the focus border is behind, to be aligned with the appearance of the sy-tab

schilchSICKAG (approved):

Issue #490, opened 2024-05-15
chore: 🔧 Make angular demo run with standalone dependencies to prevent errors when required version of wrappers differ


schilchSICKAG:
# Description

After multiple problems with e2e tests not working as expected, we traced the problem to an error while building our angular demo project. This tries to share the version of `@angular/core` and `@angular/forms` with the one provided in `@synergy-design-system/angular`. This leads to problems with the angular compiler, as the versions __may__ differ. In this case (e.g. when `@synergy-design-system/angular` has a different version of one of this packages, the local module resolution of angulars build system does not work anymore. However, the server itself **is still starting up**. This leads to the problem that selectors created via Playwright will never run in a timeout, making the tests stay in sort of a limbo between death and live.

> This is NOT a problem that users may likely face in the wild, as it is caused by the dedupe mechanism used in `pnpm` and the usage of a monorepo that locally links the packages together.

# Rationale

We have to make sure updates of `peerDependencies` do not cause problems with the build of the demo to make tests less flaky.
Only happens when updating Angular

schilchSICKAG:
This was tried in https://github.com/synergy-design-system/synergy-design-system/pull/502. However, it does not work correctly, because there are [multiple](https://github.com/pnpm/pnpm/issues/7811) [issues](https://github.com/pnpm/pnpm/issues/6088) with this.

schilchSICKAG:
During development of https://github.com/synergy-design-system/synergy-design-system/pull/508, we noticed that this is also happening in our react vite builds. However, in this case, only production builds seem to have the problem. This may be some internal problem when using esbuild.

schilchSICKAG:
Currently not possible in an easy way. Think we will have to live with this for the time being

PR #508, opened 2024-06-14
chore: 🔧 e2e testing on production builds

### LINKED ISSUE #505: chore: 🔧 e2e testing on production builds
# Description

Run e2e tests on production builds instead of framework specific serve-scripts.
Also enable Angular tests again.

# Rationale

Currently tests run on framework specific serve scripts (e.g. ng serve). The problem of this approach is that there is not a productive build created (which might deviate) and the framework specific webserver is used which currently results in a failing connection (probably due to OS issues).

# Acceptance criteria

- [x] the same webserver is used for all builds (vanilla, vue, ng, react)
- [x] enable ng test
- [x] webserver points to productive builds
- [ ] only one port is used (and not 4 like now)
- [x] tests are running
- [x] tests can be executed in parallel

schilchSICKAG:
# Pull Request

## 📖 Description

This PR refactors our e2e test pipeline for our demo projects to make them run smoother. For this to work, the new pipeline configuration is:

- using the production builds of the demo projects (except for React, see below)
- Not using `npm-run-all`, but a static webserver
- Making sure to just run one instance of playwright for more stable tests

### 🎫 Issues

Closes #505

## 👩‍💻 Reviewer Notes

Initially, I was hoping that I would be able to include a react production build, too. However, it turned out that the production build of our react demo runs fine __outside__ of the mono-repo, but not inside of it. There seem to be issues like #490 which effect the production build for vite on the react target only. Unfortunately, I was not able to fix those issues, so React runs on vite dev in the meantime. It may have to be something to do with how rollup bundles the files, but no configuration override or package update helped to ease this problem :(.

## 📑 Test Plan

- Check out the branch and run a clean install `pnpm clean` in the root of the repo
- Move to `packages/_private/e2e-demo-test`
- Run `pnpm start` to spin up the system
- Notice how the tests now run in sequence with only one instance of Playwright
- Also notice that the tests have valid names now, which should make debugging easier.

## ✅ DoD


klaarseSICKAG (approved):
nicely done

PR #522, opened 2024-06-26
feat: ✨add table stylings

### LINKED ISSUE #441: feat: ✨ Create Table Cell Component
### Userstory

As a user of the Synergy Design System, I would like to use the table cell component that can be used as a building block for a table

### Acceptance criteria
- [ ] Provide CSS only
- [ ] Provide stories
- [ ] Provide template
- [ ] Provide architecture

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds table-cell styles to the styles package and provides table templates

### 🎫 Issues

Closes #441 

## 👩‍💻 Reviewer Notes

- Solid resets the styling on the table class (for us on the `syn-table` class) via "all: unset". Do we want to do this also?

## 📑 Test Plan

have a look at:
- the new style stories in the docs package:  http://localhost:6006/?path=/docs/styles-syn-table-cell--docs, http://localhost:6006/?path=/docs/styles-syn-table--docs
- the new templates at the docs package: http://localhost:6006/?path=/docs/templates-table--docs
- the changes in the styles package

## ✅ DoD


packages/docs/src/helpers/styles.ts
```diff
@@ -35,7 +35,7 @@ export const renderStyles = (
 
   return html`
     <${usedTag} class=${classes}>
-      ${defaultSlot}
+      ${unsafeStatic(defaultSlot)}
```
schilchSICKAG, line 38: I guess this is needed for the table to work as a content element?

packages/docs/src/translations.ts
schilchSICKAG, line 1: Nicely done :)

packages/docs/src/vite-plugin-synergy-styles/index.ts
```diff
@@ -1,10 +1,15 @@
 import { mkdir, writeFile } from 'fs/promises';
 import { join } from 'path';
+import { execSync } from 'child_process';
 import type { Plugin } from 'vite';
 import type { Config } from './types.js';
 import { getStructure } from './structure.js';
 import { toCem } from './toCem.js';
 
+const runStylesBuild = () => execSync(
```
schilchSICKAG, line 9: As this is called in async context only, you could also go and use something like

```javascript
const runStylesBuild = async () => exec('...');
```

instead and await its result. Does not come to a different outcome, though :)

packages/docs/src/vite-plugin-synergy-styles/index.ts
```diff
@@ -66,6 +71,7 @@ export default function vitePluginSynergyStyles(
         return undefined;
       }
 
+      runStylesBuild();
```
schilchSICKAG, line 74: I am missing this on the other outputs where we are recreating the manifest, e.g. `configureServer`. I get that on `generateBundle` this should be already done, but I am not sure about configureServer.

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
```diff
@@ -14,7 +16,13 @@ import type {
 const getTypesAsArray = (tag: Tag): string[] => tag.type
   .split('|')
   .map(t => t.trim())
-  .map(t => `'${tag.name}--${t}'`);
+  .map(t => {
+    // If we don`t want a default value in a drop down, return an empty string
+    if (t === NO_DEFAULT) return '';
+
+    const separator = tag.name.includes('--') ? '-' : '--';
```
schilchSICKAG, line 23: Maybe we should add a comment here why it is needed. Otherwise we will read this line in 2 years (hopefully ;)) and don´t have a clue anymore. Mind adding some docs on this?

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
```diff
@@ -33,7 +41,14 @@ const getTypeForTag = (tag: Tag) => {
...
  */
-const tagIsAllowedToBeIncluded = (tag: Tag): boolean => tag.tag === 'variant';
+const tagIsAllowedToBeIncluded = (tag: Tag): boolean => ['variant', 'boolean'].includes(tag.tag);
+
+/**
+ * Check if a tag is a boolean type
+ * @param tag The tag to check
+ * @returns True if the tag is a boolean type
+ */
+const tagIsBooleanType = (tag: Tag): boolean => tag.tag === 'boolean' || !tag.type || tag.type.trim().length === 0 || tag.type === 'true';
```
schilchSICKAG, line 51: Why is a documentation value of `false` not also a boolean?
Just from the docs, I think both of these lines should be a boolean:

```css
@boolean { true } syn-table-cell Applies the table cell styling
@boolean { false } syn-table-cell Applies the table cell styling
```

packages/docs/stories/styles/table-cell.stories.ts
```diff
@@ -0,0 +1,186 @@
...
+        </tr>
+        <tr>
+          <td class="syn-table-cell syn-table-cell--border-bottom">Border Left</td>
+        </tr>
+      </tbody>
+    </table>
+  `,
+};
+
+export const ScrollingBehavior: StoryObj = {
```
schilchSICKAG, line 147: Does this include any logic? I saw that a `sticky` class is provided, however it is not used here.

packages/docs/stories/styles/table.stories.ts
schilchSICKAG, line 1: There is not much going on there. Do we need this story if all it does is adding a tiny amount of white-space when toggled.

On a side-note, I would have liked some automatic formatting after applying the class, something like:

```html
<table class="syn-table">
  <thead>
    <th>Headline</th>
  </thead>
  <tbody>
    <tr>
      <td>Line</td>
    </tr>
  </tbody>
</table>

.syn-table {
  border: var(--syn-table-width-default) solid var(--syn-color-of-table-border);
}

.syn-table th,
.syn-table-header {
  /* Proceed with styling as needed */
}

.syn-table tr:nth-child(even),
.syn-table-cell-odd {
  /* Insert zebra striping here */
}
```

This would be easier for users to adopt, because they just would have to add the `syn-table` class on top of their table and would not have to rely on dozens of separate classes.

You could also keep it clean for both versions, by adding the styles above to another file (`styles/src/table/auto-style.css`):

```css
/* Something like that */
.table-auto-style th {}
.table-auto-style tbody th:nth-child(even) {}
.table-auto-style td {}
```

packages/docs/stories/styles/table.stories.ts
```diff
@@ -0,0 +1,63 @@
...
+const { args: defaultArgs, argTypes } = storybookDefaults('syn-table');
+const { overrideArgs } = storybookHelpers('syn-table');
+
+const meta: Meta = {
+  args: overrideArgs([
+    {
+      name: 'default',
+      type: 'slot',
+      value: `<tbody>
+    <tr>
```
schilchSICKAG, line 24: Formatting is a bit messed up here

packages/docs/stories/styles/table.stories.ts
```diff
@@ -0,0 +1,63 @@
...
+    controls: {
+      disable: false,
+    },
+    docs: {
+      description: {
+        story: generateStoryDescription('table', 'default'),
+      },
+    },
+  },
+  render: (args: unknown) => html`
```
schilchSICKAG, line 56: I think you could just return renderStyles here, leading to a bit cleaner code. See

```typescript
export const Default: StoryObj = {
  parameters: {
    controls: {
      disable: false,
    },
  },
  render: (args: unknown) => renderStyles(args as RenderArgs),
};
```

vs

```typescript
export const Default: StoryObj = {
  parameters: {
    controls: {
      disable: false,
    },
  },
  render: (args: unknown) => html`
    ${renderStyles(args as RenderArgs)}
  `,
};
```

packages/styles/README.md
```diff
@@ -73,6 +73,9 @@ import "@synergy-design-system/styles/typography.css";
 
 <!-- BEGIN INLINE COMMENT -->
 
+- table.css
+  - table-cell.css
+  - table.css
```
schilchSICKAG, line 78: I understand the problem here, however, it feels a bit redundant to have a filesystem layout like:

```bash
./dist
  ./table.css # Where the whole bundle of the module is
  ./table # Where the subs reside
    ./table-cell.css
    ./table.css # We may run into troubles when someone loads the wrong "table.css"
```

I do not have an idea currently how we can tackle this, but we should at least think about make this better in the future.

packages/styles/src/table/table-cell.css
```diff
@@ -0,0 +1,128 @@
+/**
+ * Table cell classes
+ *
+ * @boolean { true } syn-table-cell Applies the table cell styling
+ * @boolean syn-table-cell--bg-neutral-50 Applies the table cell styling with a neutral background color
```
schilchSICKAG, line 5: I think `syn-table-cell--bg-neutral-50` and `syn-table-cell--bg-neutral-0` are quite long. Do we need names that are longer to type than to apply with our own css?

```css
.syn-table-cell--bg-neutral-50 {
    background-color: var(--syn-color-neutral-50);
}
```

vs.

```css
/* TODO: Find naming */
.syn-table-cell--light-grey {
    background-color: var(--syn-color-neutral-50);
}
```

packages/styles/src/table/table-cell.css
schilchSICKAG, line 1: Why do we have an indentation of 4 instead of 2?

packages/styles/src/table/table-cell.css
```diff
@@ -0,0 +1,128 @@
...
+    transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
+}
+
+/* I can not use var(--syn-color-neutral-950) here for some reason and therefore `grey`is used. If doing this, the fallback will not be used */
+@supports (background: color-mix(in srgb, grey 0%, transparent)) {
+    .syn-table-cell--shadow-bottom::after,
+    .syn-table-cell--shadow-top::after,
+    .syn-table-cell--shadow-start::after,
+    .syn-table-cell--shadow-end::after {
+        --shadow-start: color-mix(in srgb, var(--syn-color-neutral-950) 0%, transparent);
```
schilchSICKAG, line 82: Neat trick :)

packages/styles/src/table/table.css
```diff
@@ -0,0 +1,13 @@
+/**
+ * Table classes
+ * @variant syn-table Applies the table styling
+ */
+.syn-table {
+    /* TODO: do we want to reset all table styles via: all:unset? */
```
schilchSICKAG, line 6: I think this would be a good default, yes +1

packages/styles/src/table/table.css
```diff
@@ -0,0 +1,13 @@
...
+
+    /* TODO: do we want to set the border-spacing via css variable like solid? */
```
schilchSICKAG, line 8: I think that would make sense, too.

packages/docs/stories/templates/table.stories.tsx
```diff
@@ -0,0 +1,500 @@
...
+ */
+const createHeader = (shadow: 'top' | 'bottom' | 'start' | 'end' | undefined = undefined) => {
+  const shadowClass = shadow ? `syn-table-cell--shadow-${shadow} shadow-cell` : '';
+  // eslint-disable-next-line complexity
+  const getClassesForColumn = (column: number) => `syn-table-cell--header ${
+    ((shadow === 'end' && column === 0) || (shadow === 'start' && column === 4) || shadow === 'bottom') ? shadowClass : ''
+  }`;
+
+  return html`
+  <thead>
```
schilchSICKAG, line 41: Indentation is wrong here

packages/docs/stories/templates/table.stories.tsx
schilchSICKAG, line 1: Pretty nice I think +1. Mind adjusting the formatting when using the `html` helper? I find it a bit harder to read than needed. Example

```typescript
const createHeader = (shadow: 'top' | 'bottom' | 'start' | 'end' | undefined = undefined) => {
  const shadowClass = shadow ? `syn-table-cell--shadow-${shadow} shadow-cell` : '';
  // eslint-disable-next-line complexity
  const getClassesForColumn = (column: number) => `syn-table-cell--header ${
    ((shadow === 'end' && column === 0) || (shadow === 'start' && column === 4) || shadow === 'bottom') ? shadowClass : ''
  }`;

  return html`
  <thead>
    <tr>
      <th class="${getClassesForColumn(0)}">${getTranslation('table.header.name')}</th>
      <th class="${getClassesForColumn(1)}">${getTranslation('table.header.customer')}</th>
      <th class="${getClassesForColumn(2)}">${getTranslation('table.header.location')}</th>
      <th class="${getClassesForColumn(3)}">${getTranslation('table.header.contractStart')}</th>
      <th class="${getClassesForColumn(4)}"></th>
    </tr>
  </thead>
`;
};

should be

```typescript
const createHeader = (shadow: 'top' | 'bottom' | 'start' | 'end' | undefined = undefined) => {
  const shadowClass = shadow ? `syn-table-cell--shadow-${shadow} shadow-cell` : '';
  // eslint-disable-next-line complexity
  const getClassesForColumn = (column: number) => `syn-table-cell--header ${
    ((shadow === 'end' && column === 0) || (shadow === 'start' && column === 4) || shadow === 'bottom') ? shadowClass : ''
  }`;

  return html`
    <thead>
      <tr>
        <th class="${getClassesForColumn(0)}">${getTranslation('table.header.name')}</th>
        <th class="${getClassesForColumn(1)}">${getTranslation('table.header.customer')}</th>
        <th class="${getClassesForColumn(2)}">${getTranslation('table.header.location')}</th>
        <th class="${getClassesForColumn(3)}">${getTranslation('table.header.contractStart')}</th>
        <th class="${getClassesForColumn(4)}"></th>
      </tr>
    </thead>
  `;
};
```

I know that seems like nitpicking, but when you have multiple keys in Objects it is easier on the eye because all is in the same vertical alignment.

schilchSICKAG (changes requested):
Nice work overall 👍. I have just some smaller notes to make, nothing of this is breaking. Maybe we should discuss tomorrow.

packages/docs/src/helpers/styles.ts
kirchsuSICKAG: I needed it for the table story in `docs/stories/styles/table.stories.ts`
to do this:

```js
args: overrideArgs([
    {
      name: 'default',
      type: 'slot',
      value: `<tbody>
    <tr>
      <td>
        Cell content
      </td>
    </tr>
  </tbody>`,
    },
  ], defaultArgs),
  ```

So if someone looks at the code of the story he does not get this:

```html
<table>Cell content</table>
```

but this: 
```html
<table>
  <tbody>
    <tr>
      <td>Cell content</td>
    </tr>
  </tbody>
</table>
```

packages/docs/src/vite-plugin-synergy-styles/index.ts
kirchsuSICKAG: I just tested it with the `exec` method and for me it endlessly retriggers the `load` function when running it with `pnpm start`. But maybe I have also done something wrong  😅

Bildschirmfoto 2024-07-01 um 17 15 35

packages/docs/src/vite-plugin-synergy-styles/index.ts
kirchsuSICKAG: I just forgot the others, because for the `pnpm start` I only needed it in the `load` function 😅

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
kirchsuSICKAG: I tried my best to explain it^^

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
kirchsuSICKAG: if you would write the documentation like this, it would be handled correctly. Because then the `tag.tag === 'boolean'` would come to action. But if someone would for some reason has the idea to do this (use `variant`instead of `boolean`):

```js
@variant { true } syn-table-cell Applies the table cell styling
```

then it would also use the boolean type. But I'm not sure if we need this at all..
Maybe we should just remove this, because if someone wants to do a boolean class he should just use `@boolean` and not `@variant`. 
Or what do you think? 
If we let it like this, maybe I should just add `|| tag.type === 'false` so this is also possible: 
```js
@variant { false } syn-table-cell Applies the table cell styling
```

packages/docs/stories/styles/table-cell.stories.ts
kirchsuSICKAG: The sticky class is used in line 159 and 160.
If the shadow should be visible, we need either a `position: relative` or a `position: sticky` on the table-cell with the shadow, otherwise it is not visible / on the correct place

packages/docs/src/vite-plugin-synergy-styles/index.ts
schilchSICKAG: Uff, then lets just stay with the sync version, its pure semantics overall, so that does not matter much :)

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
schilchSICKAG: I ment the reason for `--` to `-` thingy. I think the NO_DEFAULT is fine as it is. However I don´t think we will be guessing why `tag.name.includes('--') ? '-' : '--';` is needed in a year or two from now.

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
schilchSICKAG: Ah, I get it. So currently, you could use `@variant` as a boolean trigger, too? Not sure where we want to go with this. On the one hand, it would distinguish booleans and variants when using the first approach. On the other hand, when allowing the second one, we could skip `@boolean` completely. However in this case, you would have problems with comments like this:

```css
@variant { false } syn-table-cell Clearly a boolean
@variant { true } syn-table-cell Clearly a boolean
@variant { false | true } syn-table-cell Clearly a boolean
@variant { false | true | blue } syn-table-cell This is not a boolean anymore?
```

packages/docs/stories/styles/table-cell.stories.ts
schilchSICKAG: As someone that has a lot of bad experience with tables and sticky columns: Is `position: sticky` working on table cells with all browsers now? Did not know that.

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
kirchsuSICKAG: Yes. I added a comment with explanation on this line with the separator . My comment to your thread was meant for my commit with the new explanation  ;)

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
kirchsuSICKAG: In the case of the last two examples it would not get triggered. Because the type would be either `type = 'false | true'` or for the other case `type = 'false | true | blue`

packages/docs/stories/styles/table-cell.stories.ts
kirchsuSICKAG: I tested it with all current browsers and on can I use I found this: 
Bildschirmfoto 2024-07-02 um 11 21 39

image


It seems to work for all, for a while now

schilchSICKAG: Thats great news (that are not so new it seems... I better polish up my table css skills ;))

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
schilchSICKAG: Fine for me, lets stick with that.

packages/docs/src/vite-plugin-synergy-styles/toCem.ts
schilchSICKAG: :)

packages/styles/src/table/table-cell.css
kirchsuSICKAG: My editor was misconfigured with a indentation size of 4. Fixed this

kirchsuSICKAG: Or it is not fixed.. It does now look for even more indentation.. Is it correct for you?

schilchSICKAG: It seems to be fixed. I forgot to apply the styling for css files via editorconfig. Just pushed a change with this included.

schilchSICKAG (changes requested):
I am pretty happy with the result 👍. We still have to think about how to document the stories. I saw that you are now using inline comments. Can we maybe just concat the shared doc dokens and the inplace dev docu to at least share a bit of it?

PR #542, opened 2024-07-09
feat: ✨add syn-combobox

### LINKED ISSUE #339: feat: ✨ auto-complete / combo-box

# Story

As a user of the Synergy Design System, I would like to have an input field which makes suggestion based on the already typed letters.

# Acceptance criteria

- [ ] The input field provides an API to input suggestions
- [ ] The input field provides suggestions based on the types letters
- [ ] The filter for the suggestions can be modified by the user
- [ ] A meaningful default filter is provided
- [ ]

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-combobox>` component that may act as a replacement for `<input type="text">` with a `<datalist>`.

### 🎫 Issues
Closes #339 

## 👩‍💻 Reviewer Notes
- The select was used as base for the combobox and it also looks exactly the same. Because of that the whole styling of the  select component (`select.styles.ts` and `select.custom.styles.ts`) is used for the combobox. Therefore a new step in the `pnpm vendor.set` was added, to copy these styles and adapt them for the combobox. Because of that we are able to get all styling changes from the select component also for the combobox.

- Currently slotted `<syn-icon>`'s do not have the correct sizes. this is because this bug occurs in the select component (see https://github.com/synergy-design-system/synergy-design-system/issues/573). As soon as this is fixed in the select, it will be fixed in the combobox too.

## 📑 Test Plan

have a look at:
- the component in src/components/combobox
- the at the stories in docs package: http://localhost:6006/?path=/docs/components-syn-combobox--docs
- have a look at the keyboard navigation of the combobox
- have a look at the api (especially `filter`, `getOption` properties)

## ✅ DoD


packages/components/src/components/combobox/combobox.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    min-width: 0;
+    position: relative;
+    align-items: center;
+    justify-content: start;
+    font-family: var(--syn-input-font-family);
+    font-weight: var(--syn-input-font-weight);
+    letter-spacing: var(--syn-input-letter-spacing);
+    vertical-align: middle;
+    overflow: hidden;
+    cursor: pointer;
```
schilchSICKAG, line 52: This should not be pointer when hovering over the text of the input field. When hovering on the input field, it should be `text` to indicate that we are using a regular input.

packages/components/src/components/combobox/combobox.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    left: 0;
+    width: 100%;
+    height: 100%;
+    padding: 0;
+    margin: 0;
+    opacity: 0;
+    z-index: -1;
+  }
+
+  /* Standard comboboxs */
```
schilchSICKAG, line 98: typo, should be combobox instead

packages/components/src/components/combobox/combobox.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    padding-inline: var(--syn-input-spacing-large);
+  }
+
+  .combobox--large .combobox__clear {
+    margin-inline-start: var(--syn-input-spacing-large);
+  }
+
+  .combobox--large .combobox__prefix::slotted(*) {
+    margin-inline-end: var(--syn-input-spacing-large);
+  }/* Prefix */
```
schilchSICKAG, line 167: This is missing one (better two) newlines

packages/components/src/components/combobox/combobox.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    /* Make sure it adheres to the popup's auto size */
+    max-width: var(--auto-size-available-width);
+    max-height: var(--auto-size-available-height);
+  }
+
+  .combobox__listbox ::slotted(syn-divider) {
+    --spacing: var(--syn-spacing-x-small);
+  }
+
+  .combobox__listbox ::slotted(small) {
```
schilchSICKAG, line 244: Maybe we should add "listbox" to the cspell.json?

packages/components/src/components/combobox/combobox.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  .combobox__prefix,
+  .combobox__suffix {
+    color: var(--syn-input-help-text-color);
+  }
+
+  /* Listbox */
+  .combobox__listbox {
+    /* @todo: Should be --syn-border-radius-medium, which should be set to 0 */
```
schilchSICKAG, line 331: Just had a look, it seems it is set to 0 (`--syn-input-border-radius-medium: var(--syn-border-radius-none);`). Maybe you could use the real token here now?

packages/components/src/components/combobox/combobox.styles.ts
```diff
@@ -0,0 +1,344 @@
...
+    border-radius: 0;
+    box-shadow: var(--syn-shadow-medium);
+  }
+
+  /**
+   * Make sure to hide the syn-divider for the first syn-optgroup
+   * Note! ::slotted does currently not work with ::part, so we
+   * opted for using a css variable here.
+   */
+  .combobox__listbox ::slotted(syn-optgroup:first-of-type) {
```
schilchSICKAG, line 341: I think this can be removed as it is specific to `syn-select[multiple]`.

packages/components/src/components/combobox/combobox.custom.styles.ts
```diff
@@ -0,0 +1,23 @@
+import { css } from 'lit';
```
schilchSICKAG, line 1: If we are overriding the combobox.styles, why don´t we add the code here to it?

packages/components/src/components/combobox/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @returns New array of all found syn-optgroup's
+ */
+export const filterOnlyOptgroups = (items: HTMLElement[]) => items.filter(isOptgroup);
+
+/**
+ * Normalize a string by removing accents and converting to lowercase
+ * @param str The string to normalize
+ * @returns The normalized string
+ */
+export const normalizeString = (str: string) => str.normalize('NFD').replace(/[\u0300-\u036f]/g, '').toLowerCase();
```
schilchSICKAG, line 51: Thats a nice one :)

packages/components/src/components/combobox/combobox.styles.ts
schilchSICKAG: Just seen there is a demo for this indeed. Thats quite nice I think, ignore this comment :)

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,963 @@
...
+      if (newIndex > filteredOptions.length - 1) newIndex = 0;
+    } else {
+      newIndex = currentIndex - 1;
+      if (newIndex < 0) newIndex = filteredOptions.length - 1;
+    }
+    this.setCurrentOption(filteredOptions[newIndex]);
+    scrollIntoView(this.currentOption!, this.listbox, 'vertical', 'auto');
+  }
+
+  private getAllFilteredOptions() {
```
schilchSICKAG, line 510: I am not sure where this comes from, but when using the demos, you will get duplicates for each item. In the demo applications, you have entries for `American` and `american`, but only `American` is set as an option initially. This happens in all framework demo projects. Mind to have a look into this?

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,963 @@
...
+
+      await stopAnimations(this);
+      this.listbox.hidden = false;
+      this.popup.active = true;
+
+      const { keyframes, options } = getAnimation(this, 'combobox.show', { dir: this.localize.dir() });
+      await animateTo(this.popup.popup, keyframes, options);
+
+      this.emit('syn-after-show');
+    } else {
```
schilchSICKAG, line 631: You could omit the `else` here by returning in the `if` statement, which could make the code easier to read.

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,963 @@
...
+  focus(options?: FocusOptions) {
+    this.displayInput.focus(options);
+  }
+
+  /** Removes focus from the control. */
+  blur() {
+    this.displayInput.blur();
+  }
+
+  private createComboboxOptionsFromQuery(queryString: string) {
```
schilchSICKAG, line 707: The code is pretty hard to read in my opinion, maybe we can come up with an easier version, something like this (not tested):

```typescript
private createComboboxOptionsFromQuery(queryString: string) {
    this.filteredOptions = this.getSlottedOptions()
      .filter(o => this.filter(o, queryString) || queryString === '')
      .map(o => {
        const clonedOption = option.cloneNode(true) as SynOption;
        // Check if the option has a syn-optgroup as parent
        const hasOptgroup = option.parentElement?.tagName.toLowerCase() === 'syn-optgroup';

        if (!hasOptgroup) {
          return clonedOption;
        }

        const optgroup = option.parentElement as SynOptGroup;
        const filteredOptgroup = this.filteredOptions.find((el) => el.id === optgroup.id);
        // Check if the optgroup was already added to the filteredOptions
        if (!filteredOptgroup) {
          const clonedOptgroup = option.parentElement.cloneNode() as SynOptGroup;
          clonedOptgroup.appendChild(clonedOption);
          return clonedOptgroup;
        } else {
          filteredOptgroup?.appendChild(clonedOption);
        }
      });
  }
```

packages/components/src/components/combobox/combobox.component.ts
schilchSICKAG: This can also be seen in the "Highlight Query" demo

schilchSICKAG (changes requested):
There are a couple of comments that I have left on various files, and also some usability problems that I encountered which I will point out here:

Focus handling:

- [x] It should automatically focus the text field when you have a value already and you are clicking somewhere near the text. Currently, when you click for example the border (which indicates is clickable via `cursor: pointer`) and something is already entered, the input does not get focus.
- [ ] When clicking the down arrow, nothing happens if something is already typed.
- [ ] I would have suspected that the selection can not only be used as a input/datalist, but also as a replacement for select tags, e.g. only available options are valid. This is currently not possible. Maybe we should allow this as a feature (e.g. via a boolean prop?)
- [x] After a selection has happened for an option and the dropdown is closed and opened again, there is always the selected option in the dropdown as an only child. Does this make sense? Maybe we should not show the dropdown at all if it has only one child and this child matches 1:1? Same is true when using keyboard navigation.
- [x] Group Elements story: I think the first visible item should not have a border drawn above it.
- [x] Async Example: When you type something like `_8` and the item comes in after the defined timeout, the list is not updated. However, if you manually restart the search, the item is found. Suggestion: Maybe we should rerender this on `slotchange` if the list is open?

Despite the issues outlined above: Thats some solid work, I can see the effort you put into all of this :) 👍

reichst1SICKAG (approved):
All done from design prespective.

packages/components/src/components/combobox/combobox.custom.styles.ts
kirchsuSICKAG: @schilchSICKAG 
Mario mentioned he would find it better, that we vendor the select styles and don`t just copy it initially, because if something is changed in the select styling, we need to keep in mind to change it also in the combobox. 

(we use the exact same styling as for the select)

packages/components/src/components/combobox/combobox.styles.ts
kirchsuSICKAG: Not the `--syn-input-border-radius-medium` token but `--syn-border-radius-medium` is used here (it is not the input element itself, but the open list underneath the input).
And this token is still not set to 0: 
Bildschirmfoto 2024-09-09 um 13 39 10

schilchSICKAG: Ah, got it. Why don´t we use the input-border-radius here? Or just use `syn-border-radius-none` directly?

kirchsuSICKAG: good question. I can change it to use the syn-border-radius-none instead for the select, so we get it also in the combobox.
But I would let the comment stay there, because if at some point we use the correct token, this can be used from others to better adapt the theme (if they want to use rounded border)

packages/components/src/components/combobox/combobox.styles.ts
kirchsuSICKAG: the input field itself, now has cursor:text

packages/components/src/components/combobox/combobox.component.ts
kirchsuSICKAG: I needed to adapt some things, but I tried to make it more readable :)

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,967 @@
...
+
+              <input
+                class="combobox__value-input"
+                type="text"
+                ?disabled=${this.disabled}
+                ?required=${this.required}
+                .value=${this.value}
+                tabindex="-1"
+                aria-hidden="true"
+                @focus=${() => this.focus()}
```
schilchSICKAG, line 891: I think this could be written without an arrow function for better performance, e.g.
```suggestion
                @focus=${this.focus}
```

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,967 @@
...
+          return undefined;
+        }
+
+        const clonedOptgroup = optgroup.cloneNode() as SynOptGroup;
+        clonedOptgroup.appendChild(clonedOption);
+        optgroups.push(clonedOptgroup);
+        return clonedOptgroup;
+      })
+      // we need to remove the undefined values here
+      .filter(el => el) as Array<SynOption | SynOptGroup>;
```
schilchSICKAG, line 729: Theres a nice shortcut here that I often use and may be seen as more "elegant" by some:


```suggestion
      .filter(Boolean) as Array<SynOption | SynOptGroup>;
```

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,967 @@
...
+    return getAllOptions(getAssignedElementsForSlot(this.defaultSlot)).flat();
+  }
+
+  private getSlottedOptGroups(): SynOptGroup[] {
+    return filterOnlyOptgroups(getAssignedElementsForSlot(this.defaultSlot));
+  }
+
+  /* eslint-disable no-param-reassign */
+  private handleDefaultSlotChange() {
+    const slottedOptions = this.getSlottedOptions();
```
schilchSICKAG, line 767: You could get rid of some parts of the complexity be returning early. It would also make the code more performant
```suggestion
    // Rerun this handler when <syn-option> is registered
    if (!customElements.get('syn-option')) {
      // eslint-disable-next-line @typescript-eslint/no-floating-promises
      customElements.whenDefined('syn-option').then(() => this.handleDefaultSlotChange());
      return;
    }

    const slottedOptions = this.getSlottedOptions();
    const slottedOptgroups = this.getSlottedOptGroups();

    slottedOptions.forEach((option, index) => {
      option.id = option.id || `syn-combobox-option-${index}`;
    });

    slottedOptgroups.forEach((optgroup, index) => {
      optgroup.id = optgroup.id || `syn-combobox-optgroup-${index}`;
    });

    this.createComboboxOptionsFromQuery(this.value);

    if (this.hasFocus && this.value.length > 0 && !this.open) {
      // eslint-disable-next-line @typescript-eslint/no-floating-promises
      this.show();
    }
```

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,967 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      this.hide();
+    }
+  };
+
+  private handleLabelClick() {
+    this.displayInput.focus();
+  }
+
+  private handleComboboxMouseDown() {
```
schilchSICKAG, line 413: I don´t know if I am the only one that has a usability issue here (may be personal preference), but I would suggest to have a slight delay on this. Reason: Some people (including me ;)) use a double click mouse gesture to select the whole input text so they can remove it. When using such a gesture, it will open, close and finally reopen the dropdown.

packages/components/src/components/combobox/combobox.component.ts
schilchSICKAG: Ah I see, this is fine for me then :)

packages/components/src/components/combobox/combobox.component.ts
kirchsuSICKAG: nice catch! Never tried this until now

kirchsuSICKAG: after discussion we decided to not fix the open / close of the listbox on text selection.
This would lead to other implications like clicking on the combobox will open the listbox delayed, which feels very strange.
Also others also don`t do anything here like spectrum, whose component has the same problem

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,967 @@
...
+  }
+
+  /** Gets the associated form, if one exists. */
+  getForm(): HTMLFormElement | null {
+    return this.formControlController.getForm();
+  }
+
+  /** Checks for validity and shows the browser's validation message if the control is invalid. */
+  reportValidity() {
+    return this.valueInput.reportValidity();
```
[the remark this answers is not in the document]
kirchsuSICKAG: the combobox will now submit the form on enter, if the listbox is not open

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,974 @@
...
+
+  /** Removes focus from the control. */
+  blur() {
+    this.displayInput.blur();
+  }
+
+  private createComboboxOptionsFromQuery(queryString: string) {
+    const optgroups: SynOptGroup[] = [];
+
+    this.filteredOptions = this.getSlottedOptions()
```
schilchSICKAG, line 710: There is a smaller issue when using disabled options (which would be valid):

```html
<syn-combobox>
  <syn-option disabled>Dis</syn-option>
  <syn-option>hello</syn-option>
</syn-combobox>
```

When clicking the combobox, it will open the menu but if nothing is yet filled in, close it automatically and set the invalid state. This may happen because the first item will receive focus but can´t because its disabled. This only seems to happen when clicking **outside** of the start of the combobox (I assume outside of the input?)

schilchSICKAG: After further investigation, this seems to be a regular issue when the available size is too small: The widget will trigger the click on the option that is overlying the comboboxes input and select it. Just set your screen size really low in one of the framework demos to see this effect.

packages/components/src/components/combobox/combobox.custom.styles.ts
```diff
@@ -0,0 +1,34 @@
...
+   * Hide the default slot, because the filtered options are rendered in the listbox__options
+   */
+  .combobox__listbox slot:not([name]) {
+    display: none;
+  }
+
+  /**
+   * The filtered options
+   */
+  .listbox__options syn-option mark {
```
schilchSICKAG, line 14: There is no `mark` tag used in the component? Is this from the custom optionrenderer you are using? If so, we may have to defined this outside of the component.

packages/components/src/components/combobox/combobox.custom.styles.ts
```diff
@@ -0,0 +1,34 @@
...
+    color: var(--syn-color-neutral-950);
+    font: var(--syn-body-medium-bold);
+  }
+
+  .listbox__options syn-option[aria-selected='true'] mark {
+    color: var(--syn-color-neutral-0);
+  }
+
+  .combobox__display-input {
+    cursor: text;
```
schilchSICKAG, line 25: This should only happen if the combobox is not disabled.

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -0,0 +1,974 @@
...
+  /* eslint-enable @typescript-eslint/no-floating-promises */
+
+  /**
+   * Selects the following or previous option.
+   *
+   * @param isNext - A boolean indicating whether to select the following option (true)
+   *                 or the previous option (false).
+   */
+  private selectNextOption(isNext: boolean) {
+    const filteredOptions = this.getAllFilteredOptions();
```
schilchSICKAG, line 502: Maybe this should also take into account that the next wanted option can be disabled, e.g.:

```html
<syn-combobox>
  <syn-option>1</syn-option>
  <syn-option disabled>2</syn-option>
  <syn-option>3</syn-option>
</syn-combobox>
```

1. When using keyboard navigation, I have to press three times to move to "3". Maybe we should skip the disabled one?
2. When clicking a disabled option, the value of the field resets to `""`

packages/docs/stories/templates/form.stories.tsx
```diff
@@ -178,7 +178,12 @@ export const ContactForm = {
               `)}
             </syn-select>
 
-            <syn-input label="${getTranslation('contactForm.referenceContactLabel')}" name="salesPerson" required></syn-input>
+            <syn-combobox label="${getTranslation('contactForm.referenceContactLabel')}" name="salesPerson" required>
```
schilchSICKAG, line 181: The flyout has some weird graphic behaviour, propably hoisting?

packages/components/src/components/combobox/combobox.custom.styles.ts
kirchsuSICKAG: Yes, this comes if someone wants to highlight the characters with a `mark` element. 
The alternative would be, that everyone has to add this styling in their `getOption` render method, if they would like to have our recommended styling for highlighted elements. 

They would need to do this in the `getOption`:
image

And then the styling is still not exactly correct, because this styling would be missing: 
image

and the result would then look like this for keyboard highlighting of an option (the `g` should be white also):
image

For slotted elements we also sometimes do it for the people, to style the elements (e.g. Slotted syn-icon) so they don't have to do it.

Therefore I would let the styling stay there

packages/components/src/components/combobox/combobox.component.ts
kirchsuSICKAG: Our select also behaves like this, that you can navigate to an disabled option with the keys. But you are right, the disabled option should definitely not be selectable, but it should happen nothing when selecting a disabled option.

packages/components/src/components/combobox/combobox.custom.styles.ts
schilchSICKAG: I am fine with this, as long as we document it properly (e.g. something like: "when using `getOption`, make sure to use a `<mark>` element for brand compliance")

packages/docs/stories/templates/form.stories.tsx
kirchsuSICKAG: Seems like this is not only a problem of the combobox, but also with the select.
On the left is the combobox, on the right the select: 


image

kirchsuSICKAG: this can't be fixed via hoist :/

packages/components/src/components/combobox/combobox.custom.styles.ts
kirchsuSICKAG: Where would you add the documentation? To the property documentation ( in the table at the top) or at the highlight story? 
Both documentations are already relatively long😅

packages/docs/stories/templates/form.stories.tsx
kirchsuSICKAG: I created an issue in [our backlog](https://github.com/synergy-design-system/synergy-design-system/issues/612) and in [shoelace](https://github.com/shoelace-style/shoelace/issues/2171).

I added a temporary fix in the form template. As soon as the issue is fixed ( if even possible) this can be removed.

packages/components/src/components/combobox/combobox.custom.styles.ts
schilchSICKAG: I think I would add it to the documentation for `getOption`. Maybe we should also think about to alter `getOption`s signature, maybe like:

```typescript
// Current
getOption(option: SynOption, queryString: string)`

// New:
getOption(option: SynOption, queryString: string, autoMark?: boolean = true)`

// or even:
type GetOptionOptions = {
  autoMark?: boolean;
};

getOption(
  option: SynOption,
  queryString: string,
  opts?: GetOptionOptions = { autoMark: true, }
)`
```

We could then go the route of defining our default like so:

```typescript
@property() getOption: (option: SynOption, queryString: string, autoMark?: boolean = true) => TemplateResult | string | HTMLElement = (option, q, autoMark) => autoMark ? YOUR_CUSTOM_LOGIC_CURRENTLY_IN_TEMPLATE : option;
```

kirchsuSICKAG: If we will integrate the highlighting function directly in the combobox, maybe it would make sense to add a boolean property for it. 
So the people just need to do something like this: 

```html
<syn-combobox highlight ></syn-combobox>
```

if we do it as option in the getOption function, they would need to go into javascript

```js
const combobox = document.querySelector('syn-combobox');
const oldGetOption = combobox.getOption;
combobox.getOption = (option, query, opts)=> oldGetOption(option, query, { ... opts, autoMark: true } );
```

schilchSICKAG: Yeah, this was probably not well thought through. We could also go another route and export two getOption functions, one with and one without highlighting. However, this would also require setting it on every instance, which is unfortunate.

mariohamann: Just a thought from the outside (😂): I think it's totally fine to set specialiced options on every instance. As soon as you alter things, there will be many things involved anyways. You could export a "mark" method as well that could be used then on the content.

The most important thing might be a good default and then I believe it's totally fine to give some effort to the devs.

kirchsuSICKAG: Added a highlight option render utility function, which can be imported and used

packages/components/src/components/combobox/combobox.component.ts
kirchsuSICKAG: should now be fixed for the examples. This is a problem for the combobox as well as for the select, if giving the listbox of them a fixed `max-height` via `::part`.
If there is no longer enough space for the height, we enforce it with this fix height. The solution is doing following: 

```html
max-height: min(var(--auto-size-available-height), 112px);
```
The `--auto-size-available-height` is a css variable calculated from the syn-popup, how much space for the height is left

schilchSICKAG: Seems to work fine, thank you :)

packages/components/src/components/combobox/combobox.custom.styles.ts
kirchsuSICKAG: done 👍

schilchSICKAG (approved):

PR #551, opened 2024-07-12
feat: ✨ syn-range

### LINKED ISSUE #325: feat: ✨ syn-range

# Story

As a [Developer] using the Synergy Design System, I would like to give my users the ability to select a value of one to multiple ranges more easily. This may for example be used for things like percentual values (e.g. 0 - 100) or other such types of controls where exact precision is **not required**.

> Please have a look if https://github.com/shoelace-style/shoelace/pull/1912 can be used as a base.
> Maybe we can adjust this PR or use it as a base.

# Acceptance criteria

- [ ] The control must be able to act as a simple range slider, according to the current [specs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range).
- [ ] The control must be part of form submission
- [ ] The control should support all form attributes that a native `<input type="range" />` is able to (e.g. `required`, `min`, `max`).
- [ ] The control should support all form css selectors that a native `<input type="range" />` is able to (e.g. `syn-range::valid`, `syn-range::invalid`)
- [ ] All parts of the control (e.g. track and knob) must be available for external styling via css parts.
- [ ] The control supports showing multiple tick items.
- [ ] The control is able to show multiple `<syn-range-knob>` components.
- [ ] It should be able to be navigated via mouse/touch and keyboard interaction.
- [ ] It should by accessible via screenreaders and follow a11y guidelines.
- [ ] It should allow to place an arbitary number of knobs that can be used to "block" the next value with a defined `min` and `max` value.
- [ ] shoelace is informed about this component and offered for a merge

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-range>` form control that may act as a replacement for `<input type="range">` in `<form>`. It also includes the optional `<syn-range-tick>` element that may be used to create tickmarks for a given `<syn-range>` in a declarative way.

### 🎫 Issues

Closes #325

## 👩‍💻 Reviewer Notes

The following features are included:

- [x] Doc Tokens
- [x] May act as a single or multi range slider
- [x] Track is clickable. When using multiple knobs and clicking the track, the closest knob will move to the clicked position.
- [x] Aria handling is implemented for both versions (has to be checked by an expert, though @mariohamann)
- [x] Value Clamping according to min and max values)
- [x] Form-Submission and Validation (Validation has to be done via `setCustomValidity` as values are always clamped to min/max, see below)
- [x] Two way data binding enabled for Angular and Vue
- [x] Demos with single knob available in our test project (Happiness range)
- [x] Keyboard-Navigation for single and multi knobs
- [x] Ability to provide a custom tooltip formatter (Currently this sets the tooltips `content` prop, so no html is possible. Do we want to allow this?).
- [x] **To be discussed**: Bigger invisible touch target for knobs (have a look at the `.handle::after` definition for this. I like it as it makes hitting the knobs easier, but your milage may vary.
- [x] Define a bigger horizontal screen estate when clicking on the track
- [x] Tooltip-Disabled story broke after last refactoring.
- [x] Add tests for using multiple knobs (Currently all features are tested with one knob only).
- [x] Type cleanups have to be implemented to make the code feel more uniform with Shoelace conventions (e.g. `public getForm(): HTMLFormElement | null {` should be just `public getForm() {`
- [x] Add a multi range slider to our demos (a single range one is included in all of our test projects).
- [x] Adjust storybook to only include Stories from Figma (currently there are some more for testing)
- [x] Restrain knobs to overlay each other (currently you can move the knobs anywhere you like. Do we want this at all)
- [x] As the size for a knob is only adjusted while it is hovered and resets on press or mouse-out, the Tooltip tends to "jump" around.
- [x] Used `classNames` do not follow **BEM** currently. Will get updated soon.
- [x] Mobile Systems do not release the pointer, leaving the selected knob in its "big" state.
- [x] Browser validation popup is shown if setCustomValidity is set and in form or reportValidity used

### List of work that **still** has to be done:

- [ ] Final adjustments for layout additions.

## 📑 Test Plan

- Look at the stories in Storybook and play around with them. Make sure they still work when resizing the browser window, use keyboard navigation in single and multi sliders.
- Have a look at the demo projects at how the slider is used and submitted in forms.
- Also make sure to test with touch enabled browsers to test the touch behaviour.

## ✅ DoD


klaarseSICKAG:
- very nice. many things working already really smooth!
- the knob stays big if finger is lifted from it
- clicking on the track puts the knob there, great! but the track is quite hard to target since is so slim. could the touch area be bigger than the track itself? (maybe similar to --thumb-clickable-area)
- --thumb-clickable-area doesnt seem to work
- I wouldn't restrain knobs to overlay each other. this is application specific and should be possible via js !?
- couldn't get the tooltip formatter working. just tried ()=>"test"
- would be great to have some templates, e.g. for logarithms, custom legend labels, non crossing knobs, ...

schilchSICKAG:
Hey, thanks for the quick response :).

> * the knob stays big if finger is lifted from it
Yes, this is because it is defined like this in Figma (at least currently). Also note the "active" handle is having the same background as the inactive one, so there is technically a pressed state (currently visualized via `cursor`), but no other highlight. Will talk about this with @mariohamann next week. I also think we should leave it big or should not make it bigger at all?

> * clicking on the track puts the knob there, great! but the track is quite hard to target since is so slim. could the touch area be bigger than the track itself? (maybe similar to --thumb-clickable-area)

Great Idea, @kirchsuSICKAG mentioned it already, will at it to the list.

> * --thumb-clickable-area doesnt seem to work

Maybe this is poorly documented. I am using `scale` in here, so you will have to use values like `1.4` to make it scale by 40%. This is easier because in this case I am able to give the clickable area negative margins and the center stays the same (see https://github.com/synergy-design-system/synergy-design-system/blob/feat/325-syn-range/packages/components/src/components/range/range.styles.ts#L140 and https://github.com/synergy-design-system/synergy-design-system/blob/feat/325-syn-range/packages/components/src/components/range/range.styles.ts#L170).

> * I wouldn't restrain knobs to overlay each other. this is application specific and should be possible via js !?

Good catch, we might have to implement a custom event `syn-move` (or something like that) that is able to be monitored and canceled to make this work. With such an event, it will be possible to calculate the overlap.

> * couldn't get the tooltip formatter working. just tried ()=>"test"
This does not seem to work in Storybook as it basically just "knows" that there is a method, but is unable to call it :(. See the [Tooltip Formatter Story](https://github.com/synergy-design-system/synergy-design-system/blob/feat/325-syn-range/packages/docs/stories/components/range.stories.ts#L370) for how to use it. The current caveat is however you will need to return a string. If we would use the slot of `syn-tooltip` we could also allow arbitary html. Maybe we should opt for this (currently an open question in the PR description). 

> * would be great to have some templates, e.g. for logarithms, custom legend labels, non crossing knobs, ...

Yes, we will need to have a follow up ticket for this.

packages/docs/stories/components/range.stories.ts
```diff
@@ -0,0 +1,453 @@
...
+    <syn-range
+      tooltip="bottom"
+      max="100"
+      min="0"
+      value="50"
+    ></syn-range>
+  `,
+};
+
+export const TooltipDisabled: Story = {
```
mariohamann, line 392: image

Attribute differs from description.

packages/docs/stories/components/range.stories.ts
```diff
@@ -0,0 +1,453 @@
...
+      .syn-range-with-tick nav {
+        justify-content: space-between;
+        flex-direction: row;
+        display: flex;
+      }
+    </style>
+  `,
+};
+
+export const TooltipPlacement: Story = {
```
mariohamann, line 374: Doesn't show correct position in docs.

CleanShot 2024-07-16 at 10 32 20

packages/docs/stories/components/range.stories.ts
```diff
@@ -0,0 +1,453 @@
...
+            if (knob === target.knobs[1] && value < values[0]) {
+              e.preventDefault();
+            }
+          });
+        });
+    </script>
+  `,
+};
+
+export const Ticks: Story = {
```
mariohamann, line 342: I just realized: I think the middle of the knob should define point 0, not the left side. Same for the tick. (That's how we defined in Figma.) But maybe we have to recheck. I'm aware that Shoelace and HTML does it differently.

CleanShot 2024-07-16 at 10 32 07

packages/docs/stories/components/range.stories.ts
```diff
@@ -0,0 +1,453 @@
...
+        story: generateStoryDescription('range', 'multi-knob'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-range max="100" min="0" value="30 70"></syn-range>
+  `,
+};
+
+export const MultiKnobWithRestrictedMovement = {
```
mariohamann, line 302: <video src="https://github.com/user-attachments/assets/0b1bc4c6-cb1d-472d-a2ff-f5bed1dd7fd2"></video>

It behaves a bit strange when I try to move, depending on speed etc. it's not at the same position as the right knob. (Safari)

packages/docs/stories/components/range.stories.ts
```diff
@@ -0,0 +1,453 @@
...
+    ></syn-range>
+    <style>
+      .custom-track-color {
+        --track-color-active: var(--syn-color-success-600);
+      }
+    </style>
+  `,
+};
+
+export const CustomTrackOffset: Story = {
```
mariohamann, line 266: CleanShot 2024-07-16 at 10 30 10

Maybe not correct in design, but it should have `min="-50" max="50"` so that the middle is 0.

packages/docs/stories/components/range.stories.ts
schilchSICKAG: Leftover from original API, fixed with latest push.

packages/docs/stories/components/range.stories.ts
schilchSICKAG: I get it, seems I did not look good enough into the template :). Fixed with next version.

packages/docs/stories/components/range.stories.ts
schilchSICKAG: I see what you mean, have the same problem. I will try to have a look at what this is causing. Maybe it is because the hover of the other knob interfers and slallows the event.

packages/docs/stories/components/range.stories.ts
schilchSICKAG: Its quite tricky to make this effect work in the first place. In this cse, you are suggesting to start the track a bit further on the right (on the first tick?). If this is the case, the widget itself would get smaller and not aligned with the label anymore, am I correct?

packages/docs/stories/components/range.stories.ts
schilchSICKAG: Fixed

packages/docs/stories/components/range.stories.ts
mariohamann: No, the knob would overlap the label, which makes the tick align with the label as well.

schilchSICKAG: Get it, I will see what I can do :)

packages/docs/stories/components/range.stories.ts
schilchSICKAG: This is now fixed. It was lazy programming, not the event type itself :). Also, with the latest changes it also works with keyboard navigation (which was previously unaccounted)

mariohamann:
@klaarseSICKAG Having in mind, that we try stick to Shoelace's naming as far as possible, especially having in mind, that we maybe want to try to bring it in there as well, I believe we should stick to `thumb-size`.

https://shoelace.style/components/range#custom-properties

Regarding `thumb-clickable-area`: `hit-area-scale` feels somehow made it up to me, especially the term `scale`. Although would be correct from a "typing" perspective, I believe `hit-area-size` would feel more intuitive, as we want to influence the size – and if it's stated in the description, that it refers to a scaling factor, I think that's enough. It feels a bit off to reduce the `thumb-` part, but you#re right, that it feels not correct, if it affects both areas. Maybe therefore it should be separated into `thumb-hit-area-size` and `track-hit-area-size`?

Not having big thoughts on `track-active-offset` etc. – but I believe it should be totally clear, that the offset is from the left, as the track always starts there. (And sometimes even trial and error is okay. 😆)

schilchSICKAG:
> @klaarseSICKAG Having in mind, that we try stick to Shoelace's naming as far as possible, especially having in mind, that we maybe want to try to bring it in there as well, I believe we should stick to `thumb-size`.
> 
> https://shoelace.style/components/range#custom-properties
> 
> Regarding `thumb-clickable-area`: `hit-area-scale` feels somehow made it up to me, especially the term `scale`. Although would be correct from a "typing" perspective, I believe `hit-area-size` would feel more intuitive, as we want to influence the size – and if it's stated in the description, that it refers to a scaling factor, I think that's enough. It feels a bit off to reduce the `thumb-` part, but you#re right, that it feels not correct, if it affects both areas. Maybe therefore it should be separated into `thumb-hit-area-size` and `track-hit-area-size`?
> 

I think that is actually a good idea. Would go with `thumb-hit-area-size` and `track-hit-area-size`.

> Not having big thoughts on `track-active-offset` etc. – but I believe it should be totally clear, that the offset is from the left, as the track always starts there. (And sometimes even trial and error is okay. 😆)

The `track-*` values are strictly copy and pasted from the shoelace implementation, including their documentation. I would stick with the naming, but have adjusted the doc to make clear it starts from the left.

klaarseSICKAG:
Notices an issue with real touch devices (iphone 14). Releasing the knob after staying on one spot for about 1s leads to a jumping knob. It doesn't use the position the touch is released but uses a random position. If this issue is too hard to tackle a follow up issue would be ok.

mrjfalk:
Overall looks promising, a few things I noticed when I tested them in Storybook:

- Two thumbs with restrictions: Moving the thumbs quickly towards each other causes them to get "stuck" before it actually reaches the other thumb.
- Touch target of thumbs are too large, even when at the following spacing, if the cursor comes from the right direction (so it at least once touches the right thumb), then you can have the cursor completely over the left thumb, but the right thumb is still being active:
 
image

- Tooltip can get hidden under other thumbs:   
image

- The tooltip is the same size for all sizes

kirchsuSICKAG:
@mrjfalk Thank you very much for your feedback!

> * Two thumbs with restrictions: Moving the thumbs quickly towards each other causes them to get "stuck" before it actually reaches the other thumb.

This is now fixed

> * Touch target of thumbs are too large, even when at the following spacing, if the cursor comes from the right direction (so it at least once touches the right thumb), then you can have the cursor completely over the left thumb, but the right thumb is still being active:

We adapted it so the touch area has the size of the hover scale area.

> * Tooltip can get hidden under other thumbs:

This is now fixed

> * The tooltip is the same size for all sizes

This behavior is on purpose. On the one hand there are no different tooltip sizes, on the other hand from ux perspective it is better readable for the user (in case of the small size).

kirchsuSICKAG (approved):

PR #552, opened 2024-07-12
docs: add codepen button

mariohamann:
# Pull Request

## 📖 Description

<video src="https://github.com/user-attachments/assets/3816f0af-70d3-436d-84b5-8f6bc82ff626"></video>

Currently missing:
- [x] I'm currently not able to preserve the edited code, as this button is removed by the second code area. It actually WOULD work, but I can't find a way to distinguish if it's the first or the second code view. Maybe someone has an idea?
- [x] We should use the actual version number in the esm.sh-links. I'm not sure if we already have it available from Vite side, but that should be easily doable.
- [x] assets, theme, utilities and components are included with correct versions
- [ ] Little clean up is definitely possible, maybe even having a separate file for that.

## ✅ DoD


mariohamann:
Everything works now. Maybe some cleanup is needed. I would give this PR now to @kirchsuSICKAG and @schilchSICKAG if that's fine? One addition feature wise: Could be nice, to add the correct font, so that everything looks a bit nicer? :)

schilchSICKAG:
> Could be nice, to add the correct font, so

Will surely have a look into this later this day, thank you :)

packages/docs/.storybook/preview.ts
```diff
@@ -92,4 +189,4 @@ const preview: Preview = {
   },
 };
 
-export default preview;
+export default preview;;
```
kirchsuSICKAG, line 192: A second `;` has sneaked in

```suggestion
export default preview;
```

packages/docs/src/vite-plugin-package-versions/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  return {
+    name: 'vite-plugin-package-versions',
+    configResolved(config) {
+      const versions = options.packagePaths.reduce((acc, packagePath) => {
+        try {
+          const packageJson = JSON.parse(
+            readFileSync(join(config.root, packagePath, 'package.json'), 'utf-8'),
+          );
+          acc[packageJson.name] = packageJson.version;
+          console.log(acc);
```
kirchsuSICKAG, line 19: I guess the `console.log` can be removed / is a leftover of creating this PR

```suggestion
```

packages/docs/.storybook/preview.ts
```diff
@@ -70,7 +72,102 @@ const preview: Preview = {
       toc: {
         headingSelector: 'h2, h3',
       },
-      source: { format: 'html' }
+      source: {
+        format: 'html', transform: (code: string, storyContext: StoryContext) => {
```
kirchsuSICKAG, line 76: I think it would be nicer / a little bit more clearer, if we would outsource this functionality into its own file, so the `preview.ts` will not become too overloaded

kirchsuSICKAG (changes requested):

kirchsuSICKAG:
> Could be nice, to add the correct font, so that everything looks a bit nicer? :)

I finished all so far, what I had as review comments and also added the font. 
From my side it, everything looks good

kirchsuSICKAG (approved):

PR #563, opened 2024-07-22
feat: ✨ create syn-file

### LINKED ISSUE #402: feat: ✨ create syn-file

# Component

I suggest to add the feature to the following component: syn-file

# Story

As a user of the Synergy Design System, I would like to provide means of selecting file(s) so they can e.g. be uploaded somewhere.

# Acceptance criteria

- [x] Have a button as a trigger for the file chooser
- [x] A file-open dialog is opened once the button is clicked
- [x] Have a drop zone
- [x] The developer can define a file filter (e.g. "*.png|*.jpg")
- [x] A FileList can be obtained by the developer after selection is over
- [x] ~~Event contains selected file(s)~~ Filelist may be obtained via `SynChange.target.files`, as in most of our other elements.
- [ ] shoelace is informed about this component and offered for a merge

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-file>` component that may act as a replacement for `<input type="file">`, as well as a custom dropzone.

The following issues must still be resolved before this can enter testing:

- [x] There is currently no template in Storybook. We will have to work out where we put it. (Will be tackled in #569)
- [x] When using `droparea`, you are always able to upload multiple files. This should be fixed if possible.
- [x] Tests are missing nearly completely
- [x] Must be added to framework demos in a meaningful way.
- [x] Design alignment is not 100% through.
- [x] The component implements `SynergyFormsModule`, but constraint validation is currently an empty shell.
- [x] Expose all events needed (e.g. drag&drop, file)
- [x] There is currently a custom defined `hover` state for the dropzone that has to be aligned with design.
- [x] There is currently no disabled state while drag&drop is in place.
- [x] Storybook Figma links are currently broken and pointing to the wrong story.
- [x] 🔥 Completely compatible with native `<input type="file">` by leveraging `formData` and binary data values

### 🎫 Issues

Closes #402 

## 👩‍💻 Reviewer Notes

- There where a11y errors for the placeholder (contrast ratio of 2.4 instead of 4.5). Has to be talked about with design. I disabled the rule for the time being.
- As there is no possibility to programmatically set the value of a inputs `files` object, there is no way for two way databindings for angular or vue here. This is a limitation of the platform, not of the component itself.

## 📑 Test Plan

- Have a look at the provided stories
- Have a look at the framework demos to see how it can be used in the "real" world.
- Make also sure to test with keyboard navigation.
- Check if the static translations make sense

## ✅ DoD


xkabylgSICKAG:
Looks like there's no support for setting `webkitdirectory`. Will you add support for that? It's something that we would like to use within dStudio and dTag, and not supporting it was one of the reasons why we moved away from the davinci file components.

kirchsuSICKAG:
> Looks like there's no support for setting `webkitdirectory`. Will you add support for that? It's something that we would like to use within dStudio and dTag, and not supporting it was one of the reasons why we moved away from the davinci file components.

This property is now added

mariohamann (approved):

PR #566, opened 2024-07-26
docs: 📚 Revised LIMITATIONS docu

klaarseSICKAG:
# Pull Request

## ✅ DoD

- [x] checked improved documentation

### DIFF
--- a/packages/components/LIMITATIONS.md
+++ b/packages/components/LIMITATIONS.md
@@ -33,6 +33,7 @@ Rendering is a crucial task of current frontend frameworks. Most frameworks will
 ```
 
 #### Solution
+
 ```html
 <syn-button>
   <span>DYNAMIC_TEXT</span>
@@ -80,6 +81,7 @@ In other cases, make sure to suppress the other elements emitted `syn-hide` even
 ```
 
 #### Solution 1
+
 ```html
 <syn-dialog open>
   <syn-details open summary="Example"></syn-details>
@@ -154,6 +156,7 @@ Use `currentTarget` instead, which aims for the element on which the listener is
 ```
 
 #### Solution
+
 ```html
 <syn-nav-item href="/">
   Home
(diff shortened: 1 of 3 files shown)


packages/components/LIMITATIONS.md
```diff
@@ -145,12 +133,11 @@ Synergy components make heavy use of [ShadowDOM](https://developer.mozilla.org/e
 
 ### Proposed Solution
 
-Try to match on the next `<syn-nav-item>` with the closest selector. This will also match for slotted elements.
+Use `currentTarget` instead, which aims for the element on which the listener is attached on instead of the event origin.
```
schilchSICKAG, line 138: This does not always work as the main event may be added to another element higher up in the DOM. For example, we are using a listener like the following to make it work:

```typescript
    const root = document.querySelector('syn-side-nav');
    root?.addEventListener('click', (e: Event) => {
      e.preventDefault();
      e.stopPropagation();

      console.log(e.currentTarget, e.target, e.target.closest('syn-nav-item'));

      const clickedNavItem = (e.target as HTMLElement).closest('syn-nav-item') as SynNavItem;

      if (clickedNavItem) {
        history.push(history.createHref(clickedNavItem.href));
      }
    });
  });
```

In this case, `e.currentTarget` will point to the `<syn-side-nav>`. It is much easier in my opinion to select this via the `closest` selector as people will not always add the event to the associated `<syn-nav-item>` (e.g. for performance-reasons).

What do you think about adding your variant, too as another solution. I can easily see people using both versions (e.g. adding the listener directly to the `<syn-nav-item>` and to the `<syn-side-nav>`.

schilchSICKAG (changes requested):
I am mostly fine. Don´t know for sure why I uppercase any word starting with an "I", seems to be a habit from old times, sorry :). I like that the Problem and Solution are split now, makes copy&paste easier in the first place.

However, I think the click listener example still holds true when using event delegation, so we should maybe add both variants instead of exchanging the one version with the other when both work fine and it depends on the use case when to use which.

packages/components/LIMITATIONS.md
klaarseSICKAG: fixed

PR #586, opened 2024-08-15
feat: ✨ add prefix/suffix slots to the tab-group component

xkabylgSICKAG:
# Pull Request

## 📖 Description

Adds two new slots to the tab-group, prefix and suffix.

This implementation is good enough for what I'd need it for, but there's probably room for improvement. :)

### 🎫 Issues

No relevant issues exist

## 👩‍💻 Reviewer Notes

Do you think it makes more sense for the prefix/suffix to be inside or outside the scrollable area?

Do you think it makes more sense for the prefix/suffix to also have the border-bottom that the tabs do or not?

## ✅ DoD

### DIFF
--- a/packages/angular/src/components/tab-group.component.ts
+++ b/packages/angular/src/components/tab-group.component.ts
@@ -24,12 +24,14 @@ import '@synergy-design-system/components/components/tab-group/tab-group.js';
  *
  * @slot - Used for grouping tab panels in the tab group. Must be `<syn-tab-panel>` elements.
  * @slot nav - Used for grouping tabs in the tab group. Must be `<syn-tab>` elements.
+ * @slot prefix - An optional prefix element that will be shown before the navigation tabs.
+ * @slot suffix - An optional suffix element that will be shown after the navigation tabs.
  *
  * @event {{ name: String }} syn-tab-show - Emitted when a tab is shown. The payload of the event returns the "panel" attribute of the shown tab.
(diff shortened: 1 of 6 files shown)


mariohamann:
Thanks @xkabylgSICKAG for your valuable effort – highly appreciated, that you took the time for this feature!

- While we understand, that this could help you in your specific case, we currently have big hesitation to introduce that especially from a design perspective, as this opens up the door for  patterns, that make it harder to identify the tabs as... tabs. As we're enforcing an unified approach wherever possible, we therefore don't want to introduce this feature in development as well.
- In general it would help us a lot, if you would first open up an issue or even a discussion, describing specific use cases, where you see potential for a feature. This helps us to identify, if we should cover a need as a component feature, a template, a CodePen or not at all.
- We're working with a so co-called `vendorism` approach, therefore the change would have to be done in e. g. [this file](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/scripts/vendorism/custom/tab-group.vendorism.js), as otherwise our test pipeline would fail

All in all we would like to close the PR for the moment. Two recommendations:
1. As already shared internally, this is how you could approach this via CSS: https://codepen.io/mariohamann/pen/jOjYmbp
2. It could make more sense, to use `syn-prio-nav` in combination with `syn-navigation-item` instead of tabs in your cases, see https://synergy-design-system.github.io/?path=/docs/components-syn-prio-nav--docs This makes it already possible to show as many navigation points as possible.

PR #589, opened 2024-08-20
feat: ✨ link css

### LINKED ISSUE #405: feat: ✨ link css

# Story

As a user of the Synergy Design System, I would like to have support for links in a document.

# Acceptance criteria

- [x] Styling for links are provided in the styles package (default, hover, visited, etc.)
- [x] icons have same color as text
- [ ] add entry to migration guide

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds new `link` related styles to the `@synergy-design-system/styles` package. It provides the following classes:

- `syn-link` (base): Adds typography and color definitions for regular links
- `syn-link--small` | `syn-link-medium` | `syn-link-large`: Define different sizes for links
- `syn-link--quiet`: Dimmed variant
- `syn-link--disabled`: Disabled version for links. As this is not possible via css, please make sure your link is **really disabled**, e.g. via a click listener. Also add `tabindex="-1" to make sure it is not selectable via keyboard interaction!

### 🎫 Issues

Closes #405 

## 👩‍💻 Reviewer Notes

Per default, the link uses a css display value of `inline-flex` to style links that have suffix and/or prefix icons. As defined in our story "Inline" links that do not have a `syn-icon` use a display value of `inline` to make them usable in regular text. This is done via a css `:has` selector. Support is pretty good nowadays, so it should not matter for our audience. See https://caniuse.com/?search=has for details.

The figma component seems to have a small padding in the component. I have omitted this padding to make sure it does not conflict with the flow of text when using the inline style.

## 📑 Test Plan

- Have a look at the stories and just play around with them.

## ✅ DoD


packages/styles/src/link/index.css
```diff
@@ -0,0 +1,82 @@
...
+  color: var(--syn-link-color);
+  display: inline-flex;
+  font: var(--syn-body-medium-regular);
+  gap: var(--syn-spacing-x-small);
+}
+
+/* Use inline text styling when there is no syn-icon */
+.syn-link:not(:has(syn-icon)) {
+  display: inline;
+}
```
mariohamann, line 18: I would do it the other way round:

```suggestion
.syn-link {
  color: var(--syn-link-color);
  font: var(--syn-body-medium-regular);
}

/* Use inline text styling when there is no syn-icon */
.syn-link:has(syn-icon) {
  display: inline-flex;
  align-items: center;
  gap: var(--syn-spacing-x-small);
}
```

packages/styles/src/link/index.css
```diff
@@ -0,0 +1,82 @@
...
+.syn-link:not(.syn-link--disabled):focus-visible {
+  outline: var(--syn-focus-ring);
+  outline-offset: var(--syn-focus-ring-offset);
+}
+
+/* Overrides for smaller links */
+.syn-link--small {
+  font: var(--syn-body-small-regular);
+  gap: var(--syn-spacing-2x-small);
+}
```
mariohamann, line 37: The icons are way bigger in design.

```
.syn-link--small syn-icon {
    font: var(--syn-spacing-medium);
}
```
Maybe this would fit. Check out the sizes in Figma.

packages/styles/src/link/index.css
```diff
@@ -0,0 +1,82 @@
...
+ * Link styles
+ * @variant {small | medium | large} syn-link The size of the link
+ * @boolean { false } syn-link--quiet Applies the quiet link styling
+ * @boolean { false } syn-link--disabled Applies the disabled link styling
+ */
+.syn-link {
+  align-items: center;
+  color: var(--syn-link-color);
+  display: inline-flex;
+  font: var(--syn-body-medium-regular);
```
mariohamann, line 11: Per default we should neither have a size or a weight here. If someone drops a syn-link inside a text, it should just apply the blue color and underline, but not change any size. And if someone wants to make it bold, it should just work.

packages/styles/src/link/index.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/*
+ * Overrides for medium links.
+ * Note: The default styles are already set for medium links, just here for completeness
+ */
+.syn-link--medium {
+  font: var(--syn-body-medium-regular);
+  gap: var(--syn-spacing-x-small);
+}
```
mariohamann, line 46:
```suggestion
.syn-link--medium {
  font: var(--syn-body-medium-regular);
  gap: var(--syn-spacing-x-small);
}
```

We should not use `--syn-body-medium-regular` but `--syn-font-size-medium` here (as done in design).

Otherwise this would lead to:
```html
<div class="font-bold">
Reprehenderit aliquip pariatur <a href="syn-link"> proident nulla fugiat Lorem ex.
</div>
```

Here the link would not be bold anymore.

packages/styles/src/link/index.css
```diff
@@ -0,0 +1,82 @@
...
+  color: var(--syn-link-low-emphasis-color);
+}
+
+.syn-link--quiet:hover {
+  color: var(--syn-link-low-emphasis-color-hover);
+}
+
+.syn-link--quiet:active {
+  color: var(--syn-link-low-emphasis-color-active);
+}
```
mariohamann, line 64:
```suggestion
.syn-link--quiet {
  color: var(--syn-link-quiet-color);
}

.syn-link--quiet:hover {
  color: var(--syn-link-quiet-color-hover);
}

.syn-link--quiet:active {
  color: var(--syn-link-quiet-color-active);
}
```

mariohamann: THis should not work on disabled.

packages/styles/src/link/index.css
```diff
@@ -0,0 +1,82 @@
...
+ * We set pointer-events to none to prevent the link from being clicked
+ * This is done in :active to allow using a custom cursor
+ */
+.syn-link--disabled:active {
+  pointer-events: none;
+}
+
+.syn-link--disabled:focus {
+  outline: none;
+}
```
mariohamann, line 82:
```suggestion
/* Styling for disabled links */
.syn-link--disabled:not([href]), a.syn-link:not([href]) {
  cursor: not-allowed;
  opacity: var(--syn-input-disabled-opacity);
}
```

Your suggestion would lead to inaccessible HTML, as the element would still be tabbable.

I suggest to make this only werk if no `href` is set and in addition to make it more convenient that dropping `href` on a link makes it disabled out of the box.

packages/tokens/src/figma-tokens/semantic/link.json
```diff
@@ -0,0 +1,30 @@
...
+      "type": "color"
+    },
+    "color-hover": {
+      "value": "{primary.900}",
+      "type": "color"
+    },
+    "color-active": {
+      "value": "{primary.950}",
+      "type": "color"
+    },
```
mariohamann, line 14: I believe this should link to the button default/hover/active colors, right? @LeaLeRose

packages/tokens/src/figma-tokens/semantic/link.json
```diff
@@ -0,0 +1,30 @@
...
+      "color-hover": {
+        "value": "{primary.900}",
+        "type": "color"
+      },
+      "color-active": {
+        "value": "{primary.950}",
+        "type": "color"
+      }
+    }
+  }
```
mariohamann, line 29:
```suggestion
    "quiet": {
      "color": {
        "value": "{typography.color.text}",
        "type": "color"
      },
      "color-hover": {
        "value": "{primary.900}",
        "type": "color"
      },
      "color-active": {
        "value": "{primary.950}",
        "type": "color"
      }
    }
  }
```

packages/docs/stories/styles/link.stories.ts
```diff
@@ -0,0 +1,180 @@
...
+      description: {
+        story: generateStoryDescription('link', 'disabled'),
+      },
+    },
+  },
+  render: () => html`
+    <a href="javascript:alert('Here')" class="syn-link syn-link--disabled">
+      <syn-icon name="keyboard_arrow_right"></syn-icon>
+      Link Label
+    </a>
```
mariohamann, line 82:
```suggestion
    <a class="syn-link syn-link--disabled">
      <syn-icon name="keyboard_arrow_right"></syn-icon>
      Link Label
    </a>
```

packages/styles/src/link/index.css
schilchSICKAG: Yes, there where no docs yesterday, adjusted this now

packages/styles/src/link/index.css
schilchSICKAG: I don´t have a strong opinion on this, I am fine with it either way, so just changed it to your suggestion.

packages/styles/src/link/index.css
schilchSICKAG: You are right, I did not think about this. Adjusted.

packages/styles/src/link/index.css
schilchSICKAG: Sounds like a good idea!

packages/tokens/src/figma-tokens/semantic/link.json
schilchSICKAG: Done. Must be updated on Figma, too if not yet done

packages/styles/src/link/index.css
schilchSICKAG: Done by using font-size / font-family

schilchSICKAG:
@mariohamann: Thanks for the input. I have made a couple of adjustments that should make it more stable and reliable. Only thing that is not doable right now is providing a working docs version of disabled in the default story (as it always has a href per default, disabled knob will not do anything). Please have another look and tell me what you think.

packages/tokens/src/figma-tokens/semantic/link.json
schilchSICKAG: It does. However, we do not have semantic tokens for the button as of current, therefore we use the primitives here.

packages/styles/src/link/index.css
```diff
@@ -0,0 +1,87 @@
...
+ * @variant {small | medium | large} syn-link The size of the link
+ * @boolean { false } syn-link--quiet Applies the quiet link styling
+ * @boolean { false } syn-link--disabled Applies the disabled link styling
+ */
+.syn-link {
+  /* The size of the icon */
+  --syn-link-icon-size: var(--syn-spacing-large);
+
+  /* The gap that is used when there are pre or suffix icons */
+  --syn-link-icon-gap: var(--syn-spacing-x-small);
```
mariohamann, line 12: On inheritat, we need something that uses em, as otherwise it maybe doesn't fit together.

schilchSICKAG: Inherit should use the latest font-size defined by an upper element, I do not see any issues with using it here? See https://codepen.io/schilchSICKAG/pen/PoraZyg for a demo.

mariohamann: CleanShot 2024-08-26 at 08 05 46@2x

The icons in inherit have to adapt proportionally as well. Otherwise it will look like... this.

schilchSICKAG: Ah, now I see what you mean :). Fixed in latest version. I adjusted both default values to `1em` to make it more explicit. I have played around with relative sizes (e.g. a default of `1.2em`), but that does not work out any time as icon sizes for defined size properties are not strictly aligned. Thanks for the catch.

packages/styles/src/link/index.css
```diff
@@ -6,13 +6,13 @@
  */
 .syn-link {
   /* The size of the icon */
-  --syn-link-icon-size: var(--syn-spacing-large);
+  --syn-link-icon-size: 1em;
 
   /* The gap that is used when there are pre or suffix icons */
   --syn-link-icon-gap: var(--syn-spacing-x-small);
```
mariohamann, line 12: Can you please add a default gap according to `em` as well, pls? :)

schilchSICKAG: Sure, just set it to `0.25em`.

**Please note that the maximal amount of tokens for ChristianGTP is reached. Payment possible via Coffee** 😜

mariohamann (approved):

PR #599, opened 2024-08-26
feat: ✨ provide accessible solution for validation

### LINKED ISSUE #412: feat: ✨ provide accessible solution for validation

# Component

I suggest to add the feature to the following component: all input elements

# Story

As a user I want to get informed about invalid input in an accessible way.

Have a look at https://shoelace.style/getting-started/form-controls#inline-form-validation

# Acceptance criteria

- [ ] Provide documentation for validation of form-controls (see shoelace docs: https://shoelace.style/getting-started/form-controls)
- [ ] Wrapper component (e.g. <syn-validation><syn-input></syn-input></syn-validation>) takes care of when to trigger the validation (e.g. on change or input)
- [ ] Adjust framework demos to make use of it.
- [ ] Add note in Angular Limitations that ngNativeValidate must be set
- [ ] Try adding this feature into Shoelace

### LINKED ISSUE #623: docs: 📚 syn-header focus story is broken
# Description
The syn-header focus story looks different in safari. It has much more whitespace on the left side of the sick logo.

Appearance in safarI:
Bildschirmfoto 2024-09-19 um 13 17 49

What it should look like (screenhotted in chrome): 
image

Additionally to the different appearance, there is another problem. The logo is not focusable in safari. It seems like an `<a/>` tag in safari is not focusable per default, which behaves different than in chrome and firefox. To make an `<a/>` tag focusable in safari a tabindex need to be added. 
Have a look at following codepen and try it out in the different browsers:
https://codepen.io/kirchsuSICKAG/pen/JjQgrbr

# Rationale
The template does not behave correctly in all 3 major browsers, what it should!

# Acceptance criteria

- [ ] sick logo is focusable in safari (for the focus story)
- [ ] the focus story looks the same in all 3 major browsers

schilchSICKAG:
# Pull Request

## 📖 Description

This PR provides a new component `<syn-validate>` that may be used to wrap native input elements and SynergyFormElements to provide inline validation.

### 🎫 Issues

Closes #412 
Closes #623 

## ✅ DoD


packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,268 @@
...
+   *
+   * <!-- Validate on live and custom events (invalid, focus, input, change) -->
+   * <syn-validate on="live focus change"></syn-validate>
+   * ```
+   */
+  @property({
+    converter: {
+      fromAttribute: (e: string) => e.split(' ').map(s => s.trim()),
+      toAttribute: (a: string[] | string) => {
+        if (!Array.isArray(a)) return a;
```
kirchsuSICKAG, line 71: If the input given to `on` as property is a string, it should not just be returned but also the same should be done as in the fromAttribute `a.split(' ').map(s => s.trim())` before returning it.

Because as a user, when I see the description of the `on` property I would assume I can either add a string oder a string array value.

image

Everything works fine if adding only a single event as string:

```js
      const validate = document.querySelector('syn-validate');
      setTimeout(() => {
        validate.on = 'change';
      }, 3000);
```

The events array will then look like following which is correct: ` ['syn-change', 'syn-invalid']`

But if adding more than one event as a string like following:

```js
      const validate = document.querySelector('syn-validate');
      setTimeout(() => {
        validate.on = 'change blur';
      }, 3000);
```

The events array will then look like following which is incorrect: ` ['syn-change blur', 'syn-invalid']`

packages/docs/src/validate-demo-radio.ts
```diff
@@ -0,0 +1,176 @@
...
+  checkValidity() {
+    return this.internals.checkValidity();
+  }
+
+  reportValidity() {
+    return this.internals.reportValidity();
+  }
+
+  get validity() {
+    console.log('here');
```
kirchsuSICKAG, line 127: console.log should be removed 😉

```suggestion
```

packages/_private/angular-demo/src/app.component.html
```diff
@@ -1,4 +1,13 @@
-<syn-header label="synergy-design-system/angular Form Demo">
+<syn-header label="synergy-design-system/angular Components Demo">
+  <a
+    class="custom-logo"
+    href="/"
+    routerLink="/"
+    slot="logo"
+  >
+    <syn-icon name="logo-color" library="system" />
+  </a>
```
kirchsuSICKAG, line 9: This (using an `<a/>` tag around a syn-icon in the `logo` slot) makes the header look wrong with safari. There is too much space at the left side of the logo: 
image

kirchsuSICKAG (changes requested):

packages/_private/react-demo/src/Layout.tsx
```diff
@@ -22,7 +22,15 @@ export const Layout: FC = () => {
...
     <>
-      <SynHeader label="@synergy-design-system/react Form Demo">
+      <SynHeader label="@synergy-design-system/react Components Demo">
+        <Link
+          className="custom-logo"
+          slot="logo"
+          to="/"
+        >
+          <SynIcon name="logo-color" library="system" />
+        </Link>
```
kirchsuSICKAG, line 32: This (using an <a/> tag around a syn-icon in the logo slot) makes the header look wrong with safari. There is too much space at the left side of the logo. See image in angular demo

kirchsuSICKAG (changes requested):

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  private renderInlineValidation() {
+    if (!this.inline || !this.validationMessage) {
+      return '';
+    }
+
+    return html`
+      <syn-alert
+        open
+        exportparts="base:alert,message:alert-message"
```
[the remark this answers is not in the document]
schilchSICKAG: Fixed in latest

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  /**
+   * Triggers a validation run, showing the validation message if needed.
+   */
+  private validate = (e: Event) => {
+    // Make sure to always prevent the invalid event when using inline validation
+    if (isInvalidEvent(e.type) && this.inline) {
+      e.preventDefault();
+      e.stopPropagation();
```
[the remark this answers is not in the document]
schilchSICKAG: Problem is you have to stop it, otherwise you will trap the input in case of inline validation. As the event uses event capturing (e.g. the event flows from "the top down" instead of default "from the bottom up" to inform the parent form, it is still possible to listen to the event.

When throwing an invalid event on native inputs, the input will automatically receive focus, this is why this was done.

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: After thinking deeper about it, I now use a string always. All frameworks have custom syntax making it unneccesairy hard to use this. Also, we can get rid of most of the logic when sticking to a simple string.

packages/docs/src/validate-demo-radio.ts
schilchSICKAG: :)

packages/_private/angular-demo/src/demoformvalidate/demoformvalidate.component.ts
```diff
@@ -0,0 +1,63 @@
+import type { SynChangeEvent, SynRange } from '@synergy-design-system/components';
+import { serialize } from '@synergy-design-system/components';
+import { Component, ElementRef, ViewChild } from '@angular/core';
+import { FormBuilder, FormGroup } from '@angular/forms';
+
+const initialData = {
+  code: '',
+  comment: '',
+  date: '',
```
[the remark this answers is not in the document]
schilchSICKAG: This is a pretty good idea. I have just added this to the angulars e-mail field to showcase a required or e-mail field.

packages/_private/vue-demo/src/DemoFormValidate.vue
```diff
@@ -0,0 +1,328 @@
+<script setup lang="ts">
```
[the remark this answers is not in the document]
schilchSICKAG: Thats nice to know. The hard part of using libraries (e.g. vue, react) is that there are myriads of different solutions out there. I thought about adding this to the known issues/limitations, but this is not quite fitting. For this reason, I leave this stand as is. Maybe if there is a need, we will add documentation for it based on your findings.

packages/_private/angular-demo/src/app.component.html
schilchSICKAG: Fixed. We did not get this initially as Storybook was not tested with Safari when we created the header.

packages/_private/vue-demo/src/App.vue
```diff
@@ -1,29 +1,39 @@
...
   router.push(route);
 }
 </script>
 
 <template>
-  <SynVueHeader label="@synergy-design-system/vue Form Demo">
+  <SynVueHeader label="@synergy-design-system/vue Components Demo">
+    <RouterLink class="custom-logo" to="/" slot="logo">
+      <SynVueIcon name="logo-color" library="system" />
+    </RouterLink>
```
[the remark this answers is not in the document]
schilchSICKAG: Fixed. We did not get this initially as Storybook was not tested with Safari when we created the header.

packages/_private/react-demo/src/Layout.tsx
schilchSICKAG: Fixed. We did not get this initially as Storybook was not tested with Safari when we created the header.

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,268 @@
...
+
+  static dependencies = {
+    'syn-alert': SynAlert,
+  };
+
+  controller = new AbortController();
+
+  @queryAssignedElements() private slottedChildren: HTMLElement[];
+
+  @state() validationMessage = '';
```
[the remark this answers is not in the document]
schilchSICKAG: Thats a good idea, just implemented it.

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  /** Show the validation message underneath the element, using a syn-alert */
+  @property({ reflect: true, type: Boolean }) inline = false;
+
+  /** Do not show the error icon when using inline validation */
+  @property({ attribute: 'hide-icon', reflect: true, type: Boolean }) hideIcon = false;
```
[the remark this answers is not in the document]
schilchSICKAG: Bildschirmfoto 2024-10-02 um 13 18 09

See attached image. As stated in the documentation, this hides the icon in the `<syn-alert>` displayed when using `<syn-validate inline>`

packages/docs/stories/components/validate.stories.ts
```diff
@@ -0,0 +1,201 @@
...
+    <syn-validate
+      class="validation-custom-form-field"
+      inline
+      on="live"
+    >
+      <validate-demo-radio name="color"></validate-demo-radio>
+    </syn-validate>
+  `,
+};
+
```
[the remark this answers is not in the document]
schilchSICKAG: Sure, we could add this as a dev story.

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  /**
+   * Defines the events that trigger the validation.
+   * Defaults to the `invalid` and `change` events.
```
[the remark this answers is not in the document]
schilchSICKAG: Leftover from the original version, I have cleaned up the comment.

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,288 @@
...
+   * `invalid` will always automatically be included.
+   * You may also use the `live` keyword to validate on every input change.
+   * `live` will make sure to listen to the `invalid`, `input` and `blur` events.
+   * @example ```html
+   * <!-- Use default validation (invalid) -->
+   * <syn-validate></syn-validate>
+   *
+   * <!-- Validate on invalid and change events (invalid, change) -->
+   * <syn-validate on="invalid change"></syn-validate>
```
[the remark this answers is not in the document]
schilchSICKAG: Was there for the sake of documentation only, to have it in the same format. Removed as it is duplicated

packages/docs/stories/components/validate.stories.ts
```diff
@@ -0,0 +1,201 @@
...
+      <syn-input
+        label="Inline Validation"
+        type="email"
+        value="team(at)synergy.com"
+      ></syn-input>
+    </syn-validate>
+  `,
+};
+
+export const Live: Story = {
```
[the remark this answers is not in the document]
schilchSICKAG: Not exactly sure what you mean here? The Live story shows how to use `<syn-validate>` to live validate a `<syn-radio-group>` and how this shows the inline error message. The default story does neither use inline error messages nor live validation (which can be seen when leaving the default story after entering something invalid).

packages/docs/stories/components/validate.stories.ts
```diff
@@ -0,0 +1,201 @@
...
+    docs: {
+      description: {
+        story: generateStoryDescription('validate', 'custom-validation'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-validate
+      class="validation-custom-validation"
+      custom-validation="Include an &quot;@&quot; in the email address, otherwise you will never get our marvelous newsletter"
```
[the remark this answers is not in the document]
schilchSICKAG: Bildschirmfoto 2024-10-02 um 13 31 12

You will have to trigger validation. As we want to showcase the validation and when it is happening, I have opted not to show the error itself per default. So just focusing the Story and pressing enter (or doing a keystroke when "live" is used) should be enough to show the error message.

packages/vue/src/components/SynVueValidate.vue
```diff
@@ -0,0 +1,80 @@
...
+`live` will make sure to listen to the `invalid`, `input` and `blur` events.
+ */
+  on?: SynValidate['on'];
+
+  /**
+* Custom validation message to be displayed when the input is invalid.
+Will override the default browser validation message.
+Set to an empty string to reset the validation message.
+ */
+  customValidation?: SynValidate['customValidation'];
```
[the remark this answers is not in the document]
schilchSICKAG: Adjusted in latest version

kirchsuSICKAG:
Currently if using multiple syn-validate's + syn-input's with the `inline` validation, only the first invalid input shows it's inline error message.
Is this our wanted behavior, so we have the same behavior as the standard browser tooltip? 

image

```html
  <form>
    <syn-validate class="validation-inline" inline > 
      <syn-input label="Inline Validation" type="email" value="team(at)synergy.co" ></syn-input>
    </syn-validate> 
    <syn-validate class="validation-inline-2" inline > 
      <syn-input label="Inline Validation 2" type="email" value="team(at)synergy.co" ></syn-input>
    </syn-validate> 
    <button type="submit">submit</button>
  </form>
```

kirchsuSICKAG:
Another weird behavior is when mixing syn-input's with html input's in a form and pressing the submit button.
The syn-input doesn't seem to get validated at all (no red border) and the input gets validated (even if it is the last element)

```html
  <form>
    <syn-validate class="validation-inline" inline > 
      <syn-input label="Inline Validation" type="email" value="team(at)synergy.co" ></syn-input>
    </syn-validate> 
    <syn-validate class="validation-inline-2" inline > 
      <syn-input label="Inline Validation 2" type="email" value="team(at)synergy.co" ></syn-input>
    </syn-validate> 
    <syn-validate class="validation-inline-3" inline > 
      <input type="email" value="team(at)synergy.co"/>
    </syn-validate> 
    <button type="submit">submit</button>
  </form>
````

image

**Comment by schilchSICKAG** [comment]
> Currently if using multiple syn-validate's + syn-input's with the `inline` validation, only the first invalid input shows it's inline error message. Is this our wanted behavior, so we have the same behavior as the standard browser tooltip?
> 
> image
> 
> ```
>   <form>
>     <syn-validate class="validation-inline" inline > 
>       <syn-input label="Inline Validation" type="email" value="team(at)synergy.co" ></syn-input>
>     </syn-validate> 
>     <syn-validate class="validation-inline-2" inline > 
>       <syn-input label="Inline Validation 2" type="email" value="team(at)synergy.co" ></syn-input>
>     </syn-validate> 
>     <button type="submit">submit</button>
>   </form>
> ```

As far as I can tell, this is the behavior that is used by shoelaces internal implementation, so we are just hooking into this. As we have to cancel the `syn-invalid` event to prevent the item to always get a highlight, it seems the only way to be sure that it works.

**Comment by schilchSICKAG** [comment]
> Another weird behavior is when mixing syn-input's with html input's in a form and pressing the submit button. The syn-input doesn't seem to get validated at all (no red border) and the input gets validated (even if it is the last element)
> 
> ```
>   <form>
>     <syn-validate class="validation-inline" inline > 
>       <syn-input label="Inline Validation" type="email" value="team(at)synergy.co" ></syn-input>
>     </syn-validate> 
>     <syn-validate class="validation-inline-2" inline > 
>       <syn-input label="Inline Validation 2" type="email" value="team(at)synergy.co" ></syn-input>
>     </syn-validate> 
>     <syn-validate class="validation-inline-3" inline > 
>       <input type="email" value="team(at)synergy.co"/>
>     </syn-validate> 
>     <button type="submit">submit</button>
>   </form>
> ```
> 
> image

I think this stems from the issue that Shoelace components are not fully form associated controls. There are other open issues at the shoelace repo for this, like https://github.com/shoelace-style/shoelace/issues/2139. Basically they override the native form behavior to be part of the form submission. The elements are not validated in order of appearance, but but the internal shoelace logic. Most of the time this works, but there are circumstances when it does not seem to work well.

**Review by kirchsuSICKAG — CHANGES_REQUESTED** [review]

File: packages/components/src/components/validate/validate.component.ts [hunk 18]
```diff
@@ -0,0 +1,284 @@
...
+   *
+   * <!-- Validate on live events (invalid, blur, input)-->
+   * <syn-validate on="live"></syn-validate>
+   *
+   * <!-- Validate on live and custom events (invalid, blur, input, focus, change) -->
+   * <syn-validate on="live focus change"></syn-validate>
+   * ```
+   */
+  @property({
+    converter: onConverter(),
```
kirchsuSICKAG, line 70: I would remove the attribute converter and just work completely with strings. The conversion can then be done on the specific places, where it is needed (I think it is only needed in the `handleListenerChange` line 87)

Because there are some strange behaviors / something is not completely correct.

Following html results in following DOM structure:

```html
<syn-validate>
      <syn-input></syn-input>
</syn-validate>

<syn-validate on="change blur">
      <syn-input></syn-input>
</syn-validate>
```

image

So the array is shown in the dom (see the comma separation).


Also the arguments of the `handleListenerChange`  are never from type `string []` but always from type `string`. No matter if changing it via attribute or via property (see both gifs). I am logging the `typeof` of the arguments 
image

validate-attribute-change

validate-change-prop

kirchsuSICKAG (changes requested):

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: I think that I might have messed up the `fromAttribute` conversion. Makes sense to fix this

mariohamann:
For everything that's not live, could you add a submit button?

Just tested on mobile (as I've sent the PR someone) and it's annoying to have to go intonthe field to activate it

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,284 @@
...
+    // live always means on input and blur
+    if (events.includes('live')) {
+      events.push('input');
+      events.push('blur');
+    }
+
+    // Make sure to remove duplicated events and the live property
+    // and map the events to the correct event names
+    return Array
+      .from(new Set(events.filter(e => e !== 'live')))
```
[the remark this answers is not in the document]
schilchSICKAG: Done in latest version

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: This is now always a string. Also makes it more convenient to set in Angular or React, because both would have needed something like `<SynValidate on={['a', 'b']} />`, which was kind of arkward.

schilchSICKAG:
> For everything that's not live, could you add a submit button?
> 
> Just tested on mobile (as I've sent the PR someone) and it's annoying to have to go intonthe field to activate it

Done in latest version

packages/docs/stories/components/validate.stories.ts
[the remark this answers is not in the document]
schilchSICKAG: We changed this to the default story now, using e-mail validation

packages/docs/stories/components/validate.stories.ts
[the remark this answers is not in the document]
schilchSICKAG: Done in latest version (dev only, not exported to chromatic screenshots)

kirchsuSICKAG (approved):

kirchsuSICKAG (approved):

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,285 @@
...
+} from './utility.js';
+import styles from './validate.styles.js';
+
+/**
+ * @summary Validate is a helper that may be used to wrap
+ * synergy input fields and forms to provide validation message.
+ *
+ * @dependency syn-alert
+ *
+ * @slot - The input field or form element to be validated.
```
klaarseSICKAG, line 22: whats the difference between input filed and form element?

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,285 @@
...
+   * - **inline**: Show the validation message underneath the element, using a `<syn-alert>`
+   */
+  @property({ reflect: true }) variant: 'native' | 'inline' = 'native';
+
+  /** Do not show the error icon when using the inline variant validation */
+  @property({ attribute: 'hide-icon', reflect: true, type: Boolean }) hideIcon = false;
+
+  /**
+   * Defines the events that trigger the validation.
+   * `invalid` will always automatically be included.
```
klaarseSICKAG, line 59: maybe its good to link to msdn (https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/invalid_event) so people know it's a standard html feature ...

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,285 @@
...
+   *
+   * <!-- Validate on live and custom events (invalid, blur, input, focus, change) -->
+   * <syn-validate on="live focus change"></syn-validate>
+   * ```
+   */
+  @property({ reflect: true }) on: string = '';
+
+  /**
+   * Custom validation message to be displayed when the input is invalid.
+   * Will override the default browser validation message.
```
klaarseSICKAG, line 77: but only if type=inline

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,285 @@
...
+import {
+  getEventNameForElement,
+  isBlurEvent,
+  isInvalidEvent,
+  normalizeEventAttribute,
+} from './utility.js';
+import styles from './validate.styles.js';
+
+/**
+ * @summary Validate is a helper that may be used to wrap
```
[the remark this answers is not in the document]
schilchSICKAG: No, however it is optional. As we do not support slotting forms at all, I changed the summary to "__Validate provides form field validation messages in a unified way__". Maybe this fits better?

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: This is now directly linked and available in Storybook, as well as in type information.

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: You are right. This is not on purpose. I think we should set the message always as this results in two different outputs. Will adjust this!

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,285 @@
...
+ * synergy input fields and forms to provide validation message.
```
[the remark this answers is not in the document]
schilchSICKAG: Tried to clarify it with the latest version:

> Validate provides form field validation messages in a unified way.
> It does this by using [the native browser validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation) and showing the validation message in a consistent, user defined way.

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: Fixed. The custom message will now always override the original one and also happily switch between browser and inline validation. When reset, it will use the original validation message from the input.

packages/components/src/components/validate/validate.component.ts
```diff
@@ -0,0 +1,321 @@
...
+  /**
+   * Set this to true to validate the input immediately when it is rendered.
+   * Best used with a `variant` of `inline`.
+   * When setting eager, the input will not be focused automatically.
+   *
+   * When using a `variant` of `native` the browser will focus
+   * the last eager field as it is using a tooltip.
+   * In this case it is better to just provide one eager field.
+   */
+  @property({ type: Boolean }) eager = false;
```
kirchsuSICKAG, line 98: I tried out setting the `eager` prop after initialization.
I would have expected, that the validation would trigger and show the error message, but nothing happened.

I'm not sure, if this is also the behavior, we really want, but I just wanted to share my expectation on this.
What do you mean, @schilchSICKAG ?

kirchsuSICKAG (changes requested):

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: You are right, will add this, too 👍

schilchSICKAG: Fixed in latest

huberpaSICKAG:
Hi, I just played a little with the new syn-validate component and want to share my feedback with you: 

* Works quite nicely with HTML5 client side validation and I like the feature that you can determine for which events the validation message is shown. 
* I noticed that when I use `on=live` the focus outline stays blue even when the value is invalid and only turns red after losing focus. I'd expect to turn red immediately on input. 

Aside from this "technical" feedback, here some further thoughts that would make syn-validate better suited for our application (SICK ID): 

* For our tech-stack (angular) it would be nice to have a way to trigger the validation/display of the validation message via javascript. This way the syn-validate component could better be integrated with angular's validation module. (Or maybe there is already a way and I haven't seen it yet?)
* Would it be possible to make the entire container of the inline validation message customizable? In our application we don't want to show alert components but only "red text". 

Thank you for all the effort.

schilchSICKAG:
> * of the inline validation message customizable? In our application we don't

Hi @huberpaSICKAG,

first of all, thank you for your input. I will have a look into it and will respond tomorrow. For Angular, maybe something like in our demo will work ([here](https://github.com/synergy-design-system/synergy-design-system/blob/feat/412-accessible-validation/packages/_private/angular-demo/src/demoformvalidate/demoformvalidate.component.ts#L54) you will find the simple logic for this and [there](https://github.com/synergy-design-system/synergy-design-system/blob/feat/412-accessible-validation/packages/_private/angular-demo/src/demoformvalidate/demoformvalidate.component.html#L63) the view adjustment for this)?

kirchsuSICKAG (approved):

Issue #582, opened 2024-08-13
fix: 🐛 SynSideNav: Parent nav item retains the selected state until being expanded


wujeSICKAG:
# Checklist (for the issue creator)

- [ ] I have verified the bug exists in the latest version of `@synergy-design-system/components` (e.g. via `pnpm install @synergy-design-system/components`).
- [ ] I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...).
- [ ] I would be willing to provide a PR for this issue.

# Version

The bug occurred with the following version of the component library: 2.4.1

# Component

The issue arises when using the following component(s): syn-side-nav

# Environment

The issue occurred with the following operating system(s): all
The issue occurred in the following browser(s): all

# Description

Using syn-side-nav and configure nested nav items in the application, when user clicked on the nested nav item and collapsed the child nav item, the parent nav item is highlighted. The parent nav item will not be de-selected even if user click on other nav item. Until user expand the child nav items.

https://github.com/user-attachments/assets/e83ad408-7af4-4c48-ace7-f26d8903a887

# Acceptance criteria

- [ ] All DoD items successfully completed

Issue #410, opened 2024-04-10
feat: ✨ SynPrioNav: Adjust syn-prio-nav shrinking mechanism


kirchsuSICKAG:
# Checklist (for the issue creator)

- [x] I have looked at the documentation site and was not able to find the required feature.
- [ ] I have looked into the Figma boards and was not able to find the required feature.
- [ ] I would be willing to provide a PR for this feature.

# Component

I suggest to add the feature to the following component: syn-prio-nav

# Story

Currently the syn-prio-nav shows the priority menu only, if there is no space left for any navigation item. If there is space for one navigation item, it shows this navigation item. See following image for the current behavior:
![image](https://github.com/synergy-design-system/synergy-design-system/assets/115139104/ad1fc069-371a-4980-af4e-456aad34618e)

From UX perspective the priority menu only should also be shown if there is only one space left for one navigation item. So in the image above the first syn-prio-nav should look the same like the second with the priority menu only.

# Acceptance criteria

- [x] All DoD items successfully completed

- [x] syn-prio-nav shows priority menu only, if there is space for just one navigation item left
- [x] docs are updated in doc tokens about the behavior

reichst1SICKAG:
Need to check of there is Design Work to do.

Issue #409, opened 2024-04-10
fix: 🤔 syn-nav-item text is not centered if horizontal=true, prefix icon is available and text is too short


kirchsuSICKAG:
# Checklist (for the issue creator)

- [x] I have verified the bug exists in the latest version of `@synergy-design-system/components` (e.g. via `pnpm install @synergy-design-system/components`).
- [ ] I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...).
- [ ] I would be willing to provide a PR for this issue.

# Version

The bug occurred with the following version of the component library: 1.15.0

# Component

The issue arises when using the following component(s): syn-nav-item

# Environment

The issue occurred with the following operating system(s): all
The issue occurred in the following browser(s): all

# Description
When using syn-nav-item with horizontal=true and with a prefix icon, it is possible that the text is not centered to the icon. Depending of the length of the text, the text is either centered (see left syn-nav-item of image) or it is not centered (see right syn-nav-item of image)
![image](https://github.com/synergy-design-system/synergy-design-system/assets/115139104/63d4e941-7ae1-49ab-8280-7cf17779e474)

### Example code
```html
<syn-nav-item horizontal>
      <syn-icon name="wallpaper" slot="prefix" library="default"></syn-icon>
      abcdefghi
</syn-nav-item>

<syn-nav-item horizontal>
      <syn-icon name="wallpaper" slot="prefix" library="default"></syn-icon>
      abc
</syn-nav-item>

```

# Acceptance criteria

- [x] All DoD items successfully completed

- [x] the text is always centered to an icon, no matter how long the text is

PR #608, opened 2024-09-12
fix: 🐛 syn-prio-nav: priority menu position is always calculated from the absolute left of the document

### LINKED ISSUE #639: fix: 🐛 syn-prio-nav: priority menu position is always calculated from the absolute left of the document

# Version

The bug occurred with the following version of the component library: >= 2.0.0

# Component

The issue arises when using the following component(s): syn-prio-nav

# Environment

# Description

The `<syn-prio-nav>` is able to hide all items that may not be visible on the screen in its own priority menu, as seen in https://synergy-design-system.github.io/?path=/docs/components-syn-prio-nav--docs#priority%20menu. However, the calculation is always taken from the top left position of the screen, making it unusable if the navigation is placed somewhere inside a document (e.g. by applying a padding to it).

This was originally found and requested by @xkabylgSICKAG.

Examples: https://codepen.io/xkabylgSICKAG/pen/YzoMjaQ
Original PR: https://github.com/synergy-design-system/synergy-design-system/pull/608

# Acceptance criteria

- [ ] The priority menu should work whereever the `<syn-prio-nav>` is placed.

xkabylgSICKAG:
# Pull Request

## 📖 Description

This fixes an issue that is very clear if the syn-prio-nav is not on the left edge of the viewport, where it will start displaying the dropdown menu a lot earlier than is needed.

It also makes sure to not take into account the size of the dropdown button unless the dropdown button is actually shown.

Also slightly related, the cache should probably be purged if the size of the elements change, but I'll perhaps tackle that at another point.

### 🎫 Issues

Closes #639 

## 👩‍💻 Reviewer Notes

This seems like it'd be quite easy to create a test for. Can someone point me in the right direction where to add such a test?

## ✅ DoD

### DIFF
--- a/packages/components/src/components/prio-nav/prio-nav.component.ts
+++ b/packages/components/src/components/prio-nav/prio-nav.component.ts
@@ -122,12 +122,14 @@ export default class SynPrioNav extends SynergyElement {
    * @param items The items to cache the position for
    */
   private cacheItemPositions(items: SynNavItem[]) {
+    const { left } = this.horizontalNav.getBoundingClientRect();
     items.forEach(item => {
       // We have to measure while the items are in the primary slot,
       // else we will just get the placement in the priority menu
       item.removeAttribute('slot');
       const { right } = item.getBoundingClientRect();
-      item.dataset.right = right.toString();
+      // We also need to consider the location of the parent
+      item.dataset.right = (right - left).toString();
     });
 
     this.itemPositionsCached = true;
--- a/packages/components/src/components/prio-nav/prio-nav.test.ts
+++ b/packages/components/src/components/prio-nav/prio-nav.test.ts
@@ -72,17 +72,19 @@ describe('<syn-prio-nav>', () => {
      * Create a fixture for our horizontal nav
      * Note we are forcing a fixed width of the items
      * to make sure they are available in all test environments
+     * Without flex-shrink: 0, the items will shrink to fit the container,
+     * rendering most of the tests pointless
      * @param width The width to use
      * @returns A horizontal nav fixture
      */
-    const createFixture = async (width = 800) => fixture<SynPrioNav>(html`
-      <syn-prio-nav style="width: ${width}px;">
-        <syn-nav-item horizontal style="width: 100px">Item 1</syn-nav-item>
-        <syn-nav-item horizontal style="width: 100px">Item 2</syn-nav-item>
-        <syn-nav-item horizontal style="width: 100px">Item 3</syn-nav-item>
-        <syn-nav-item horizontal style="width: 100px">Item 4</syn-nav-item>
-        <syn-nav-item horizontal style="width: 100px" id="button-item" role="button">Item 5</syn-nav-item>
-        <syn-nav-item horizontal style="width: 100px">Item 6</syn-nav-item>
+    const createFixture = async (width = 800, marginLeft = 0) => fixture<SynPrioNav>(html`
+      <syn-prio-nav style="width: ${width}px; margin-left: ${marginLeft}px;">
+        <syn-nav-item horizontal style="width: 100px; flex-shrink: 0;">Item 1</syn-nav-item>
+        <syn-nav-item horizontal style="width: 100px; flex-shrink: 0;">Item 2</syn-nav-item>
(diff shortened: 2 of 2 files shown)


schilchSICKAG:
@xkabylgSICKAG: Thank you for your Input, we will have a look at it. Do you mind sharing an example what is broken in the first place (e.g. via a Code-Pen. and maybe creating an issue for this? We are not notified per default when a new PR is created, so it slipped through initially).

xkabylgSICKAG:
@schilchSICKAG sure, I'll create a codepen. I was thinking I'd create a test for this anyway, and starting there would probably be easy to convert to a test later.

schilchSICKAG:
> @schilchSICKAG sure, I'll create a codepen. I was thinking I'd create a test for this anyway, and starting there would probably be easy to convert to a test later.

Thanks a lot, really appreciated :)

xkabylgSICKAG:
@schilchSICKAG 

https://codepen.io/xkabylgSICKAG/pen/YzoMjaQ

You can see quite clearly here that if you set the width to 500px there is enough room for all the nav-items but on the nav that has margin it's still collapsed.

Also, even for the lower nav without the margin, when you set it to 450px there's still clearly enough room to show all the items but it still collapses one. Both of these should be fixed with this PR.

Is there a way I can create a codepen using my updated code to show that it fixes the issue?

xkabylgSICKAG:
I found this in dStudio, when adding the prio-nav to this sub-part of the page it collapses _very_ early.
image

xkabylgSICKAG:
I added two tests that fail before this fix, and that pass after this fix.

xkabylgSICKAG:
Hi, what's currently blocking this PR, and is there anything I can do to help move this forward?

schilchSICKAG:
Hey @xkabylgSICKAG,

just had a look at your PR. There are some other open issues (#582, #631, #410, #409) that I also want to tackle that are all part of the navigation and navigation items. As they are all somehow related, I will create an umbrella branch to fix them all at once and also include your bugfix (which now has its own issue at #639).

xkabylgSICKAG:
@schilchSICKAG Any possibility that we could merge this PR earlier than the other things get fixed? We're currently blocked from releasing a new version of dStudio until this issue gets fixed, since it majorly affects a part of our application.

schilchSICKAG:
> @schilchSICKAG Any possibility that we could merge this PR earlier than the other things get fixed? We're currently blocked from releasing a new version of dStudio until this issue gets fixed, since it majorly affects a part of our application.

Hey @xkabylgSICKAG,

in this case I will have a look at it tomorrow. Maybe we can also add a reduced version of the bigger PR (e.g. just fixing the use of `getBoundingClientRect`?) as a quick fix and adding the rest of the features after this has passed?

xkabylgSICKAG:
@schilchSICKAG Yeah, it'd be nice to fix the other sizing issue too, but just fixing the issue with setting a margin would be good enough so we can ship usage of this component.

Want me to split this into two PRs? The first one to merge now would be the test changes minus the `should not hide items if there's enough space` test and the update to the `cacheItemPositions()` method. And the rest of the changes would go in the second PR.

schilchSICKAG:
> @schilchSICKAG Yeah, it'd be nice to fix the other sizing issue too, but just fixing the issue with setting a margin would be good enough so we can ship usage of this component.
> 
> Want me to split this into two PRs? The first one to merge now would be the test changes minus the `should not hide items if there's enough space` test and the update to the `cacheItemPositions()` method. And the rest of the changes would go in the second PR.

Hi @xkabylgSICKAG, thats actually a great offer, thank you. Yees, would be happy to add this as a quick fix today if you split the PRs ❤️

schilchSICKAG:
The other tickets are now tackled at https://github.com/synergy-design-system/synergy-design-system/issues/644 under an umbrella ticket and branch.

xkabylgSICKAG:
@schilchSICKAG I rebased it on the `main` branch and only included the parts that consider the parent when calculating the position of the elements.

schilchSICKAG (approved):
Looks good to me! Thank you for putting your time into this, this is really appreciated!

Issue #553, opened 2024-07-15
fix: 🤔 Vue: Using registerIconLibrary with wrappers lead to bigger bundle size


schilchSICKAG:
# Checklist (for the issue creator)

- [x] I have verified the bug exists in the latest version of `@synergy-design-system/components` (e.g. via `pnpm install @synergy-design-system/components`).
- [x] I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...).
- [x] I would be willing to provide a PR for this issue.

# Version

The bug occurred with the following version of the component library: 2.4.1

# Component

The issue arises when using the following component(s): Icons

# Environment

The issue occurred with the following operating system(s): Node (LTS), Vite (latest)
The issue occurred in the following browser(s): All

# Description

When using `registerIconLibrary` to use a different icon-set, it leads to:
-  warnings when using lit setups. This is caused by the fact that Wrappers (e.g. for Vue) will load their own version of the components package when importing.
- treeshaking not possible
- components and lit is packaged 2 times

Simple Example for a Vue.js APP:

```ts
// main.tsx, loads vue app
// Add this line to enable the light theme for your application
import "@synergy-design-system/tokens/themes/light.css";
import "@synergy-design-system/components/index.css";

// Optional: Import the styles package
import "@synergy-design-system/styles/index.css";

import "./assets/main.css";

import { createApp } from "vue";

// 👇 Leads to the multiple version loaded warning
import { registerIconLibrary } from "@synergy-design-system/components";

// 👇 Does not work at all anymore as its its own scope,
// but gets rid of the error message :(
import { registerIconLibrary } from '@synergy-design-system/components/utilities/icon-library.js';

// @todo: DOCUMENT THAT THIS IMPORT HAS TO HAPPEN AFTER ICON LIB!
import App from "./App.vue";

registerIconLibrary("material-symbols", {
  resolver: (name) => {
    return `/material-symbols/${name}.svg`;
  },
  mutator: (svg) => {
    return svg.setAttribute("fill", "currentColor");
  },
});

createApp(App).mount("#app");
```

# Acceptance criteria

- [x] All DoD items successfully completed
- [x] Make sure to provide registerIconLibrary in a way that does not load lit twice (e.g. reexport from the wrapper package?).
- [x] Documentation is adjusted and point out that order matters (known issues)

schilchSICKAG:
[...]

PR #611, opened 2024-09-16
fix: 🤔 Build: Use prettier for formatting generated framework wrappers

### LINKED ISSUE #593: chore: 🔧 Build: Use prettier for formatting generated framework wrappers
# Description

A complete component build takes quite some time. This is mostly caused by eslint, which is used to apply formatting on the created wrappers. Switching to prettier could result in ca. 50% faster build times (for full component builds) and much easier configuration, as we could add the needed dependencies and formatting rules directly to the corresponding projects instead of bloating the build code with [custom](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/scripts/jobs/angular/format.js) [eslint](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/scripts/jobs/react/format.js) [configurations](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/scripts/jobs/vue/format.js).

# Rationale

Faster builds mean faster update cycles during debugging and ci builds. It also lessens the dependence on eslint as we are not linting the created wrappers at all.

# Acceptance criteria

- [ ] A `prettierrc` is provided for all wrappers, providing framework options and other needed settings.
- [ ] Current formatting jobs via eslint are removed in favor of an exec of prettier

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adjusts our builds to use prettier instead of eslint for prettifying the created wrapper code. This fastens up the build step significantly and removes some now unneeded dependencies for eslints vue configuration.

As can be seen at the following screenshot, times for builds have lowered from `32sec` to `6sec`, making it ~ 5 faster than before.

Bildschirmfoto 2024-09-16 um 08 36 24

### 🎫 Issues

Closes #593 

## 👩‍💻 Reviewer Notes

This has to be a fix, because it adjusts the output of the generated wrappers. Functionality wise, it may have been a `chore`, but as the code itself has changed, I think we should opt for a fix here.

## 📑 Test Plan

- Check out this branch (absolutely needed this time!)
- Start via a clean state via `pnpm clean`
- See how fast the build is now :)

## ✅ DoD


packages/angular/src/components/alert.component.ts
```diff
@@ -6,14 +6,16 @@
...
+  EventEmitter,
 } from '@angular/core';
-import type {
-  SynAfterHideEvent, SynAfterShowEvent, SynAlert, SynHideEvent, SynShowEvent,
-} from '@synergy-design-system/components';
+import type { SynAlert } from '@synergy-design-system/components';
+import type { SynShowEvent } from '@synergy-design-system/components';
+import type { SynAfterShowEvent } from '@synergy-design-system/components';
+import type { SynHideEvent } from '@synergy-design-system/components';
+import type { SynAfterHideEvent } from '@synergy-design-system/components';
```
kirchsuSICKAG, line 18: Could we get a problem with importing everything for its own instead of doing this: 

```js
import type {
  SynAfterHideEvent, SynAfterShowEvent, SynAlert, SynHideEvent, SynShowEvent,
} from '@synergy-design-system/components';
```

If don't think so, right?

kirchsuSICKAG:
Really nice! it's so much faster now!

packages/angular/src/components/alert.component.ts
schilchSICKAG: No, there should not be a problem at all with this, even if I am unsure why it does that. However, as can be seen in [the react wrappers](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/react/src/components/alert.ts), they always behaved that way.

While having a look at #553, I also thought this would crash. However, the types will always be stripped via typescript, so it will not effect the bundle size in either way.

kirchsuSICKAG (approved):

Issue #583, opened 2024-08-14
feat: ✨ drag and drop of directories on syn-file


kirchsuSICKAG:
# Checklist (for the issue creator)

- [x] I have looked at the documentation site and was not able to find the required feature.
- [x] I have looked into the Figma boards and was not able to find the required feature.
- [ ] I would be willing to provide a PR for this feature.

# Story

As a user of the Synergy Design System, I would like to be able to drag and drop not only files but also directories on the syn-file.

# Acceptance criteria

- [x] All DoD items successfully completed
- [x] Drag and drop of directories are possible
- [x] Include all files recursively (flat list)
- [x] Change text from "file" to "folder"
- [x] Create a folder story
- [x] Test in all browsers
- [x] Check how mime type handling is done (document if necessary)

PR #616, opened 2024-09-17
feat: ✨ add support for folder dnd in syn-file

klaarseSICKAG:
- enable folder dropping
- change button and drop-area text according to webkitdirectory
- added story for folder drop

# Pull Request

## 📖 Description

Adds dnd support for folders

### 🎫 Issues

#583 

## 👩‍💻 Reviewer Notes

Have a look at the new "Directory" story

## 📑 Test Plan

1. set webkitdirectory
2. set accept
3. drag a folder with accept-files and not accept files onto the drop area
4. the number of files should match the number of proper accept files

## ✅ DoD


packages/components/src/components/file/file.component.ts
```diff
@@ -300,6 +302,54 @@ export default class SynFile extends SynergyElement implements SynergyFormContro
...
+    function arrayToFileList(f: File[]): FileList {
+      const dataTransfer = new DataTransfer();
+      f.forEach(file => dataTransfer.items.add(file));
+      return dataTransfer.files;
+    }
+
+    const acceptArray = acceptStringToArray(this.accept);
+
+    // Validate the files against the accept attribute
+    return arrayToFileList(Array.from(files).filter(f => fileHasValidAcceptType(f, acceptArray)));
```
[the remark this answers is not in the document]
klaarseSICKAG: then you're fine with handling file-filtering being different in the drop and in the select scenario?
why not setting the accept to an empty string for that scenario?
fyi, if you use the default html input element for dropping, the accept attribute also leads to a filtered list ...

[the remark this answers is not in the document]
xkabylgSICKAG: I'm pretty sure that accept=".txt" still allows me to select a png with a regular input element.

I can very easily change the filtering in Windows from only .txt files to all files:
image

And dropping 3 pngs on a `<input accept=".txt" multiple>` doesn't seem to do any filtering either:
image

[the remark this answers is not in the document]
klaarseSICKAG: good for pointing that out. on a mac this is not possible changing to "all files"

reichst1SICKAG (commented):
I would remove the Help texts in the directory story, as we only have the relevant story-feature active.

Also the Value is "No files chosen" instead of "No folder" chosen.

Please use the Docs-Token, I've created them just now. Please check also the content ;)

I also reorderd the stories to be more in line with the order we have used in the other components, see Figma branch: https://www.figma.com/design/bZFqk9urD3NlghGUKrkKCR/branch/nQAVfXO0PU9UTogUm0YEkh/Synergy-Digital-Design-System?node-id=21709-49135&node-type=frame&t=lnrFPLDbIsc4PlfH-11

packages/components/src/components/file/file.component.ts
```diff
@@ -331,10 +382,20 @@ export default class SynFile extends SynergyElement implements SynergyFormContro
...
+
+    let { files } = e.dataTransfer;
+
+    if (this.webkitdirectory) {
+      // in case of a folder drop, overwrite the files with the folder content
+      const items = e.dataTransfer?.items;
+      files = await this.handleTransferItems(items);
+    }
 
     if (files) {
```
schilchSICKAG, line 397: You could write this with lower indentation levels by using early returns, e.g. something like this:


```suggestion
this.userIsDragging = false;

    if (!files) {
      return;
    }

    if (!this.multiple && !this.webkitdirectory && files.length > 1) {
      this.emit('syn-error');
      return;
    }
    // Use the transferred file list from the drag drop interface
    const hasTrigger = this.hasSlotController.test('trigger');
    if (!hasTrigger) {
      const disappearAnimation = getAnimation(this.inputChosen, 'file.text.disappear', { dir: this.localize.dir() });
      const appearAnimation = getAnimation(this.inputChosen, 'file.text.appear', { dir: this.localize.dir() });

      if (this.droparea) {
        const dropIconAnimation = getAnimation(this.dropareaIcon, 'file.iconDrop', { dir: this.localize.dir() });
        // eslint-disable-next-line @typescript-eslint/no-floating-promises
        animateTo(this.dropareaIcon, dropIconAnimation.keyframes, dropIconAnimation.options);
      }
      // eslint-disable-next-line max-len
      await animateTo(this.inputChosen, disappearAnimation.keyframes, disappearAnimation.options);
      this.handleFiles(files);
      await animateTo(this.inputChosen, appearAnimation.keyframes, appearAnimation.options);
    } else {
      this.handleFiles(files);
    }

    this.input.dispatchEvent(new Event('change'));
```

packages/components/src/components/file/file.component.ts
```diff
@@ -331,10 +382,20 @@ export default class SynFile extends SynergyElement implements SynergyFormContro
     e.preventDefault();
     e.stopPropagation();
 
-    const files = e.dataTransfer?.files;
+    if (!e.dataTransfer) {
+      return;
+    }
+
+    let { files } = e.dataTransfer;
```
schilchSICKAG, line 389: This potentially overrides the original dataTransfer. Maybe something like this would be possible, too?


```suggestion
    const files = this.webkitdirectory
      ? await this.handleTransferItems(e.dataTransfer?.items)
      : e.dataTransfer.files;
```

packages/components/src/components/file/file.component.ts
```diff
@@ -300,6 +301,56 @@ export default class SynFile extends SynergyElement implements SynergyFormContro
     this.files = files;
   }
 
+  private async handleTransferItems(items: DataTransferItemList | null): Promise<FileList> {
+    if (!items || items.length !== 1) {
+      this.value = '';
+      return new Promise((_resolve, reject) => { reject(new Error('No proper items found')); });
+    }
+
+    const itemsArray = Array.from(items);
```
schilchSICKAG, line 310: This is more taste, but I would combine those statements (and also get rid of the forced type as TS is able to guess it pretty well):


```suggestion
    const entries = Array
      .from(items)
      .map(item => item.webkitGetAsEntry());
```

packages/components/src/components/file/file.component.ts
```diff
@@ -300,6 +301,56 @@ export default class SynFile extends SynergyElement implements SynergyFormContro
...
+
+    const entries: (FileSystemEntry | null)[] = itemsArray.map((item) => item.webkitGetAsEntry());
+
+    if (!entries[0]?.isDirectory) {
```
schilchSICKAG, line 314: Because I don´t have a clue: Can we be sure that a directory entry is always position zero? What if a user drags three images and a directory at the same time? Don´t know if there are predefined settings for this.

packages/components/src/components/file/file.component.ts
```diff
@@ -300,6 +301,56 @@ export default class SynFile extends SynergyElement implements SynergyFormContro
...
+      this.value = '';
+      return new Promise((_resolve, reject) => { reject(new Error('Not a folder')); });
```
schilchSICKAG, line 316: What are we doing with the rejection here? We do not throw `syn-invalid` or another event to the outside as far as I can see. Also, we don´t check for rejection in the calling method.

packages/components/src/components/file/file.component.ts
```diff
@@ -300,6 +301,56 @@ export default class SynFile extends SynergyElement implements SynergyFormContro
...
+    }
+
+    const filesPromises = entries.map(entry => this.getFilesFromEntry(entry));
+    const filesArray = await Promise.all(filesPromises);
+    const files = filesArray.flat();
+
+    function arrayToFileList(f: File[]): FileList {
```
schilchSICKAG, line 323: Inlining this would prevent the creation of a needless closure and get rid of unneeded indentation:


```suggestion
    const dataTransfer = new DataTransfer();
    Array.from(files).forEach(f => dataTransfer.items.add(f));
    return dataTransfer.files;
```

You could also return the dataTransfer object itself, which would make it possible to keep the descructuring in handleDrop like this:

```javascript
  private async handleDrop(e: DragEvent) {
    e.preventDefault();
    e.stopPropagation();

    if (!e.dataTransfer) {
      return;
    }

    const { files } = this.webkitdirectory
      ? await this.handleTransferItems(items)
      : e.dataTransfer;

```

schilchSICKAG:
Just had a look at this, seems to work pretty well. However, I noted that with the following tests:

> set webkitdirectory
> set accept
> drag a folder with accept-files and not accept files onto the drop area
> the number of files should match the number of proper accept files

I do get a valid input field when setting `accept` to `*.png` and uploading a folder full of `woff` files. Currently, it adds all the files in the folder to the `files` property of the `<syn-file>`. This was intended as far as I know. However, there is no user feedback about errors (e.g. if there where files selected that are not available in the `accept` property. Maybe we should at least emit a `syn-invalid` or something like this? As far as I can see, I did not do this for the original implementation, too, so we might have to discuss this in another issue.

schilchSICKAG (approved):
Nice work! :)

reichst1SICKAG (approved):

PR #629, opened 2024-09-20
docs: 📚 add footer templates

### LINKED ISSUE #501: feat: ✨ footer template

# Story

As an application or website developer, I want to add a footer that includes all the required legal information. Since every website and application is legally required to include this, the footer should look the same (or similar) on all SICK websites. Otherwise, everyone will make their own footer that behaves and looks differently.

Example footers:
- https://www.sick.com/de/en/
- https://id.sick.com
- https://cloud.sick.com/

What the components needs to do:
- Be displayed at the bottom of the page
- Include links to: 
  - [Imprint](https://www.sick.com/de/en/w/imprint)
  - [Terms and conditions](https://www.sick.com/de/en/w/tac)
  - [Terms of Use](https://www.sick.com/de/en/w/terms-of-use)
  - [Privacy Policy](https://www.sick.com/de/en/w/dataprotection)
- Include a copyright notice
- Provide translations for our 18 or so supported languages (or be translatable)
- Reactive layout that adapts to different resolutions (e.g. sick.com changes the position of the copyright notice depending on the size of the viewport)

image

Alternative design: Provide a way to add links to the footer from the application level. Then, synergy does not need to provide the translations.

Application context: Every website or application that is customer facing needs this at the bottom of the page for legal reasons.

# Acceptance criteria

- [ ] A footer template is included into the synergy library (AppShell)
- [ ] Add to FW demos
- [ ] Dog footing of link styling
- [ ] Optionally add footer style into styles-package (background color, paddings,...)
- [ ] add responsiveness

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR provides template for the footer.

### 🎫 Issues
Closes #501.

## 👩‍💻 Reviewer Notes
- [x] footer desktop template added (with responsiveness)
- [x] footer mobile template added
- [x] appshell footer template added
- [x] available appshell templates updated
- [x] appshell and footer template share components
- [x] some small bugs where fixed for the appshell templates in the docs page

## 📑 Test Plan
- have a look at the footer stories and test it also via changing width

## ✅ DoD


reichst1SICKAG (approved):
Design is done

packages/docs/src/shared-components/appshell.ts
schilchSICKAG, line 1: I think it is quite nice that this is now shared between the stories :)

packages/docs/src/shared-components/appshell.ts
```diff
@@ -0,0 +1,350 @@
...
+
+      .footer-content {
+        align-items: flex-start;
+        flex-direction: column;
+        padding: var(--syn-spacing-large) var(--syn-spacing-medium);
+      }
+    }
+  </style>
+  <footer class="synergy-footer-demo">
+    <div class="footer-content">
```
schilchSICKAG, line 52: From what I see from the applied styles, we may get away with the following markup:

```html
<footer class="synergy-footer-demo">
  <nav class="syn-footer-navigation-items">
    <a class="syn-link syn-link--medium syn-link-quiet" href="HREF">LINK/a>
    <a class="syn-link syn-link--medium syn-link-quiet" href="HREF">LINK/a>
    <a class="syn-link syn-link--medium syn-link-quiet" href="HREF">LINK/a>
    <a class="syn-link syn-link--medium syn-link-quiet" href="HREF">LINK/a>
  </nav>
  <p class="copyright">&copy; 2024 SICK AG</p>
</footer>
```

This would remove some unneeded markup from this function and also ease the css, as intermediate styling for the now unneeded li could be dropped in favor of something like:

```css
.syn-footer-navigation-items {
  ... Stuff from link-wrapper here, container stuff, too
}
```

What do you think?

schilchSICKAG (changes requested):
Looks good to me. I just have some questions about the used markup (maybe we can remove some nesting from there, see comment)

packages/docs/src/shared-components/appshell.ts
kirchsuSICKAG: I didn't had the `<div class="footer-content"/>` in my previous version without the container query. But with container query we unfortunately need it..  😕 
The element, where the "container-type" is added, cannot be targeted for container queries
If we have the `<footer />` as container, we can't use it for changes like following:

```
@container (max-width: 768px) {
  footer {
      // style different if smaller
  }
}
```

Whether we leave out the `<ul />` and `<li />` we can discuss about it. You are right, that it is much cleaner without it. 
I did it, because I had a look at the `footer` components from others and they often wrap the `<a />`s into `<ul />`  and `<li/>`
( e.g. https://designsystem.digital.gov/components/footer/, https://edisondesignsystem.com/footer?tab=2)

What do you mean? Could it have any accessibility advantages?

reichst1SICKAG: We've had the idea to add a link list component (Figma) and CSS styles for this <ul> <li> thing.
Maybe we refactor the footer link list later with it?

packages/docs/src/shared-components/appshell.ts
```diff
@@ -0,0 +1,340 @@
...
+      height: var(--syn-spacing-3x-large);
+      justify-content: center;
+    }
+
+    .synergy-demo-footer :is(a, strong) {
+      color: var(--syn-input-help-text-color);
+    }
+
+    .synergy-demo-footer a:hover {
+      color: var(--syn-color-primary-600);
```
schilchSICKAG, line 145: As this may be removed by a reset somewhere else: Do we want to have a `text-decoration: underline;` here, too? For accessability reasons, I think it would be mandatory to not only work with color during hover.

schilchSICKAG (changes requested):
Just one more thing :).

packages/docs/src/shared-components/appshell.ts
kirchsuSICKAG: just saw, that this is not needed anyways, because I created the shared footer style and this is therefore no longer used

schilchSICKAG: As this is using our `syn-link`, it may not be needed at all :)

schilchSICKAG (approved):

PR #655, opened 2024-10-18
fix: 🐛 Incorrect validations in angular forms

### LINKED ISSUE #596: fix: 🐛 Mark form as invalid when invalid (out of bounds) numeric value is provided

# Version

The bug occurred with the following version of the component library: `2.5.0`

# Component

The issue arises when using the following component(s): `syn-input`

# Environment

The issue occurred with the following operating system(s): `Windows 10`
The issue occurred in the following browser(s): `Google Chrome`

# Description

When using the `syn-input` component, with assigned `"number"` type, and also setting `[min]`, `[max]`, `[pattern]` attributes, the form that wraps this input does not get marked as "invalid".

**Context**: We are using Angular 18 so, according to that, we are using `ngForm`'s `ng-valid`/`ng-invalid` (or, in our script `form.valid` property) to control whether a (Next/Apply/Finish) button would be enabled or disabled.

Here is a snippet of how our code, that should achieve that, looks like:

HTML template:

``` html
<form #form="ngForm" [appForm]="form">
  <syn-input 
    type="number"
    [min]="0"
    [(ngModel)]="fenceTimeout"
    name="fenceTimeout">
  </syn-input>
</form>
```

Our custom directive which purpose is to just report (among some other stuff) on validity of the underlying form:

``` ts
@Directive({
  selector: '[appForm]',
  exportAs: 'form',
})
export class FormDirective implements AfterViewInit {
  appForm = input.required<NgForm>();
  isValid$ = new BehaviorSubject<boolean>(false);

  ngAfterViewInit(): void {
    this.isValid$.next(this.appForm().form.valid);
  }

  @HostListener('input')
  onInput(): void {
    this.isValid$.next(this.appForm().form.valid);
  }
}
```

# Acceptance criteria

- [ ] Having invalid values provided to numeric `syn-input`s with `[min]`, `[max]`, ... attributes in the form should be picked up by that form and mark it as invalid

kirchsuSICKAG:
# Pull Request

## 📖 Description
Some form validations are not working correctly in angular:
 - `<syn-input type="number" />` with `min` attribute
 - `<syn-input type="number" />` with `max` attribute
 - `<syn-checkbox />` with `required` attribute

Angular automatically adds custom css classes to all form input elements and to the `<form />` element itself. For more information about this have a look at [the details of the angular css classes](https://v17.angular.io/guide/form-validation#control-status-css-classes).
The classes are added after the [angular built-in validator functions](https://v17.angular.io/api/forms/Validators) (like a required validator) are executed for the input element.

Some of the validator functions are working out of the box with our synergy input elements, because they are added automatically. This is because the selector, which decides if a validator is used, suits for them.

A working example is the required validator. It is used correctly for our components because the selector looks like following and therefore takes action:
```
 ':not([type=checkbox])[required][formControlName],:not([type=checkbox])[required][formControl],:not([type=checkbox])[required][ngModel]',
```

This is not the case for the ["min"](https://github.com/angular/angular/blob/e603e5616d2c3830a641d4cf76c891af0059798c/packages/forms/src/directives/validators.ts#L276), ["max"](https://github.com/angular/angular/blob/e603e5616d2c3830a641d4cf76c891af0059798c/packages/forms/src/directives/validators.ts#L224) and ["checkbox required"](https://github.com/angular/angular/blob/e603e5616d2c3830a641d4cf76c891af0059798c/packages/forms/src/directives/validators.ts#L429) validator. They are specific for an `<input />` element and therefore do not activate for our components:
```
 'input[type=number][min][formControlName],input[type=number][min][formControl],input[type=number][min][ngModel]',
```

In this PR I added custom validators for our components for the "min", "max" and "checkbox required" validations

### 🎫 Issues
Closes #596 

## 👩‍💻 Reviewer Notes

- I tried adding tests for the form validation in our e2e test, but unfortunately we would need to add angular specific tests and therefore we would need to check for angular classes in our general tests... 

## ✅ DoD


packages/components/scripts/jobs/angular/createValidatorDirectives.js
```diff
@@ -0,0 +1,123 @@
+import fs from 'fs/promises';
+import path from 'path';
+import { createHeader, job } from '../shared.js';
+
+const headerComment = createHeader('angular');
+
+/**
+ * Creates the synergy angular module, located at packages/angular/src/modules/synergy.module.ts
+ */
+export const runCreateFValidatorDirectives = job('Angular: Creating ValidatorDirectives...', async (outDir) => {
```
schilchSICKAG, line 10: There is a small typo in here, `runCreateFValidatorDirectives` should be named `runCreateValidatorDirectives` (without "F")

packages/components/scripts/jobs/angular.js
```diff
@@ -28,13 +28,16 @@ export const runCreateAngularWrappers = async ({
...
+
   const distDir = path.join(angularPackageDir, './dist');
 
   await runAdjustPackageVersion('Angular: Adjusting angular package.json version field...')(componentPackageDir, angularPackageDir);
-  await createRunPrepare('Angular: Cleaning up artifacts...')(outDir, distDir, modulesDir, componentsDir);
+  await createRunPrepare('Angular: Cleaning up artifacts...')(outDir, distDir, modulesDir, componentsDir, directivesDir);
   await jobs.runCreateComponents(metadata, componentsDir);
   await jobs.runCreateNgModule(metadata, modulesDir);
   await jobs.runCreateFormsModule(modulesDir);
+  await jobs.runCreateFValidatorDirectives(directivesDir);
```
schilchSICKAG, line 40: should be `await jobs.runCreateValidatorDirectives(directivesDir);`;

schilchSICKAG (changes requested):
Very nice work! Just a minor spelling thing in the internal stuff, otherwise I am fine with this :)

schilchSICKAG (approved):
Thanks :)

PR #656, opened 2024-10-21
fix: 🐛 Incorrect scaling of syn-icon with font-size if using spritesheet

### LINKED ISSUE #649: fix: 🐛 Icons do not scale with font-size when using `spriteSheet: true` for custom icon library

# Version

The bug occurred with the following version of the component library: 2.10.1

# Component

The issue arises when using the following component(s): syn-icon

# Environment

The issue occurred in the following browser(s): Chromium, Safari (Firefox not tested)

# Description

When using SVG sprite sheets as custom icon library the icons size does not scale with the font size of the `syn-icon` component. 
I use `svg2sprite-cli` to generate the sprite sheet from a given set of individual SVG icons. The library is registred using 
``` ts
registerIconLibrary('sprite', {
  resolver: (name) => `./assets/sprites/sprite.svg#${name}`,
  mutator: (svg) => {
    svg.setAttribute('fill', 'currentColor');
  },
  spriteSheet: true,
});
```

When adding `height=100%` and `width=100%` attributes to the `<use>` tag in the `syn-icon`'s shadow dom the icon scales correctly. Maybe these should be added to [icon.component.ts](https://github.com/synergy-design-system/synergy-design-system/blob/a63815aba7d158199b786ad7770782492ea403f8/packages/components/src/components/icon/icon.component.ts#L68)? 

Thanks in advance. 

P.S.: I also noticed that the code snippet in the [documentation on 'Sprite Sheet Usage'](https://github.com/synergy-design-system/synergy-design-system/blob/a63815aba7d158199b786ad7770782492ea403f8/packages/components/src/components/icon/icon.component.ts#L68) seems to be wrong (the first `resolver: (name) => {` must be deleted). 

# Acceptance criteria

kirchsuSICKAG:
# Pull Request

## 📖 Description

If having a sprite sheet with following two svg's
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <!-- Not scaling svg -->
    <symbol id="settings" width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" fill='currentColor'>
        <path .../>
    </symbol>
    <!-- Scaling svg -->
    <symbol id="refresh" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" fill='currentColor'>
        <path .../>
    </symbol>
</svg>
```
The first svg (id="settings") is not scaling with the font-size while the second svg (id="refresh") is scaling.
This is because id="settings" has a fix `width` and `height` set on the svg and therefore does not scale with the parent font-size.

There are 2 possible ways to handle this problem:

1. Add to the documentation, that if using sprite sheets, the `width` and `height` properties need to be removed from the svg's. 
   The problems with this solution are:

   - the syn-icon with sprite sheet does not work out of the box and does need changes from our users
   - the users may use an external tool to create the sprite sheet and can't influence how the sprite sheet is created

2. add `width="100%"` and `height="100%"` on the `<use />` element

   - it seems that everything is working like before
   - even if the width and height on the svg's are set, they are scaled with font-size (this is because the width and height of the svg is overwritten with the 100%)
- 

### 🎫 Issues
Closes #649 

## 👩‍💻 Reviewer Notes
We need to decide, if we want to go with proposal 1 or 2.
The current state of this PR goes with proposal 2, so the user does not need to do something on his side.
Additionally I added the shoelace tests for icons with sprite sheets.

## ✅ DoD


packages/components/public/sprite.svg
schilchSICKAG, line 1: Are these manually generated for the demos? Just want to make sure we do not have the original value when someone builds the icons again...

packages/components/src/components/icon/icon.custom.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      await elementUpdated(icon);
+
+      const svg = icon.shadowRoot!.querySelector("svg[part='svg']")!;
+      const use = svg.querySelector("use[href='/public/sprite.svg#refresh']")!;
+
+      expect(svg).to.be.instanceof(SVGElement);
+      expect(use).to.be.instanceof(SVGUseElement);
+
+      // This is kind of hacky...but with no way to check "load", we just use a timeout
+      await aTimeout(1000);
```
schilchSICKAG, line 32: Don´t know why this was there at all, the test seems to run fine without this line (at least on my local system). Does not matter that much, just wanted to point it out.

packages/components/src/components/icon/icon.custom.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      await elementUpdated(icon);
+
+      const svg = icon.shadowRoot!.querySelector("svg[part='svg']")!;
+      const use = svg.querySelector("use[href='/public/sprite.svg#refreshWithoutWidth']")!;
+
+      expect(svg).to.be.instanceof(SVGElement);
+      expect(use).to.be.instanceof(SVGUseElement);
+
+      // This is kind of hacky...but with no way to check "load", we just use a timeout
+      await aTimeout(1000);
```
schilchSICKAG, line 73: Don´t know why this was there at all, the test seems to run fine without this line (at least on my local system). Does not matter that much, just wanted to point it out.

schilchSICKAG (approved):
Seems to work fine 👍. Just tried it locally using Chrome and Safari and everything works as intended.

schilchSICKAG:
I think going with approach 2 is fine for most of the users as they may not be able to control the source of the svg. You could technically also do the following (I think ^^):

```javascript
      registerIconLibrary('sprite', {
        resolver(name) {
          return `/public/sprite.svg#${name}`;
        },
        mutator(svg) {
          const use = svg.querySelector('use');
          use.setAttribute('width', '100%');
          use.setAttribute('height', '100%');
          return svg.setAttribute('fill', 'currentColor');
        },
        spriteSheet: true
      });
```

bit it feels like the wrong answer for me, too.

packages/components/public/sprite.svg
kirchsuSICKAG: Yes I manually generated them. Not for the demo, but for the unit tests.
The storybook demo has its own sprite sheet here: `./packages/docs/public/sprite.svg`

packages/components/src/components/icon/icon.custom.test.ts
kirchsuSICKAG: I took it from shoelace.
If not doing it, the test runs two times, because otherwise the first `width` and `height` are `0`.

Running tests with the timeout:

image

Running the tests without the timeout (sometimes it even seems to fail):
image
image

PR #646, opened 2024-10-02
fix(prio-nav): only take size of dropdown button into account for shrinkage if it's shown

xkabylgSICKAG:
# Pull Request

## 📖 Description

Follow up from #608 

## ✅ DoD

- [ ] I have tested my changes manually.
- [ ] I have added automatic tests for my changes (unit- and visual regression tests).
- [ ] I have updated the project documentation to reflect my changes (e.g. CHANGELOG, Storybook, ...).
- [ ] I have added documentation to the DaVinci migration guide (if need be)
- [ ] I have made sure to follow the projects coding and contribution guides.
- [ ] I have made sure that the bundle size has changed appropriately.
- [ ] I have validated that there are no accessibility errors.
- [ ] I have used design tokens instead of fix css values

PR #665, opened 2024-11-04
fix: 🐛 syn-prio-nav takes size of prio menu into account when it's not shown

### LINKED ISSUE #651: fix: 🐛 syn-prio-nav takes size of prio menu into account when it's not shown

# Version

The bug occurred with the following version of the component library: 2.11.0

# Component

The issue arises when using the following component(s): syn-prio-nav

# Environment

# Description

The prio-nav takes the size of the prio menu into account even when that menu is not shown.

I previously created this PR #646 which fixed that issue, but I guess it would need to be updated to handle some of the recent changes in #645

Codepen:

https://codepen.io/xkabylgSICKAG/pen/JjgEGxY

Failing test, can be added into `prio-nav.test.ts`:

```ts
    it.only("should not hide items if there's enough space", async () => {
      const nav = await createFixture();
      // Need to force an update for some reason, just setting it initially won't work.
      // 6 items of 100px = 600px
      // 5 gaps of 24px = 120px
      // 720px should be enough, but give it 1px extra since we're dealing with floats in some parts
      await forceSizeUpdateOnFixture(nav, 721);

      const [itemsInDefaultSlot, itemsInPrioritySlot] = getSlottedChildrenAsTuple(nav);

      expect(itemsInDefaultSlot).to.have.length(6);
      expect(itemsInPrioritySlot).to.be.empty;
    });
```

# Acceptance criteria

- [x] The provided failing test passes

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds a fix that makes sure the size of the prio nav menu is not taken into account if it is not visible at all, saving some 50 pixels in the calculation and therefore making it possible to show another item in edge cases.

### 🎫 Issues

Closes #651 

## 👩‍💻 Reviewer Notes

Changes from the [original implementation](https://github.com/synergy-design-system/synergy-design-system/pull/646/commits/7e5f2ff727501f878e362835cf0059750f77b3df) where taken into account, however the fix seems to be relatively easy to do without resorting to reverse list calculations, so I opted to created a smaller version based on the code that we currently have.

## 📑 Test Plan

Play around with storybook, removing the fixed widths from the samples.

## ✅ DoD


packages/components/src/components/prio-nav/prio-nav.component.ts
kirchsuSICKAG, line 1: We should not do the check for **>=** on line 166 for the `parseFloat(item.dataset.right!) >= finalWidth)` but just do an **>** so it is `parseFloat(item.dataset.right!) >= finalWidth)`
Otherwise it already shrink for a syn-prio-nav width of 720px although there is 1px spacing left:

Bildschirmfoto 2024-11-05 um 14 31 25

```html
<div style="width:720px; background-color: red;">
  Nav set to 720px, it shrinks the last item even though it actually fits.
</div>
<syn-prio-nav style="width: 720px">
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>1</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>2</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>3</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>4</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>5</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>6</syn-nav-item>
</syn-prio-nav>
<div style="width:721px; background-color: red;">
  Nav set to 721px, all items fit
</div>
<syn-prio-nav style="width: 721px">
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>1</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>2</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>3</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>4</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>5</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>6</syn-nav-item>
</syn-prio-nav>

<style>
  syn-prio-nav {
    background-color: yellow;
  }
  syn-nav-item {
  background: rgba(0,0,200,0.5);
}
</style>
```

kirchsuSICKAG, line 1: It was probably already there before, but I just discovered this weird behavior. It seems there is something wrong, when resizing the prio-nav so the last item should already have enough space to show up, but it doesn't (see gif)

prio-nav

```html
<div style="width:calc(600px + calc(var(--syn-spacing-large) * 5)); background-color: red;">
  Nav without any size set, see how all items fit within the 720px width
</div>
 <syn-prio-nav >
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>1</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>2</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>3</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>4</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>5</syn-nav-item>
        <syn-nav-item style="width: 100px; flex-shrink: 0;" horizontal>6</syn-nav-item>
</syn-prio-nav>


<style>
  syn-prio-nav {
    background-color: yellow;
  }
  syn-nav-item {
  background: rgba(0,0,200,0.5);
}
</style>
```

**EDIT**
just found out what the problem is.
we need to adapt line 165-167 to something like this:
```js
 const itemVisibilities = navItems.map((item, index) => {
      const newWidth = index === navItems.length - 1 ? width : finalWidth;
      const isHidden = !!(firstHiddenItemRightPos || parseFloat(item.dataset.right!) > newWidth);
 ```
 
 If it is the last item, we need to take the width of the prio-nav and not the `width minus prioMenuSize` into account

kirchsuSICKAG (changes requested):

packages/components/src/components/prio-nav/prio-nav.component.ts
schilchSICKAG: Fixed in latest version

schilchSICKAG: That was a good catch! Fixed in latest version

kirchsuSICKAG (approved):

PR #682, opened 2024-11-19
feat: ✨ adjust padding for syn-menu + syn-dropdown checkmarks

### LINKED ISSUE #368: feat: ✨ adjust padding for syn-menu + syn-dropdown checkmarks
# Component

I suggest to add the feature to the following component: syn-menu-item

# Story

The current menu has always the same padding on the left side. Current Padding is huge, since the Component can have a Select checkmark. The improvement would be, the menu noticed if the menu has a Select function inside and if no item has a select icon we reduce the padding

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the functionality to hide the checkmark space in `<syn-menu>` when there are no children that either:

- have a `type` of `checkbox` or
- are in the state `loading`

This means that regular `<syn-menu>` elements will not show a large spacing anymore unless there is at least one sub-item that has meets one of the requirements. This leads to a denser basic layout for the menu. Note that this is not working for `<syn-select>` as we show the checkmark as activity indicator!

This is now done via a new global `class decorator` named `emitEventForPropertyUpdates` that adds the possibility to emit a `CustomEvent` named `syn-attributes-changed` when one of the monitored properties change. This is now listened for in `<syn-menu>`, which will do a recalculation of the width via its new `handleUpdateCheckmarks` private property.

### 🎫 Issues

Closes #368 

## 👩‍💻 Reviewer Notes

Just compare the stories for `<syn-menu>`.

I opted to use a `css custom variable` to make this work. However, I do not think that this should be public api, so I did not add any comments to `<syn-menu-item>`.

## 📑 Test Plan

Just have a look at the `<syn-menu>` stories that where changed. You can add a `loading` prop or `type="checkbox"` to any of the items and should see that this adds the original padding back. You can also create a submenu and add or remove an item there to see the effect in submenus.

I also have added a new story (called `Demo`) to the stories of `<syn-menu>`. The story is currently commented out as it is a dev only story for us to demo the new feature. Just comment it in locally and play around with it to see the new feature live.

## ✅ DoD


kirchsuSICKAG (changes requested):

cspell.json
```diff
@@ -28,6 +28,8 @@
     "keyframes",
     "autofocus",
     "listbox"
```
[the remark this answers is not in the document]
schilchSICKAG: Yes, this happened during the merge yesterday. Will have a look at it after finishing the footer :)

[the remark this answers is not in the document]
schilchSICKAG: fixed

packages/components/src/components/menu/menu.component.ts
```diff
@@ -118,6 +121,10 @@ export default class SynMenu extends SynergyElement {
   private handleSlotChange() {
     const items = this.getAllItems();
 
+    // #368: Treat a menu as having checkmarks if it has any checkboxes or items with loading states
+    // The loading indicator has to be checked as well, as it's specially placed over the check mark
+    this.hasMenuItemsWithCheckmarks = items.some(item => item.type === 'checkbox' || item.loading);
```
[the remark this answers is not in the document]
schilchSICKAG: You are right, slotchange will only fire if there are changes to the DOM itself, not adjusting the values. Will have a look into this

packages/components/src/components/menu/menu.component.ts
```diff
@@ -33,10 +36,27 @@ export default class SynMenu extends SynergyElement {
...
+  @state() hasMenuItemsWithCheckmarks = false;
+
+  private handleUpdateCheckmarks(items: SynMenuItem[]) {
+    // #368: Treat a menu as having checkmarks if it has any checkboxes or items with loading states
+    // The loading indicator has to be checked as well, as it's specially placed over the check mark
+    this.hasMenuItemsWithCheckmarks = items.some(item => item.type === 'checkbox' || item.loading);  
+  }
+
+  private updateCheckMarksByChildPropChange = (e: SynAttributesChangedEvent) => {
+    e.stopPropagation();
```
kirchsuSICKAG, line 48: Currently if adding an EventListener on the syn-menu component itself the `syn-attributes-changed` from the menu-item can be catched. 
E.g. following code results in both console.logs are done when changing the syn-menu-item to e.g. loading=true: 

```js
    <syn-menu style="max-width: 240px;">
      <syn-menu-item>
        <syn-icon slot="prefix" name="house"></syn-icon>
        Home
      </syn-menu-item>
    </syn-menu>
    <script type="module">
      const menu = document.querySelector('syn-menu');
      menu.addEventListener('syn-attributes-changed', (event) => {
        console.log('##### syn-menu');
      });
      const menuItem = document.querySelector('syn-menu-item');
      menuItem.addEventListener('syn-attributes-changed', (event) => {
        console.log('##### syn-menu-item');
      });
    </script>
```

If we don't want the event to bubble this far, it seems we need to do a `e.stopImmediatePropagation();`.
What do you mean? Maybe it even doesn't matter, but I was just curious what happens if I listen on the syn-menu for it

packages/components/src/internal/watchEvent.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Class decorator that automatically emits a `syn-attributes-changed` event
+ * when at least one of the provided attributes have changed.
+ *
+ * @example
+ * ```typescript
+ * // Will fire a `syn-attributes-changed` event when `prop1` or `prop2` change,
+ * // but only after the first `updated` call
+ * emitEventForPropertyUpdates(['prop1', 'prop2'], {
```
kirchsuSICKAG, line 22: The `@` is missing in the example

```suggestion
 * @emitEventForPropertyUpdates(['prop1', 'prop2'], {
```

packages/components/src/components/menu/menu.component.ts
schilchSICKAG: No you are right, as this is an implementation detail, it should not process any further.

packages/components/src/components/menu/menu.component.ts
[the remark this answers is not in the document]
schilchSICKAG: This is now fixed with the newer version using attribute change detection

packages/components/src/internal/watchEvent.ts
schilchSICKAG: I adjusted the documentation for this so it works without decorator syntax (which is the default as decorators are not yet standard). :)

packages/components/src/components/menu/menu.component.ts
schilchSICKAG: Fixed in latest version

kirchsuSICKAG (approved):

reichst1SICKAG (approved):
Design is Done

PR #693, opened 2024-11-27
docs: 📚 Add icon search page

### LINKED ISSUE #660: docs: 📚 add searchable icons page
# Description

As a user I want to be able to find a suitable icon for my use case easily. Not only for their names themselves, but also for suitable other tags.

# Rationale

Material already provides tags for each icon. Tags are much better suited for finding icons than categories.

Here a .txt file for the metadata of google material icons can be downloaded, where the tags for each icon can be found: https://fonts.google.com/metadata/icons

# Acceptance criteria

- [ ] doc page is added where icons can be searched via name or other tags

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR adds an additional page to the storybook documentation at the root level, with which icons can be searched more convenient via name and tags.

### 🎫 Issues
Closes #660 

## 👩‍💻 Reviewer Notes
We have the possibility to make the icons searchable via `<syn-input>`  or via `<syn-combobox>`. 
Check out both ways. On default the combobox version is commented out.
Comment it in and compare the performance of `<syn-input>` with `<syn-combobox>` . For me it seems a bit too laggy if using the combobox.

## 📑 Test Plan

Run storybook and have a look at: http://localhost:6006/?path=/docs/icon-search--docs

## ✅ DoD


packages/docs/src/shared-components/IconSearchPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+import './IconSearchPage.css';
+
+type FontIcon = {
+  name: string,
+  categories: Array<string>,
+  tags: Array<string>,
+};
+
+// Make sure ts is not sad about custom elements
+declare global {
```
schilchSICKAG, line 21: Another idea would be to use `@synergy-design-system/react` in here:

```ts
// Before make sure to install:
// pnpm install --save-dev @synergy-design-system/react
import { SynInput } from '@synergy-design-system/react';

/*
Not needed anymore
useEffect(() => {
    registerIcons();

    searchInput.current?.addEventListener('syn-input', () => {
      setSearchTerm(searchInput.current?.value.toLowerCase() || '');
    });
    combobox.current?.addEventListener('syn-input', () => {
      setSearchTerm(combobox.current?.value.toLowerCase() || '');
    });
  }, []);*/
// Not sure this works, but you might give it a try
return (
    <div>
      <div>
        <SynInput label="Search icons" onSynInput={e => setSearchTerm(searchInput.current?.value.toLowerCase() || ''); } />
```

packages/docs/src/shared-components/IconSearchPage.tsx
```diff
@@ -0,0 +1,121 @@
...
+export const IconsSearchPage: FC = () => {
+  const searchInput = useRef<SynInputType>(null);
+  const [searchTerm, setSearchTerm] = useState<string>('');
+  const [categories, setCategories] = useState<Array<string>>([]);
+
+  useEffect(() => {
+    registerIcons();
+  }, []);
+
+  const handleSearchTermChange = () => {
```
schilchSICKAG, line 79: You could get rid of the ref stuff by adjusting this function to:

```typescript
const handleSearchTermChange = (e: SynInputEvent) => {
  const value = (e.target as SynInput).value!.toLowerCase();
  setSearchTerm(value);
}
```

schilchSICKAG (changes requested):
Seems to work fine. You could add a last change to get rid of the `ref` stuff, but that should be it I think. Nice PR ❤️

packages/docs/src/shared-components/IconSearchPage.tsx
```diff
@@ -0,0 +1,140 @@
...
+  FC, MouseEvent, useEffect, useRef, useState,
+} from 'react';
+import '../../../components/src/components/icon/icon.js';
+import '../../../components/src/components/input/input.js';
+import '../../../components/src/components/combobox/combobox.js';
+import { type SynCombobox, type SynInput } from '@synergy-design-system/components';
+import { registerIconLibrary } from '../../../components/src/utilities/icon-library.js';
+import { defaultIcons } from '../../../assets/src/default-icons.js';
+import materialIconsMetatdata from '../materialIconsMetadata.json';
+import './IconSearchPage.css';
```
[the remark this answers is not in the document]
kirchsuSICKAG: The style is no longer imported, as I only added it for the combobox, where we decided we didn't want to use it, as it makes the search page too laggy.

packages/docs/src/shared-components/IconSearchPage.tsx
kirchsuSICKAG: Add the improvement :)

schilchSICKAG (approved):
Very nice, hope we will be able to ship this feature fast 👍

PR #701, opened 2024-12-02
fix:  🐛 Change active tab via active property of syn-tab 

### LINKED ISSUE #604: fix: 🐛 changing active syn-tab / syn-panel via code is not working

# Version

The bug occurred with the following version of the component library: 2.7.0

# Component

The issue arises when using the following component(s): syn-tab, syn-panel, syn-tab-group

# Environment

# Description
Changing the active tab via the "active" property in javascript does not work.
The old active tab and tab panels are not set "active=false" and even if setting the old active tab / panel "active=false" before leads to the old tab having the indicator and even showing both panels (the old active panel and the new active one).
see following codepen: https://codepen.io/kirchsuSICKAG/pen/xxoQXLE

# Acceptance criteria

- [ ] Either shoelace provides the fix
- [ ] Or we need to vendor the fix

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR fixes the problem, that another tab can not be set to active via the active property of the `<syn-tab>` element. 

### 🎫 Issues
Closes #604 

### Open question: 
with this fix it is now possible to set another tab to active via its `active` property. If another tab is set to active via its active property, the `syn-tab-show`  and `syn-tab-hide` events are emitted. I ask myself if this is the correct behavior or if these two events should only be emitted when the user clicks a tab and when the `show()` method is used. 

**My opinion:** 
I think it is correct, that the two events are emitted. Because the events were already emitted when the `show` method was used. Also angular developers will do stuff like this to set the active tab and don't use the `show` method and they will also probably want to do other stuff on the events: 

```html
<syn-tab-group (syn-tab-show)="doStuffOnTabShow($event)">
  <syn-tab slot="nav" [panel]="appSubparts.OVERVIEW" [active]="appSubparts.OVERVIEW == (selectedSubpart$ | async)">Overview</syn-tab>
  <syn-tab-panel [name]="appSubparts.OVERVIEW" >1</syn-tab-panel>
  <syn-tab slot="nav" [panel]="appSubparts.VERSIONS" [active]="appSubparts.VERSIONS == (selectedSubpart$ | async)">VERSIONS</syn-tab>
  <syn-tab-panel [name]="appSubparts.VERSIONS" >2</syn-tab-panel>
</syn-tab-group>
```
 

## 📑 Test Plan

Have a look at the changes.
This is code which can be used to test, if it is working right now,  before it was not.
Also you could just set another tab to active via browser DOM. 
```html
<syn-tab-group>
  <syn-tab-panel name="general" >This is the general tab panel.</syn-tab-panel>
  <syn-tab-panel name="custom">This is the custom tab panel.</syn-tab-panel>
  <syn-tab-panel name="advanced">This is the advanced tab panel.</syn-tab-panel>
  <syn-tab-panel name="disabled" >This is the disabled tab panel.</syn-tab-panel>
  <syn-tab slot="nav" panel="general" >General</syn-tab>
  <syn-tab slot="nav" panel="custom">Custom</syn-tab>
  <syn-tab slot="nav" panel="advanced">Advanced</syn-tab>
  <syn-tab slot="nav" panel="disabled" active>Disabled</syn-tab>
</syn-tab-group>
<script type="module">
  const group = document.querySelector('syn-tab-group');
  group.addEventListener('syn-tab-show', async event => {
    console.log('syntabshow', event.detail);
  });
  group.addEventListener('syn-tab-hide', async event => {
    console.log('syntabshide', event.detail);
  });
  setTimeout(() => {
    const thirdTab = document.querySelector('[panel="advanced"]');
    thirdTab.active = true;
  },2000);
</script>
```

## ✅ DoD


schilchSICKAG (approved):
Seems to work fine 👍

PR #705, opened 2024-12-04
fix:  🐛  select does not respect async value property anymore

### LINKED ISSUE #696: fix: 🐛 syn-select does not honor async value prop

# Version

The bug occurred with the following version of the component library: 2.13.0

# Component

The issue arises when using the following component(s): syn-select

# Environment

# Description

When using async `value` props, the prop is not connected anymore after the update to Shoelace@2.18.0. This seems to stem from the provided fix for lazy loading options (https://github.com/shoelace-style/shoelace/pull/2204)

# Acceptance criteria

- [ ] Adjust the demo projects to include lazy options
- [ ] Write regression tests for this feature as we do not have any for this kind of problem yet
- [ ] Provide a fix for the problem

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR fixes the issue that the select does no longer respect when the value property is set async. 
It is fixed by updating to shoelace 2.19.0 as there the fix for this bug is added.
Additionally a plugin for the custom-elements.json creation needed to be added, as a new handling of shoelace for the "value" and "defaultValue" property where added. With this changes the vue and angular wrapper as well as the property table in storybook where no longer correct.

### 🎫 Issues
Closes #696 

## 📑 Test Plan
- Play around with the syn-select in all framework wrapper so it still works as it should
- Check that the bug is fixed with the changes: 

Example code for Vue:
```
<script setup lang="ts">
import { SynVueOption, SynVueSelect } from '@synergy-design-system/vue';
import { computed } from 'vue';

const blub = computed( () => {
  return 'bar';
});
</script>
<template>
  <SynVueSelect :value="blub">
    <SynVueOption value="foo">Foo</SynVueOption>
    <SynVueOption value="bar">Bar</SynVueOption>
  </SynVueSelect>
</template>
```

Example Code for angular (exchange the home.component.ts with this code) : 
```
import { Component, signal } from '@angular/core';

@Component({
  selector: 'home',
  standalone: false,
  styleUrls: ['./home.styles.css'],
  template: `
  <syn-select [value]="selected()">
    <syn-option value="foo">Foo</syn-option>
    <syn-option value="bar">Bar</syn-option>
    <syn-option value="baz">Baz</syn-option>
  </syn-select>
`,
})
export class Home {
  selected = signal<string | string[]>('bar');
}

```

For some reason for react it seemed to already work fine without the fix

- have a look at the storybook table so the "value" attribute is appearing correctly there: http://localhost:6006/?path=/docs/components-syn-select--docs

## ✅ DoD


packages/components/custom-elements-manifest.config.js
```diff
@@ -225,6 +225,26 @@ export default {
       }
     },
 
+    // Since these changes from shoelace (https://github.com/shoelace-style/shoelace/pull/2255), where they removed the @property from "value" and changed the "defaultValue" and "value" handling internally
```
schilchSICKAG, line 228: I am not sure if we should leave the new "defaultValue" field in here. I think it might be confusing. Consider the following react code:

```tsx
        <SynSelect
          id="select-role"
          label="Current position"
          name="role"
          required
          defaultValue="frontend"
          value={formData.role}
        >
          <SynOptgroup label="Developers">
            <SynOption value="backend">Backend Developer</SynOption>
            <SynOption value="frontend">Frontend Developer</SynOption>
          </SynOptgroup>
          <SynOptgroup label="Other">
            <SynOption value="lead">Team Lead</SynOption>
            <SynOption value="other">Other (please specify in comment section below)</SynOption>
          </SynOptgroup>
        </SynSelect>
```

In this case, I am assigning both the value and the defaultValue. The defaultValue will only be visible when using a form reset, otherwise the value property will always be preferred.

schilchSICKAG (commented):
It works as intended. However I am not sure if we should remove the  "defaultValue" from the wrappers as it may lead to problems when users do not understand the distinguishion between defaultValue and value. What do you think?

schilchSICKAG (approved):

packages/components/custom-elements-manifest.config.js
schilchSICKAG: Talked about this, defaultValue was always there, so no need to hide it.

PR #722, opened 2024-12-19
fix: 🐛 Indentation of nested nav-items in side-nav is not shown correctly for firefox

### LINKED ISSUE #708: fix: 🐛 syn-side-nav indentation-story has no indentation

# Version

The bug occurred with the following version of the component library: 2.14.3

# Component

The issue arises when using the following component(s): syn-side-nav

# Environment

The issue occurred with the following operating system(s): macOS 15.1.1
The issue occurred in the following browser(s): Firefox 133

# Description

In the indentation story, there is no indentation visible
https://synergy-design-system.github.io/?path=/story/components-syn-side-nav--indentation&globals=viewport:defaultViewPort

## Storybook
Image

## Figma
Image

# Acceptance criteria

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR fixes the issue that in firefox the syn-side-nav with nested syn-nav-items does not show an indentation for them.

### 🎫 Issues
Closes #708 

## 👩‍💻 Reviewer Notes
This problem occurs only in firefox. 
Using `getComputedStyles` in firefox with a parent window with no presentation (e.g. display: none) results in an empty object. For more information about this see [the firefox notes on caniuse](https://caniuse.com/?search=getComputedStyle).

Nesting the syn-nav-items without an syn-side-nav works fine in firefox, but our syn-side-nav uses a syn-drawer in itself, which leads to this behavior. It has nothing to do with the side-nav or the drawer but just the fact of slotting the nav-item into a custom-element which uses a custom element in it. This seems to trigger the behavior in firefox only.
My guess would be that the inner custom element (in our case the syn-drawer) is not yet finished and has therefore no presentation, which leads to this.

The solution is moving everything in the handleSlotChange for firefox in the next render cycle.

## 📑 Test Plan

Check out the indentation syn-side-nav story in firefox, that it looks correct now: http://localhost:6006/?path=/story/components-syn-side-nav--indentation&globals=viewport:defaultViewPort

## ✅ DoD

### DIFF
--- a/packages/components/src/components/nav-item/nav-item.component.ts
+++ b/packages/components/src/components/nav-item/nav-item.component.ts
@@ -223,10 +223,25 @@ export default class SynNavItem extends SynergyElement {
    * Automatically add the correct level of indentation for sub items if none is provided
    */
   private handleSlotChange() {
+    const computedStyle = getComputedStyle(this);
+    const lengthStyle = computedStyle.length;
+
+    // This workaround is needed for firefox.
+    // When the nav-item is used inside a custom-element with a custom-element in it
+    // (e.g syn-side-nav with integrated syn-drawer), the getComputedStyle property is not yet there
+    // Moving it in the next render cycle works.
+    // Probably related to the firefox special behavior. See: https://caniuse.com/mdn-api_window_getcomputedstyle
+    if (lengthStyle === 0) {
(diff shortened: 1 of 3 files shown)


schilchSICKAG (approved):
Nice find and fix 👍

Issue #714, opened 2024-12-12
fix: 🐛 syn-combobox: Slotted syn-options do not honor applied styles


schilchSICKAG:
# Checklist (for the issue creator)

- [x] I have verified the bug exists in the latest version of `@synergy-design-system/components` (e.g. via `pnpm install @synergy-design-system/components`).
- [x] I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...).
- [ ] I would be willing to provide a PR for this issue.

# Version

The bug occurred with the following version of the component library: 2.15.0

# Component

The issue arises when using the following component(s): `<syn-combobox>`

# Environment

The issue occurred with the following operating system(s): All
The issue occurred in the following browser(s): All

# Description

When setting css styles on the `<syn-option>` component, those styles are not honored when rendering in a `<syn-combobox>`. This can be seen in a simple example:

```html
<style>
syn-option::part(label) {
  background-color: red;
}
</style>

<syn-combobox>
      <syn-option>Option 1</syn-option>
      <syn-option>Option 2</syn-option>
      <syn-option>Option 3</syn-option>
</syn-combobox>

<syn-select max-options-visible="3">
  <syn-option value="Option\_1">Option 1</syn-option>
  <syn-option value="Option\_2">Option 2</syn-option>
  <syn-option value="Option\_3">Option 3</syn-option>
</syn-select>  
```

## Reason:

It is possible that this is happening because we are cloning all `<syn-option>` into kind of a "virtual" position to make them filterable. In this case, we would have to think about not cloning the nodes at all (which also seems to be the root cause of #712).

# Acceptance criteria

- [ ] All DoD items successfully completed
- [ ] Syn-Selects are styled as the user wants it. For the example above, all appearances backgrounds for `<syn-option>` in a `<syn-combobox>` should be red.

Issue #712, opened 2024-12-10
fix: 🐛 Combobox with optgroup not working in react and vue


kirchsuSICKAG:
# Checklist (for the issue creator)

- [x] I have verified the bug exists in the latest version of `@synergy-design-system/components` (e.g. via `pnpm install @synergy-design-system/components`).
- [x] I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...).
- [x] I would be willing to provide a PR for this issue.

# Version

The bug occurred with the following version of the component library: 2.14.4

# Component

The issue arises when using the following component(s): syn-combobox with syn-optgroup

# Environment

The issue occurred with the following operating system(s): all
The issue occurred in the following browser(s): all

# Description
Using a syn-combobox with syn-optgroup does not show the optgroup label.
This happens for react and vue applications, but not for vanilla or angular.

Example code for reproduction:

```
// React
    <SynCombobox>
      <SynOptgroup label="My Group">
        <SynOption value="Black">Black</SynOption>
        <SynOption value="Blue">Blue</SynOption>
        <SynOption value="Brown">Brown</SynOption>
      </SynOptgroup>
    </SynCombobox>
```

```
// Vue

  <SynVueCombobox>
    <SynVueOptgroup label="My Group">
      <SynVueOption>Black</SynVueOption>
      <SynVueOption>Blue</SynVueOption>
      <SynVueOption>Brown</SynVueOption>
    </SynVueOptgroup>
  </SynVueCombobox>

```

This results in following output: 
![Image](https://github.com/user-attachments/assets/1009fe98-51e4-4cc9-b9e7-97b1c67cc9df)

# Acceptance criteria

- [ ] All DoD items successfully completed

schilchSICKAG:
Reason: We are cloning the nodes here. However, both react and vue have heuristics how they set values for web-components. From the react documentation about web-components (https://react.dev/blog/2024/12/05/react-19#support-for-custom-elements):

> Client Side Rendering: props that match a property on the Custom Element instance will be assigned as properties, otherwise they will be assigned as attributes.

The problem here is that `element.cloneNode` does not clone the children __or__ the attributes if the first argument is not set to `true`.

schilchSICKAG:
Idea: If we still want to clone the nodes instead of writing to the DOM, we could do something like this instead:

```ts
const clonedNode = myOption.cloneNode(true);

while (clonedNode.firstChild) {
[...]

PR #723, opened 2024-12-20
fix: 🐛 Various issues with syn-combobox

### LINKED ISSUE #719: fix: 🐛 Toggle open state of combobox with expand icon not always working

# Version

The bug occurred with the following version of the component library: 2.15.0

# Component

The issue arises when using the following component(s): syn-combobox

# Environment

# Description

When using the syn-combobox in a react application within a [MUI dialog](https://mui.com/material-ui/react-dialog/), the combobox does not toggle its open state correctly, if clicking on the expand icon. It will immediately close again.

Image

# Acceptance criteria

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR fixes multiple issues with the syn-combobox:
- syn-option's do not honor applied styles 
- syn-optgroup not working correctly (e.g. no slot is working, the label property in react)
- toggling open state with the expand icon not always working correctly

the first two issues are fixed by no longer cloning the syn-options and working with the clones, but just working with the syn-options themselves.

Additionally this PR fixes a performance issue, which was added via shoelace update in the syn-option.

### 🎫 Issues
Closes #719 , #712, #714 

### Issue #719 
When using the combobox within a [MUI dialog](https://mui.com/material-ui/react-dialog/) the open state can now be toggled correctly via expand icon:
To test this install MUI in the react demo application: 
```
pnpm add @mui/material @emotion/react @emotion/styled
```

and use following code, to check that the open state is toggled correctly:

```
import { Dialog, DialogContent } from '@mui/material';
import { SynCombobox, SynOption } from '@synergy-design-system/react';
import type { FC } from 'react';

export const Home: FC = () => (
  <>
    <Dialog open>
      <DialogContent>
        <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</p>
        <SynCombobox>
          <SynOption>opt 1</SynOption>
          <SynOption>opt 2</SynOption>
          <SynOption>opt 3</SynOption>
        </SynCombobox>
      </DialogContent>
    </Dialog>
  </>
);
```

### Issue #712 

The syn-optgroup is now working correctly without the need of workarounds. The label property in react and vue is now working correctly and the all slot (label, prefix, suffix) are also working.

Before: https://codepen.io/kirchsuSICKAG/pen/pvzegMJ?editors=1110
    --> first optgroup uses label slot which is not working
    --> second optgroup uses prefix and suffix slot which are not working

### Issue #714 
Now the applied styles to syn-option are honored as we no longer clone the nodes, but use the original ones.

Before: https://codepen.io/kirchsuSICKAG/pen/pvzegMJ?editors=1110
   --> syn-options should have red text color, but is not applied

## ✅ DoD

### DIFF
--- a/packages/components/src/components/combobox/combobox.component.ts
+++ b/packages/components/src/components/combobox/combobox.component.ts
@@ -1,7 +1,6 @@
 import type { CSSResultGroup } from 'lit';
 import { classMap } from 'lit/directives/class-map.js';
 import { html } from 'lit';
-import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 import { property, query, state } from 'lit/decorators.js';
 import { animateTo, stopAnimations } from '../../internal/animate.js';
 import { defaultValue } from '../../internal/default-value.js';
@@ -18,11 +17,12 @@ import SynergyElement from '../../internal/synergy-element.js';
 import SynIcon from '../icon/icon.component.js';
 import SynPopup from '../popup/popup.component.js';
 import type { SynergyFormControl } from '../../internal/synergy-element.js';
-import type SynOption from '../option/option.component.js';
+import SynOption from '../option/option.component.js';
 import type SynOptGroup from '../optgroup/optgroup.js';
 import styles from './combobox.styles.js';
 import customStyles from './combobox.custom.styles.js';
 import {
+  createOptionFromDifferentTypes,
   filterOnlyOptgroups, getAllOptions, getAssignedElementsForSlot, normalizeString,
 } from './utils.js';
 import { scrollIntoView } from '../../internal/scroll.js';
@@ -105,6 +105,8 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
   /** The last value of a syn-option, that was selected by click or via keyboard navigation */
   private lastOptionValue = '';
 
+  private isOptionRendererTriggered = false;
+
   @query('.combobox') popup: SynPopup;
 
   @query('.combobox__inputs') combobox: HTMLSlotElement;
@@ -115,8 +117,6 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
 
   @query('.combobox__listbox') listbox: HTMLSlotElement;
 
-  @query('.listbox__options') filteredWrapper: HTMLSlotElement;
-
   @query('slot:not([name])') private defaultSlot: HTMLSlotElement;
 
   @state() private hasFocus = false;
@@ -125,7 +125,9 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
 
   @state() selectedOption: SynOption | undefined;
 
-  @state() filteredOptions: Array<SynOption | SynOptGroup> = [];
+  @state() numberFilteredOptions = 0;
+
+  @state() cachedOptions: SynOption [] = [];
 
   /** The name of the combobox, submitted as a name/value pair with form data. */
   @property() name = '';
@@ -206,7 +208,11 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
    */
   // eslint-disable-next-line class-methods-use-this
   @property() filter: (option: SynOption, queryString: string) => boolean = (option, queryStr) => {
-    const normalizedOption = normalizeString(option.getTextLabel());
+    let content = option?.textContent || '';
+    if (option instanceof SynOption) {
+      content = option.getTextLabel();
+    }
+    const normalizedOption = normalizeString(content);
     const normalizedQuery = normalizeString(queryStr);
     return normalizedOption.includes(normalizedQuery);
   };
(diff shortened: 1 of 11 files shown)


packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -712,44 +696,63 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
...
   private createComboboxOptionsFromQuery(queryString: string) {
-    const optgroups: SynOptGroup[] = [];
+    this.numberFilteredOptions = 0;
+    this.isOptionRendererTriggered = true;
 
-    this.filteredOptions = this.getSlottedOptions()
-      .filter(option => this.filter(option, queryString) || queryString === '')
-      .map(option => {
-        const clonedOption = option.cloneNode(true) as SynOption;
+    // This is needed for angular. For some reason the handleDefaultSlotChange is not triggered
```
schilchSICKAG, line 704: Still not sure about this one. It does not seem to have a noticable effect when using an empty `<syn-combobox>`, but I want to point out that, whenever `createComboboxOptionsFromQuery` it will do another run of `cacheSlottedOptionsAndOptgroups` when there are no children detected. Could we detect it otherwise? Not sure.

packages/components/src/components/combobox/option-renderer.ts
```diff
@@ -26,6 +26,28 @@ export const highlightOptionRenderer: OptionRenderer = (option: SynOption, query
     return option;
   }
 
+  const combobox = option.closest('syn-combobox');
+  // check if there is already a style for highlighting appended if not append one
+  // This is unfortunately needed as it is not possible to do something like this in css:
+  // .listbox__options ::slotted(syn-option mark)
+  if (combobox && !combobox?.querySelector('#syn-highlight-style')) {
```
schilchSICKAG, line 33: I understand where this is coming from. However, could we add this to the `<syn-option>` shadow root instead? Did not test it, but we could also add a custom marker, e.g. `<mark class="highlighted-element">` and use this in the `syn-option` as `.label::slotted(.highlighted-element`)?

As it is, the style is not overridable from the outside, not even with important. Don´t think this is a big problem, fwiw, so we may also go ahead with this approach.

packages/components/src/components/combobox/utils.ts
```diff
@@ -49,3 +51,46 @@ export const filterOnlyOptgroups = (items: HTMLElement[]) => items.filter(isOptg
...
+ * This function takes an input of type `TemplateResult`, `string`, or `HTMLElement`
+ * and creates an SynOption from it.
+ *
+ * @param option - The option which can be of type `TemplateResult`, `string` or `HTMLElement`.
+ * @returns The created SynOption or undefined if the option was not a valid type.
+ */
+export const createOptionFromDifferentTypes = (option: TemplateResult | string | HTMLElement) => {
+  if (!option) return undefined;
+
+  const getOptionOrUndefined = (element: HTMLElement) => (element.tagName.toLocaleLowerCase() === 'syn-option' ? element as SynOption : undefined);
```
schilchSICKAG, line 78: This function will be created one time in memory for each call of `createOptionFromDifferentTypes`. As this is just a simple test, I would go with either:

a) Placing it above the `createOptionFromDifferentTypes` (easiest path) or
b) it looks like we always return a html element via `createElementFromTemplateResult`, so maybe the `undefined` check is not even needed? In this case, we could return the result from `createElementFromTemplateResult` directly.

packages/components/src/components/optgroup/optgroup.styles.ts
```diff
@@ -52,4 +52,8 @@ export default css`
     color: var(--syn-color-neutral-800);
     font-size: var(--syn-spacing-large);
   }
+
+  .optgroup__options ::slotted(syn-option[hidden]) {
```
schilchSICKAG, line 56: Just curious, where is this coming from? Should an attribute of hidden noch automatically hide the item? If this is for hiding empty optgroups I think it should instead be a `has` selector, but those don´t work with slotted unfortunately :(

schilchSICKAG (changes requested):
Just a couple of smaller changes, and we should be good to go 👍

packages/components/src/components/combobox/option-renderer.ts
kirchsuSICKAG: I like the idea 👍
I was only thinking about doing something like this in the option style `.option__label::slotted(mark)`. But I didn't like that we automatically style the `<mark/>` element always.
Was not thinking about i could just add a class for this instead.

packages/components/src/components/combobox/utils.ts
kirchsuSICKAG: I would go for a).

If someone is doing something like `<div>My content</div>` in the `getOption` method, it is a htmlelement, but we should not allow this. It would break other things (they would have broken before, but we had no check). Now I react in the code itself by just using the "original" syn-option without the "getOption" changes, if it is no syn-option element: 
image

 And we also state that only syn-options are allowed in the default slot: 
image

packages/components/src/components/optgroup/optgroup.styles.ts
kirchsuSICKAG: We have the components.style.ts which handles that elements with the attribute `hidden` on them are not displayed. But unfortunately this is not done for slotted elements just with elements in the shadowRoot:
Bildschirmfoto 2024-12-20 um 14 14 53

If not doing this, options are visible although they should not be, if there is another option in an optgroup which IS visible. See following gif: 
optgroup

PR #724, opened 2025-01-03
feat: ✨ Tree shakable angular components

### LINKED ISSUE #648: feat: ✨ Make components for Angular tree shakeable

# Component

I suggest to add the feature to the following component: All, Angular Wrappers

# Story

As a developer using the Angular components of the Synergy Design System, I would like to have the possibility to load single components from the `@synergy-design-system/angular` package so that the bundle size stays small.

# Acceptance criteria

- [x] `NGPackager` must be configured in a way that allows the import of single components
- [x] The angular demo must be adjusted for at least one component as a showcase
- [ ] Documentation has to be updated to reflect those changes (e.g. https://synergy-design-system.github.io/?path=/docs/packages-components-angular-wrapper--docs#3-load-the-synergy-module-to-load-all-components must finally receive the "loading single components" sub entry)

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds support for angular to be able to import only specific components and not always get everything.

### 🎫 Issues

Closes #648.

## ✅ DoD


packages/components/scripts/jobs/shared.js
```diff
@@ -415,3 +415,54 @@ export const enrichComponentAttributes = (component) => {
...
+
+  await formatFile(packageJsonPath, 'json-stringify');
+});
+
+/**
+ * Creates an ng-package.json file for the given entry file
+ * @param {*} entryFile - the entry file it is created for
+ * @param {*} outDir - the path where it is written to
+ */
+export const createNgPackageJson = (entryFile, outDir) => {
```
schilchSICKAG, line 457: Beware that this function is async in fact as you use `fs.writeFile` instead of `fs.writeFileSync`. Either use it asynchronously by adding an `await` to `fs.writeFile` or use `fsSync.writeFile` instead to prevent possible errors

packages/components/scripts/jobs/angular/createComponents.js
```diff
@@ -148,14 +153,17 @@ export const runCreateComponents = job('Angular: Creating components', async (me
 
     index.push({
       name: `${component.name}Component`,
-      outputPath: `./${componentFileName.slice(0, -3)}`,
+      outputPath: `@synergy-design-system/angular/components/${component.tagNameWithoutPrefix}`,
     });
 
     fs.writeFileSync(componentPath, source, 'utf8');
+    createNgPackageJson(componentFileName, componentDir);
```
schilchSICKAG, line 160: Either `await` this or, if this is not possible, make `createNgPackageJson` sync. See comment in `createNgPackageJson` for further details. This may lead to problems later on.

packages/components/scripts/jobs/shared.js
schilchSICKAG: Also, as this is specific to angular, we might move this function to `jobs/angular/shared` for the sake of namespacing if nothing else is speaking against it.

packages/components/scripts/jobs/shared.js
```diff
@@ -415,3 +415,54 @@ export const enrichComponentAttributes = (component) => {
 
   return attrAndProps;
 };
+
+/**
+ * Adjust the package.json exports field located in packageDir by removing the 'src' prefix
+ * @param {string} label The label
+ * @returns {job}
+ */
+export const runAdjustPackageExports = (label) => job(label, async (
```
schilchSICKAG, line 424: This function is only needed for angular, I think it should be moved to `jobs/angular/shared.js` to make the namespacing clearer

packages/components/scripts/jobs/angular/createExports.js
```diff
@@ -11,10 +11,10 @@ export const runCreateExports = job('Angular: Creating exports file...', async (
   const outFile = path.join(outDir, 'index.ts');
```
schilchSICKAG, line 11: The outfile has changed, so we should adjust the functions comment to the new path

packages/components/scripts/jobs/angular/createFormsModule.js
```diff
@@ -61,7 +61,12 @@ const createSelectors = selectors => selectors
  * Creates the synergy angular module, located at packages/angular/src/modules/synergy.module.ts
```
schilchSICKAG, line 61: This seems to be a copy leftover by me. Either remove the comment or adjust the path

packages/components/scripts/jobs/angular/createFormsModule.js
```diff
@@ -128,22 +133,20 @@ export class SynFileValueAccessor extends DefaultValueAccessor { }
...
+  imports: [
+    SynergyValidatorsModule
   ],
-  imports: [],
 })
 export class SynergyFormsModule {}
 `.trim();
 
   await fs.writeFile(outFile, `${output}\n`, 'utf8');
+  createNgPackageJson(fileName, formsModuleDir);
```
schilchSICKAG, line 151: The sync/async issue strikes again :). This is the third place I saw it, think it would be good to make `createNgPackageJson` to prevent dozens of `await` calls. ;)

packages/components/scripts/jobs/angular/createNgModule.js
```diff
@@ -1,14 +1,21 @@
...
 import path from 'path';
-import { createHeader, getAllComponents, job } from '../shared.js';
+import {
+  createHeader, createNgPackageJson, getAllComponents, job,
+} from '../shared.js';
 
 const headerComment = createHeader('angular');
 
 /**
  * Creates the synergy angular module, located at packages/angular/src/modules/synergy.module.ts
```
schilchSICKAG, line 10: Comment points to the wrong path

packages/components/scripts/jobs/angular/createValidatorDirectives.js
```diff
@@ -120,7 +125,23 @@ export const SYN_CHECKBOX_REQUIRED_VALIDATOR: Provider = {
   host: {'[attr.required]': '_enabled ? "" : null'},
 })
 export class SynCheckboxRequiredValidator extends CheckboxRequiredValidator {}
+
+@NgModule({
```
schilchSICKAG, line 129: As I am no angular expert I honestly don´t know, but do we need to make this `standalone`? If omitted, it will be standalone by default for angular19, lower versions will have this set to false.

packages/angular/tsconfig.json
```diff
@@ -1,36 +1,190 @@
...
 
     /* Linting */
     "strict": true,
-    "noUnusedLocals": true,
     "noUnusedParameters": true,
-    "noFallthroughCasesInSwitch": true
+    "noFallthroughCasesInSwitch": true,
+
+    "paths": {
+      // We need this path resolves for typescript to be able to work correctly
```
schilchSICKAG, line 24: Always funny that a `tsconfig.json` is not a real json file :)

packages/angular/tsconfig.json
```diff
@@ -1,36 +1,190 @@
 {
   "compilerOptions": {
-    "rootDir": "src",
-    "allowSyntheticDefaultImports": true,
-    "composite": false,
-    "target": "ES2020",
-    "useDefineForClassFields": true,
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "baseUrl": ".",
+    "target": "es2022",
```
schilchSICKAG, line 4: Do we really need this whole bunch of new settings here? Our other targets use `es2020`, as was the angular module before. Did it not work with the default settings that where already applied?

packages/angular/tsconfig.json
```diff
@@ -1,36 +1,190 @@
...
+      "@synergy-design-system/angular/components/tooltip": [
+        "./components/tooltip/tooltip.component.ts"
+      ],
+      "@synergy-design-system/angular/components/validate": [
+        "./components/validate/validate.component.ts"
+      ]
+    }
   },
-  "include": ["src"],
   "angularCompilerOptions": {
```
schilchSICKAG, line 175: What is the reason for those new options?

packages/components/scripts/jobs/angular.js
```diff
@@ -24,21 +26,29 @@ export const runCreateAngularWrappers = async ({
   const manifest = path.join(componentDistDir, '/custom-elements.json');
   const metadata = JSON.parse(fs.readFileSync(manifest, 'utf8'));
 
-  // Internal react package paths for usage in sub packages
-  const outDir = path.join(angularPackageDir, './src');
+  // Internal angular package paths for usage in sub packages
+  const outDir = path.join(angularPackageDir);
```
schilchSICKAG, line 30: We will have to adjust our `.vscode/settings.json` file to also make the new file structure read only

packages/components/scripts/jobs/shared.js
```diff
@@ -415,3 +415,54 @@ export const enrichComponentAttributes = (component) => {
...
+});
+
+/**
+ * Creates an ng-package.json file for the given entry file
+ * @param {*} entryFile - the entry file it is created for
+ * @param {*} outDir - the path where it is written to
+ */
+export const createNgPackageJson = (entryFile, outDir) => {
+  const ngPackageJson = {
+    $schema: '../../../node_modules/ng-packagr/ng-package.schema.json',
```
schilchSICKAG, line 459: The schema paths never match correctly, as they are relative to the directory where the file resides. When whe know the outDir, we can create this relatively, e.g. for the file in the components folder it should be `"$schema": "../node_modules/ng-packagr/ng-package.schema.json",`, for the accordion component it has to be `"$schema": "../../node_modules/ng-packagr/ng-package.schema.json",` and so on.

packages/_private/angular-demo/src/demoformvalidate/demoformvalidate.component.ts
```diff
@@ -1,7 +1,11 @@
...
-import { highlightOptionRenderer, serialize } from '@synergy-design-system/components';
+import type { SynChangeEvent } from '@synergy-design-system/components';
+import { serialize } from '@synergy-design-system/components/utilities/form.js';
+import { highlightOptionRenderer } from '@synergy-design-system/components/components/combobox/option-renderer.js';
 import { Component, ElementRef, ViewChild } from '@angular/core';
-import { FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { CommonModule } from '@angular/common';
+import { FormBuilder, FormGroup, ReactiveFormsModule, Validators } from '@angular/forms';
+import { UsedSynergyComponentsModule } from '../modules/used-synergy.module';
+import { DemoFieldSetModule } from '../modules/demofieldset.module';
```
schilchSICKAG, line 8: I like this very much ❤️! Nice that you also updated the docs, too :)

packages/angular/tsconfig.json
schilchSICKAG: Interesting, so our own `tsconfig` was just there so the compiler did not yell at us :). I am fine with this as it does not change anything that was not different as we wanted it before (if this makes any sense to you :))

packages/components/scripts/jobs/shared.js
kirchsuSICKAG: done

packages/components/scripts/jobs/shared.js
kirchsuSICKAG: done

packages/components/scripts/jobs/angular/createFormsModule.js
kirchsuSICKAG: done

packages/components/scripts/jobs/angular/createFormsModule.js
kirchsuSICKAG: done

packages/components/scripts/jobs/angular/createComponents.js
kirchsuSICKAG: done

packages/components/scripts/jobs/angular/createExports.js
kirchsuSICKAG: done

packages/components/scripts/jobs/angular/createNgModule.js
kirchsuSICKAG: done

packages/_private/angular-demo/tsconfig.json
```diff
@@ -14,12 +14,28 @@
...
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "importHelpers": true,
     "target": "ES2022",
     "module": "ES2022",
     "useDefineForClassFields": false,
-    "lib": ["ES2022", "dom", "DOM.Iterable"]
+    "lib": ["ES2022", "dom", "DOM.Iterable"],
+    "paths": {
+      // This is only needed because we are in a monorepo. In an outside applications this is not needed
```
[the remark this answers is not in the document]
kirchsuSICKAG: Unfortunately not. It seems to have something to do with monorepos and that typescript does not seems to be able to resolve the correct paths as the packages are only locally...
I tested it in an angular application, not belonging to this monorepo and there was no problem.
But without this pathh entries out angular demo is no longer working correctly. 
That is why I also added the comment above, so we know why this is happening.
But if you say there should be no comments in a json I can also remove it.

packages/components/scripts/jobs/angular.js
kirchsuSICKAG: nice catch!

packages/components/scripts/jobs/shared.js
kirchsuSICKAG: nice catch!

packages/angular/tsconfig.json
kirchsuSICKAG: I added the angularCompilerOptions, which we already used because of the default tsconfig.json of the ng-packagr. I added them to our tsconfig.json so our compiler will work with the same settings as the build angular options

packages/components/scripts/jobs/angular/createValidatorDirectives.js
kirchsuSICKAG: No, we don't need to do this and we also can't do it , as there is no `standalone` option on `NgModule`.
It implicitly says `I am the old way` because with the new handling of angular `NgModules` are no longer needed.

packages/components/scripts/jobs/angular/createFormsModule.js
```diff
@@ -128,22 +133,20 @@ export class SynFileValueAccessor extends DefaultValueAccessor { }
     SynDefaultValueAccessor,
```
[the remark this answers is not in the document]
kirchsuSICKAG: The validators went to the `SynergyValidatorsModule`. I needed to create this new module for our validators, as ng-packagr didn't let me do it the old way.
The validators are in another "library" because they have their own ng-package.json. And somehow because of that they are getting declared in the other "library" and can't be declared here anymore. It will throw an error on build. One solution would have been to change our validators to standalone. But then we would need to do a major change. Via this solution I just created a new module for our validators, to be able to import and export them here without build errors.
For our users there should be no visible changes. Except that there is one new exported module component which they could use

packages/angular/tsconfig.json
schilchSICKAG: That was just a general observation. Its just the file ending and parsed completely different than other json files. I just think its funny that people add comments there without even recognizing they are making the file illegal. The comment makes sense, so lets just let it stay that way.

schilchSICKAG (approved):
Nice! Lets add `sideEffects: false` to `package.json` and launch a beta 👍

packages/components/scripts/jobs/angular.js
```diff
@@ -24,21 +25,29 @@ export const runCreateAngularWrappers = async ({
...
```
klaarseSICKAG, line 29: meaningful?

packages/react/README.md
```diff
@@ -7,9 +7,17 @@ This package aims for an improved UX when used in React applications:
 - Auto-completion
 - Event handling
 
+## Known issues and limitations
+
+Got any problems using our React wrappers? Currently there are no known issues relating to them. But please take a look at [our list of known issues and limitations of the web components](https://synergy-design-system.github.io/?path=/docs/limitations-known-issues-and-limitations--docs) before [creating a ticket](https://github.com/synergy-design-system/synergy-design-system/issues/new?assignees=&labels=&projects=&template=generic-bug.md&title=fix%3A+%F0%9F%90%9B+), maybe you can find you problem there.
```
klaarseSICKAG, line 12:
```suggestion
Got any problems using our React wrappers? Currently there are no known issues relating to them. But please take a look at [our list of known issues and limitations of the web components](https://synergy-design-system.github.io/?path=/docs/limitations-known-issues-and-limitations--docs) before [creating a ticket](https://github.com/synergy-design-system/synergy-design-system/issues/new?assignees=&labels=&projects=&template=generic-bug.md&title=fix%3A+%F0%9F%90%9B+), maybe you can find your problem there.
```

kirchsuSICKAG:
Doing sideEffects: false on the package.json of the angular wrapper seems to work correctly with tree shaking.
I tested it via our demo angular application by:
- adapt all @synergy-design-system/components to import only needed things directly. instead of 
```typescript
 import { highlightOptionRenderer, serialize } from '@synergy-design-system/components';
```
 doing 
 ```typescript
 import { serialize } from '@synergy-design-system/components/utilities/form.js';
import { highlightOptionRenderer } from '@synergy-design-system/components/components/combobox/option-renderer.js';
```
- don't use the SynergyComponentsModule but create a custom module with only needed components
```typescript
import { NgModule } from '@angular/core';
import {  SynButtonComponent, SynCheckboxComponent, SynComboboxComponent } from '@synergy-design-system/angular';

const components = [
  SynButtonComponent,
  SynCheckboxComponent,
  SynComboboxComponent,
// ...
// all the other components
];

@NgModule({
  imports: components,
  exports: components,
})
export class UsedSynergyComponentsModule {}
```

The result size of the created main.js file looks following:
- `sideEffects:true`:  817222 byte
- `sideEffects:false`: 705055 byte

I tried running both build results via an http-server and both builds worked the same way.

packages/components/scripts/jobs/angular.js
kirchsuSICKAG: for me this comment is fine. We have the same comment for all other wrappers, but had a copy paste here from react, this is why I changed it to `angular`
The paths, which follow the comment, are all paths that the angular build needs for sub actions

packages/react/README.md
kirchsuSICKAG: fixed

PR #733, opened 2025-01-10
feat: ✨ Allow to set global defaults for Synergy Component props

### LINKED ISSUE #674: feat: ✨ global sizing
# Story

As a user of the Synergy Design System, I would like to have only one single setting for sizing elements so that I don't have to set the size for each element.

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the ability to control the default sizes (and possibly other element settings) for all components. It does so by creating static defaults during the build of the components and providing two utilities that make it possible to override these. This makes this approach framework agnostic via a simple api.

Currently missing features:

- [ ] Create correct types for `setDefaultSettingsForElement` (currently `Record<string, unknown>` to make it work in typescript without direct casts.
- [x] Add unittests for settings and setting decorator
- [x] Add integrationtests
- [x] Hide event support behind a feature flag
- [x] Find a way to attach the events to the listening components directly.

### 🎫 Issues

Closes #674 

## 👩‍💻 Reviewer Notes

Play around with the demo projects. Each now has a size adjustment setting in the header, which toggles the sizes between three predefined ones (small, medium and large). When toggling one of the options, it will automatically adjust all components that do NOT have a initial value that is different from the initial setting (e.g. a `<syn-button>` with size large will stay large, even when the size is set to small.

This demo use the **experimental** events API for the default settings.

Make sure to read the documentation first and see if you can find out how to use it. It will be available when launching storybook.
You can use `setGlobalDefaultSettings` directly before mounting or any other time during the launch of the application. Just play around with the demos to get a feel for this. 

## ✅ DoD


packages/docs/stories/GlobalSettings.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+```typescript
+import { setDefaultSettingsForElement } from '@synergy-design-system/components';
+
+// Sets the size to small and variant to primary for all newly rendered SynButton components
+setDefaultSettingsForElement('SynButton', {
+  size: 'small',
+  variant: 'primary',
+});
+```
```
kirchsuSICKAG, line 50: It seems like there is something not completely correct with the typings. Copying this and using it in our demos leads to following typescript error:
Bildschirmfoto 2025-01-13 um 14 01 26
Bildschirmfoto 2025-01-13 um 14 01 16

packages/docs/stories/GlobalSettings.mdx
```diff
@@ -0,0 +1,134 @@
...
+You can also set defaults scoped for a specific component using the `setDefaultSettingsForElement` function.
+This works in the reverse way of `setGlobalDefaultSettings` and allows you to set defaults for a specific component.
+
+```typescript
+import { setDefaultSettingsForElement } from '@synergy-design-system/components';
+
+// Sets the size to small and variant to primary for all newly rendered SynButton components
+setDefaultSettingsForElement('SynButton', {
+  size: 'small',
+  variant: 'primary',
```
kirchsuSICKAG, line 48: We don't have a button with `type=primary` 😉

.vscode/settings.json
```diff
@@ -200,7 +201,7 @@
     "packages/components/src/internal/scrollend-polyfill.ts": true,
     "packages/components/src/internal/slot.ts": true,
     "packages/components/src/internal/string.ts": true,
-    "packages/components/src/internal/synergy-element.ts": true,
+    "packages/components/src/internal/synergy-element.ts": false,
```
kirchsuSICKAG, line 204: should be true
```suggestion
    "packages/components/src/internal/synergy-element.ts": true,
```

.vscode/settings.json
```diff
@@ -192,6 +192,7 @@
     "packages/components/src/internal/active-elements.ts": true,
     "packages/components/src/internal/animate.ts": true,
     "packages/components/src/internal/default-value.ts": true,
+    "packages/components/src/internal/defaultSettings.ts": false,
```
kirchsuSICKAG, line 195: should be true 
```suggestion
    "packages/components/src/internal/defaultSettings.ts": true,
```

packages/components/src/internal/globalSettings.ts
```diff
@@ -0,0 +1,101 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import type { PropertyValues } from 'lit';
+import type SynergyElement from './synergy-element.js';
+import { type ComponentNamesWithDefaultValues, extractDefaultSettingsForElement } from './defaultSettings.js';
+import type { SynDefaultSettingsChangedEvent } from '../events/events.js';
+
+type Constructor<T = object> = new (...args: any[]) => T;
+
+export function globalSettingsDecorator() {
```
kirchsuSICKAG, line 9: I am thinking about calling it only `globalSettings` because does the `@` in the components where it is used not say that it is a decorator? So in use it is only 

```typescript
@globalSettings()
export default class SynRadioButton extends SynergyElement {
```

instead of 

```typescript
@globalSettingsDecorator()
export default class SynRadioButton extends SynergyElement {
```

What do you think?

packages/components/scripts/vendorism/custom/globalSettings.vendorism.js
```diff
@@ -0,0 +1,53 @@
...
+  'input',
+  'radio',
+  'radio-button',
+  'radio-group',
+  'range',
+  'select',
+  'switch',
+  'tag',
+  'textarea',
+];
```
kirchsuSICKAG, line 22: Do we really need to say which components we want to use for the globalSettingsDecorator? Can’t we do it by checking for the components if they have  the attribute with defaults? Otherwise we always need to keep in mind adding components here.
But maybe it is too complicated or is even not possible, I don't know

kirchsuSICKAG (changes requested):

packages/components/src/internal/globalSettings.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      if (Object.keys(defaults).length > 0) {
+        window.addEventListener('syn-default-settings-changed', this.#overrideWithGlobalSettings, {
+          capture: true,
+        });
+      }
+    }
+
+    disconnectedCallback() {
+      window.removeEventListener(
```
kirchsuSICKAG, line 52: I think the super.disconnectedCallback() is missing here

kirchsuSICKAG:
A really nice feature!

packages/docs/stories/GlobalSettings.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+<Meta title="Configuring global settings" />
+
+# Overriding global default settings for components
+
+Most of synergies components provide a way to configure their appearance and behavior through a set of properties.
+However, it can be cumbersome to set the same properties for all components in your application.
+
+Therefore, Synergy includes a low level API that allows you to set global defaults for the rendered components.
+This can be used to set the default size or other global settings that should be applied to all rendered components.
```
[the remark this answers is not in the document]
schilchSICKAG: Oh, it can be rendered another time. However, the new settings will only effect the components added __after__ it was called. Maybe we should suppress this as it may lead to faulty behaviour?

packages/docs/stories/GlobalSettings.mdx
schilchSICKAG: Yes, I still got an issue open with the types, which are currently defined as a `Record<string, unknown>`. Its on my list of open points above. Will work on this after refactoring the file system structure for the module (which I will do next)

packages/components/src/internal/globalSettings.ts
schilchSICKAG: Done

packages/components/scripts/vendorism/custom/globalSettings.vendorism.js
schilchSICKAG: Unfortunately yes as it stands, as this is part of vendorism currently. I maybe could adjust this as a part of the build step. This would make it possible to automatically annotate our own components, what I did per hand before. Will have a look into this 👍

packages/components/src/internal/globalSettings.ts
schilchSICKAG: Fixed

packages/components/scripts/vendorism/custom/globalSettings.vendorism.js
schilchSICKAG: Removed in latest version. Files are now automatically adjusted to include the decorator via the [global attribute whitelist configuration](https://github.com/synergy-design-system/synergy-design-system/blob/feat/674-global-sizing-version-extend/packages/components/scripts/jobs/synergy/createDefaultSettings.js#L14)

.vscode/settings.json
schilchSICKAG: Fixed with latest refactoring

schilchSICKAG:
Other than the currently last open point (types for `extractDefaultSettingsForElement` are set to `unknown` instead of their real values) I think I am ready here for a first try.

Please have a look at everything again, as there was a large refactoring going on underneath. The following things where added today:

- Complete filesystem layout and renaming underneath
- Exporting of the functionality
- Tests, tests tests
- A global cache to make rendering it more performant
- The feature now automatically adjusts all components (build was slightly changed for this to work)
- Build now accepts a `--synergy` parameter. This makes it possible to do a build of synergy only which was not possible before.
- Docs have been updated

packages/components/src/internal/defaultSettings/decorator.ts
```diff
@@ -0,0 +1,95 @@
...
+type Constructor<T = object> = new (...args: any[]) => T;
+
+/**
+ * Global settings decorator for a synergy component
+ * Used in conjunction with the global settings event
+ *
+ * @param name The components name. This is used to get the default values for the component
+ * @returns Decorated class for usage with synergy components
+ */
+export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
```
klaarseSICKAG, line 22: can the classname not be passed without mentioning it explicitly in each component class?

packages/docs/stories/GlobalSettings.mdx
```diff
@@ -0,0 +1,136 @@
...
+// Sets the size to small and variant to filled for all newly rendered SynButton components
+setDefaultSettingsForElement<SynButton>('SynButton', {
+  size: 'small',
+  variant: 'filled',
+});
+```
+
+## Live update via events (experimental)
+
+> ⚠️ This feature is considered experimental and therefore hidden behind a feature flag.
```
klaarseSICKAG, line 54: what is experimental with this event? the user might be alarmed by the naming and ask himself the same question: what can go wrong? do you have any doubts?

packages/docs/stories/GlobalSettings.mdx
```diff
@@ -0,0 +1,136 @@
+import { Meta } from "@storybook/blocks";
+import { extractDefaultSettingsForElement, INITIAL_DEFAULT_SETTINGS } from '../../components/src/synergy';
+
+<Meta title="Configuring global settings" />
```
klaarseSICKAG, line 4: I would expect this within Components. For what reason is that documentation on the root level?

klaarseSICKAG (commented):
Great Work‼️

packages/_private/angular-demo/src/shared.ts
```diff
@@ -1 +1,22 @@
+import { setGlobalDefaultSettings } from '@synergy-design-system/components';
+
 export const capitalize = (s: string) => s && s[0].toUpperCase() + s.slice(1);
+
+export const setGlobalSize = (size: 'small' | 'medium' | 'large') => {
+  setGlobalDefaultSettings({
+    size: {
+      SynButton: size,
+      SynCheckbox: size,
```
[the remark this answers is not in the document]
schilchSICKAG: Thats not a problem, will add this, too.

packages/_private/react-demo/src/shared.ts
```diff
@@ -1 +1,22 @@
+import { setGlobalDefaultSettings } from '@synergy-design-system/components';
+
 export const capitalize = (s: string) => s && s[0].toUpperCase() + s.slice(1);
+
+export const setGlobalSize = (size: 'small' | 'medium' | 'large') => {
+  setGlobalDefaultSettings({
+    size: {
+      SynButton: size,
```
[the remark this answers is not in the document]
schilchSICKAG: Thats not a problem, will add this, too.

packages/_private/vanilla-demo/src/utils.ts
```diff
@@ -1,3 +1,26 @@
...
+export type AvailableSizes = 'small' | 'medium' | 'large';
+
 export const noop = () => null;
 
 export const capitalize = (s: string) => s && s[0].toUpperCase() + s.slice(1);
+
+export const setGlobalSize = (size: AvailableSizes) => {
+  setGlobalDefaultSettings({
+    size: {
+      SynButton: size,
```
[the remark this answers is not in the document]
schilchSICKAG: Thats not a problem, will add this, too.

packages/_private/vue-demo/src/shared.ts
```diff
@@ -1 +1,22 @@
+import { setGlobalDefaultSettings } from '@synergy-design-system/components';
+
 export const capitalize = (s: string) => s && s[0].toUpperCase() + s.slice(1);
+
+export const setGlobalSize = (size: 'small' | 'medium' | 'large') => {
+  setGlobalDefaultSettings({
+    size: {
+      SynButton: size,
```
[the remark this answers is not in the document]
schilchSICKAG: Thats not a problem, will add this, too.

packages/components/src/internal/defaultSettings/decorator.ts
schilchSICKAG: Yes, that is unfortunate. I had it like this initially, passing `prototype.constructor.name`, which returns the className. This worked fine in dev environments.

Then... name mangling happened :(. When using a build system that aggressively optimizes its assets (e.g. vite build, webpack -p), the name of the class will be mangled, so `class SynButton extends SynergyElement` will become something like `class XY_AB extends A0`. Because of this, automatic guessing did not work.

Therefore, I opted for the pragmatic way of adding it in the class by itself, making sure that the value does not get mangled at all (because its now a string).

See [this commit](https://github.com/synergy-design-system/synergy-design-system/pull/733/commits/a5be7e9cf831fbede0058f8beec902fdd4dfc3fe) for further information.

packages/docs/stories/GlobalSettings.mdx
schilchSICKAG: For me, experimental feature means:

- Sort of a moving target. Even if it is running fine in a given version of the software, its signature or even functionality may change even during patch releases.
- Something that seems to work fine, but may need further refinement to make it bulletproof.
- A feature that may be removed at any stage if it seems to daunting or has flaws that make it unable to be used as intended.
- Something I can provide my users so they can give me feedback about the stability and ease of use so we can further refine it.

=> Basically, experimental features are moving targets until problems may be ironed out.

There are multiple reasons I think this should be flagged as experimental:

1. We are only testing our happy path with all provided frameworks. We completely skip more complex examples, e.g. loading data remotely, make use of angular async pipes, do not know what happens when using any angular directives or react effects...
2. I am still unsure if bigger, long running apps will run into memory leaks with this enabled. We need to [hold a reference](https://github.com/synergy-design-system/synergy-design-system/pull/733/files#diff-514de6b688d4b1723022324b8bc354aeece3a640bcb4fc155d0a4bb1d4bc077bR31) to all synergy elements that where rendered and have default settings attached. While this **should** work fine in theory, truth is it is by no means guaranteed that it also works in all supported frameworks (e.g. some of them may release the element after state updates or might not call the needed `disconnectedCallback`).

So basically for me this is the way to say to my users:

Look, you can do this and we would be happy to get back some feedback on this as we do not have enough experience in real world applications to make sure it runs as intended for your setup. Please give us feedback how it behaves and if you run into any troubles. When it does not work out, we warned we may remove this at any time. When its stable enough, you will be able to upgrade to an official api (backwards compatible, for sure!).

~~tl;dr~~~

I do not want to discourage usage, but with this statement raise the awareness that this is still a moving target and may receive various changes underneath - or the feature may even be removed again if it is not working out for everyone without stepping on the toes of people that think this is a stable API. So yeah, I want the users to be sure to understand this is not something that I would see 100% production ready.

Proposed workflow:

👍 When its deemed stable enough (e.g. we have enough positive feedback):

1. Copy `enableExperimentalSettingEmitEvents` to a new function `enableSettingEmitEvents`.
2. Adjust `enableExperimentalSettingEmitEvents` to call `enableSettingEmitEvents`. Print a `console.warn` before to make sure users are notified that they can now use the official api.
3. Remove `enableExperimentalSettingEmitEvents` in the next major release.

👎 When its deemed too unstable:

1. Remove the events code from decorator and settings function code.
2. Exchange `enableExperimentalSettingEmitEvents` for a function that just outputs a `console.warn` statement, highlighting that this feature did not make it into an official release.
3. Remove `enableExperimentalSettingEmitEvents` in the next major release

packages/docs/stories/GlobalSettings.mdx
schilchSICKAG: Question is where it should be otherwise? It is not a component, and our current stance is: If its something that is globally effecting the components, it will be taken to the top.

packages/docs/stories/GlobalSettings.mdx
schilchSICKAG: Examples for experimental stuff in the wild:

Lit exposes packages as moving targets until they are officially accepted as stable enough: https://lit.dev/docs/libraries/labs/ (e.g. `@lit-labs/react` -> `@lit/react`).

Also:

Experimental features in JavaScript packages are often new functionalities introduced for testing and feedback before being officially included in the package. Here are some examples:

React Concurrent Mode:

Description: React Concurrent Mode allows React applications to render UI in a more performant and responsive way.
Release: Introduced as experimental in React 16.6.
Documentation: Detailed in the React documentation with a clear note that it is experimental and may change in future releases.

Node.js ES Modules:

Description: ES Modules provide a standardized way to organize and import/export JavaScript code.
Release: Initially released as experimental in Node.js 12.
Documentation: Documented in Node.js documentation with warnings about its experimental nature and potential changes.

Vue.js Composition API:

Description: The Composition API is a set of additive, function-based APIs that allow flexible composition of component logic.
Release: Introduced as experimental in Vue 2.7 before officially being included in Vue 3.
Documentation: Provided in Vue.js documentation with notes on its experimental status.
These features are typically released in minor or patch versions to allow developers to test them without affecting the stability of the main package. Documentation for experimental features usually includes:

Clear labels indicating the feature's experimental status.
Instructions on how to enable and use the feature.
Warnings about potential changes or removal in future versions.
Feedback channels for developers to report issues or suggest improvements.
Experimental features are often marked with specific tags or flags that need to be enabled in the package configuration to use them.

schilchSICKAG: To prove the point: https://github.com/synergy-design-system/synergy-design-system/blob/feat/674-global-sizing-version-extend/packages/components/src/internal/defaultSettings/functions.ts#L98 creates a memory leak when you first enable the event support and disable it again. The [global event map](https://github.com/synergy-design-system/synergy-design-system/blob/feat/674-global-sizing-version-extend/packages/components/src/internal/defaultSettings/functions.ts#L26) will not receive updates anymore, leading to the already destroyed instance to stay in there. Will fix this asap, but there may be other problems similar to this lurking around :(

packages/_private/angular-demo/src/shared.ts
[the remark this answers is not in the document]
schilchSICKAG: Fixed

packages/_private/react-demo/src/shared.ts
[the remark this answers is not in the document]
schilchSICKAG: Fixed

packages/_private/vanilla-demo/src/utils.ts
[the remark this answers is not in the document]
schilchSICKAG: Fixed

packages/_private/vue-demo/src/shared.ts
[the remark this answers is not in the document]
schilchSICKAG: Fixed

packages/docs/stories/GlobalSettings.mdx
schilchSICKAG: Understand and agree :). Will fix this asap

schilchSICKAG: Moved :)

packages/docs/stories/packages/components/7-GlobalSettings.mdx
```diff
@@ -0,0 +1,136 @@
...
+        return (
+          <tr key={`${c}--details`}>
+            <td><strong>{c}</strong></td>
+            <td>
+              {Object
+                .entries(settings)
+                .sort(([a], [b]) => a.localeCompare(b))
+                .map(([key, value]) => (
+                  <>
+                    <strong key={`${c}--${key}--key`}>{key}</strong>: {value}<br />
```
reichst1SICKAG, line 127: The naming of the component list don't match our standard
Is: SynCamelCaseName
Should: syn-kekab-case

packages/docs/stories/packages/components/7-GlobalSettings.mdx
```diff
@@ -0,0 +1,136 @@
+import { Meta } from "@storybook/blocks";
+import { extractDefaultSettingsForElement, INITIAL_DEFAULT_SETTINGS } from '../../../../components/src/synergy';
+
+<Meta title="Packages/Components/Configuring global settings" />
+
+# Overriding global default settings for components
+
+Most of synergies components provide a way to configure their appearance and behavior through a set of properties.
```
reichst1SICKAG, line 8: "Most Synergy components ..."

schilchSICKAG: Fixed

packages/docs/stories/packages/components/7-GlobalSettings.mdx
schilchSICKAG: I adjusted the table to include the tagname, as well as the needed base ClassName and also added usage copyable usage examples. Please have another look and tell me if this version is now better.

packages/docs/stories/packages/components/7-GlobalSettings.mdx
```diff
@@ -0,0 +1,175 @@
...
+Most Synergy components provide a way to configure their appearance and behavior through a set of properties.
+However, it can be cumbersome to set the same properties for all components in your application.
+
+Therefore, Synergy includes a low level API that allows you to set global defaults for the rendered components.
+This can be used to set the default size or other global settings that should be applied to all rendered components.
+
+## Setting global defaults
+
+The settings can be set on a global or component level via `setGlobalDefaultSettings` and `setDefaultSettingsForElement` respectively.
+
```
kirchsuSICKAG, line 18: I would add the hint, that it is important, where this functions are called. That they need to be called at an early point, where the synergy component have not been rendered yet ( e.g. in `main.ts(x)`), otherwise the global resizing will not work

packages/components/src/synergy.ts
```diff
@@ -59,5 +59,8 @@ export * from './utilities/form.js';
 
 export * from './components/combobox/option-renderer.js';
 
+// Global settings
+export * from './internal/defaultSettings/index.js';
```
kirchsuSICKAG, line 63: Maybe we can move the files under the "utilities" subfolder instead of the "internal". Because in my understanding, we should not export our "internals".
Otherwise maybe the people do `import { setGlobalDefaultSettings } from '@synergy-design-system/components/internal/defaultSettings/index.js';
`

And I would also only export the thing the people really need, and not our internal functions, which we use

kirchsuSICKAG (changes requested):

packages/components/src/synergy.ts
schilchSICKAG: Hmm, basically this is [set via the exports field](https://github.com/synergy-design-system/synergy-design-system/blob/feat/674-global-sizing-version-extend/packages/components/package.json#L27) in `package.json`, so a relative import would not be possible.

From this perspective, we really should move it to `utilities` to make it able to be included via `import * as defaultSettings from '@synergy-design-system/utilities/defaultSettings.js` as currently, all projects that load it from the main file will have the tree shackability problem :(.

Will move it for sure and adjust all paths.

packages/docs/stories/packages/components/7-GlobalSettings.mdx
schilchSICKAG: Good point. Originally, I thought this would just be a problem for the event system because old components are not registred, but you are right, calling it too late will lead to a broken UI. Maybe we could also add some debug statements, I am thinking of something like:

```typescript
// todo: Make more robust, just an example
let hasBeenCalledOnce = false;
let phaseWhereWeAllowToCallSettingsReached = false;

if (process.env.NODE_ENV !== 'production') {
  document.addEventListener('DOMContentLoaded', () => {
    phaseWhereWeAllowToCallSettingsReached = true;
  });
}

setDefaultSettingsForElements() {
  // ...
  if (!hasBeenCalledOnce && phaseWhereWeAllowToCallSettingsReached) {
    console.error('You have been calling default settings to late!');
  }
}
```

What do you think?

packages/components/src/synergy.ts
schilchSICKAG: Moved everything to `utilities`, works fine. I also export the "usable" parts in `index.js`, so we do not get any problems with the exports (at least not when the user does not forcefully import the internals in `functions.js`).

I thought about moving the internals back to internal and the externals to utilities, but for me it does not have any real world usage to do this. We should still export the decorator and types for people that may want to use the decorator in their own components, so with this in mind it makes sense to export the other stuff to, even if obusfacated :

packages/docs/stories/packages/components/7-GlobalSettings.mdx
schilchSICKAG: I opted not to make it any harder on us and not add the code sample above. Adjusted the documentation, though.

kirchsuSICKAG (changes requested):

packages/components/src/synergy.ts
kirchsuSICKAG: is there any use case where people really need the decorator?

packages/_private/angular-demo/src/main.ts
```diff
@@ -1,7 +1,9 @@
+import { enableExperimentalSettingEmitEvents } from '@synergy-design-system/components';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 
 import { AppModule } from './app.module';
 
+enableExperimentalSettingEmitEvents();
```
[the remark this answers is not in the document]
schilchSICKAG: Ah, its just because its not necessary, will update to make all demos align again.

packages/components/src/utilities/defaultSettings/index.ts
```diff
@@ -0,0 +1,10 @@
+export type * from './base.js';
+export { defaultSettings, INITIAL_DEFAULT_SETTINGS } from './base.js';
+export * from './decorator.js';
```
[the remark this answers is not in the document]
schilchSICKAG: They are only for the case when you want to reset the settings to their original meaning. I could also just export a function for this, too. Would make more sense I guess.

.vscode/settings.json
```diff
@@ -195,6 +195,7 @@
     "packages/components/src/internal/active-elements.ts": true,
     "packages/components/src/internal/animate.ts": true,
     "packages/components/src/internal/default-value.ts": true,
+    "packages/components/src/internal/defaultSettings/base.ts": true,
```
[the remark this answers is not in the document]
schilchSICKAG: Done

packages/_private/angular-demo/src/main.ts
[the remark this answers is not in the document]
schilchSICKAG: Removed

packages/components/src/utilities/defaultSettings/index.ts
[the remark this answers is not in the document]
schilchSICKAG: Export and docs for it added.

packages/components/src/synergy.ts
schilchSICKAG: Not **yet**. However, if you want to write a component using Synergy, it would be beneficial if it the developer could automatically benefit from it to allow for example the hypotetical components size property.

kirchsuSICKAG (approved):

PR #739, opened 2025-01-17
fix: 🐛 syn-validate: Validation message not removed when the slotted input is set to readonly or disabled

### LINKED ISSUE #717: fix: 🐛 syn-validate: Validation message not removed when the slotted input is set to readonly or disabled

# Version

The bug occurred with the following version of the component library: 2.15.0

# Component

The issue arises when using the following component(s): syn-validate

# Environment

# Description

When using a `<syn-validate>` element that has a `<syn-input>` with `readonly`, the validation is not triggered when the `readonly` flag is removed from the DOM. `<syn-validate>` should be smart enough to trigger a new validation run when `readonly` or `disabled` are set or unset.

# Acceptance criteria

- [ ] `<syn-validate>` takes the `readonly` and `disabled` properties of form elements into account.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue with `<syn-validate>` which lead to a visualization problem when setting or removing `disabled` or `readonly` properties on the slotted input field. If rendered with one of those properties initially and eager set, the validation would not show an error message when the property was removed later on.

### 🎫 Issues

Closes #717

## 👩‍💻 Reviewer Notes

Code-wise, I had to use `Element.getAttribute` instead of simple checks like `element.disabled`. This is because Shoelace uses other props than regular html (e.g. `readOnly` instead of `readonly`). As attributes are case insensitive, this lead to easier checks.

I opted to setup the needed `MutationObserver` on `connectedCallback` on the element itself. This was done to make sure that the observer always listens to changed nodes and also picks up complete new ones, e.g. when rendering `<syn-validate>` with asynchronous content it will still work, whereas it would not have worked when setting it up with `this.getInput()` as a target. As I am using an attribute filter, this should have not too many performance overhead.

## 📑 Test Plan

Have a look at [the provided Codepen example](https://codepen.io/schilchSICKAG/pen/RNbBbVJ?editors=1110). It will load the input with `readonly` initially. When the readonly attribute is removed from the element by clicking the "Toggle readonly" button, it will show a red lining (as the input is correctly identified as invalid), but will not show the error message.

Try the same with the changes applied to this branch to see that it will now display the error message correctly.

## ✅ DoD

### DIFF
--- a/packages/components/src/components/validate/validate.component.ts
+++ b/packages/components/src/components/validate/validate.component.ts
@@ -1,4 +1,4 @@
-import type { CSSResultGroup } from 'lit';
+import type { CSSResultGroup, PropertyValues } from 'lit';
 import { html } from 'lit';
 import { property, queryAssignedElements, state } from 'lit/decorators.js';
 import componentStyles from '../../styles/component.styles.js';
@@ -38,6 +38,8 @@ export default class SynValidate extends SynergyElement {
   };
 
   controller = new AbortController();
+
+  observer: MutationObserver;
 
   @queryAssignedElements() private slottedChildren: HTMLElement[];
 
@@ -312,7 +314,8 @@ export default class SynValidate extends SynergyElement {
     }
   };
 
-  async firstUpdated() {
+  async firstUpdated(changedProperties: PropertyValues) {
+    super.firstUpdated(changedProperties);
     this.updateEvents();
 
     // Make sure to run validation on mount if eager is set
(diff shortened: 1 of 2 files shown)


kirchsuSICKAG:
@schilchSICKAG 
Just something that got my attention: 
having an input with validate and custom-validation message: 
--> toggling it to disabled, will lead to a disabled state  **WITH** red border of the input.
--> toggling it to readonly, will lead to a readonly state **WITHOUT** red border of the input.

Looks a bit inconsistent to me. But I don't know which would be the right behavior. Thinking about it I would even say the readonly should be red and the disabled not. But would also be not consistent.

Image

But would possibly be something for a new issue or do you think it fits here?

kirchsuSICKAG:
Unfortunally I found a bug (which we already had before) 😕
Using the eager property with elements without validation on it (like `required` or `type="email"`, but only using the `custom-validation-message` do not work. The inline message is not shown.

See following code example and the corresponding image:

```html
<!-- Does not validate, because of the bug -->
<syn-validate eager custom-validation-message="hello" variant="inline">
  <syn-input label="Custom validation message"></syn-input>
</syn-validate>

<!-- Does not validate, because of the bug and as the validation with required is fulfilled -->
<syn-validate eager custom-validation-message="hello" variant="inline">
  <syn-input label="Custom validation message + NOT fulfilled validation" required value="sd"></syn-input>
</syn-validate>

<!-- Does validate, as the validation with required is fulfilled -->
<syn-validate eager custom-validation-message="hello" variant="inline">
  <syn-input required label="Custom validation message + fulfilled validation"></syn-input>
</syn-validate>
````

Image

Should I create a new issue for this?

schilchSICKAG:
> Unfortunally I found a bug (which we already had before) 😕 Using the eager property with elements without validation on it (like `required` or `type="email"`, but only using the `custom-validation-message` do not work. The inline message is not shown.
> 
> See following code example and the corresponding image:
> 
> ```
> 
> <syn-validate eager custom-validation-message="hello" variant="inline">
>   <syn-input label="Custom validation message"></syn-input>
> </syn-validate>
> 
> 
> <syn-validate eager custom-validation-message="hello" variant="inline">
>   <syn-input label="Custom validation message + NOT fulfilled validation" required value="sd"></syn-input>
> </syn-validate>
> 
> 
> <syn-validate eager custom-validation-message="hello" variant="inline">
>   <syn-input required label="Custom validation message + fulfilled validation"></syn-input>
> </syn-validate>
> ```
> 
> Image
> 
> Should I create a new issue for this?

Good catch. This was done in https://github.com/synergy-design-system/synergy-design-system/pull/745 :)

kirchsuSICKAG (approved):

PR #748, opened 2025-01-21
fix: 🐛 React: style and ref for react19 not working

### LINKED ISSUE #746: fix: 🐛 React: style and ref for react19 not working

# Version

The bug occurred with the following version of the component library: 2.16.0

# Component

The issue arises when using the following component(s): All

# Environment

# Description

Our new type wrappers work fine in react 19 projects. However, we have missed to include some properties or even misinterpreted types for working ones:

### React Ref:

The `ref` key is missing in our definition of `SynCustomElement`, which leads to errors when using something like:

```jsx
const MyConponent = () => {
  const cardRef = useRef<SynCard>(null);
  return (
    <syn-card ref={cardRef} />
  );
};
```

The property `key` has to be added. It should have a type of `HTMLElement | null`.

### Style property

Currently defined as `React.CSSStyleDeclaration`, this works fine for most use-cases. However, it does not work fine when using custom css properties like in the following example:

```jsx
const MyComponent = () => (
  <syn-button
    style={{
      '--syn-primary-color-300': 'black', // 👈 Won´t work with TS!
      backgroundColor: 'red', // Works fine
    }}
  >Button</syn-button>
);
```

# Acceptance criteria

- [ ] Ref Type is added
- [ ] Style type is adjusted to include css vars
- [ ] Demo is created in the demo project

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes two issues with our new type only react support. IT addresses the following issues:

- `ref` was not a supported property on the wrappers, making it impossible to use `useRef()` with TypeScript
- The generic `styles` property in React does not support css custom properties.

### 🎫 Issues

Closes #746

## 👩‍💻 Reviewer Notes

I was slightly tempered to only include our Synergy Tokens in the allowance list, which would have forced their usage. However, I think as we don´t quite know what other component libraries are used by our customers, I went with a dynamic key in `interface CSSProperties`.

## 📑 Test Plan

Just check out this branch and play around with the react demo. The following code will throw typescript errors:

```jsx
// DemoForm.tsx
import { useRef, useEffect } from 'react';
export const DemoForm = () => {
  // Try to change this to the following to check strong typings here :)
  // const firstInputRef = useRef<HTMLDivElement>(null);
  const firstInputRef = useRef<SynInput>(null);

  useEffect(() => {
    firstInputRef.current!.focus();
  }, []);

  return (
    <div>
      <syn-input
        style={{
          '--syn-color-primary-500': 'white',
        }}
        ref={firstInputRef}
      />
    </div>
  );
```

This should not lead to any errors after this fix.

Check out the branch and copy the example above into the react demo. On a current system without the fix applied, it should produce two errors: ref is an unknown property for `syn-input` and `--syn-color-primary-500` not allowed in StyeMap.

After the fix, everything should work fine and the input should receive focus.

## ✅ DoD

### DIFF
--- a/packages/components/scripts/jobs/react/intrinsic-elements.js
+++ b/packages/components/scripts/jobs/react/intrinsic-elements.js
@@ -31,7 +31,7 @@ export const runCreateIntrinsicElements = job('React: Creating react intrinsic e
 
   // List of imports needed for the generated types
   const imports = `
-    import type { DOMAttributes, RefObject } from 'react';
+    import type { CSSProperties, DOMAttributes, RefObject } from 'react';
     import type { ${synergyTypes.join(',')} } from '@synergy-design-system/components';
 
     /**
@@ -53,12 +53,13 @@ export const runCreateIntrinsicElements = job('React: Creating react intrinsic e
       SynElement extends HTMLElement,
       Events extends SynEventTuple[] = [],
     > = Partial<
-      SynElement &
+      Omit<SynElement, 'style'> &
       DOMAttributes<SynElement> &
       {
         children?: any;
         key?: any;
         ref?: RefObject<SynElement | null>;
+        style?: CSSProperties | undefined;
       } &
       SynEventMap<Events>
     >;
--- a/packages/components/scripts/jobs/react/intrinsic-elements.js
+++ b/packages/components/scripts/jobs/react/intrinsic-elements.js
@@ -31,7 +31,7 @@ export const runCreateIntrinsicElements = job('React: Creating react intrinsic e
 
   // List of imports needed for the generated types
   const imports = `
-    import type { DOMAttributes } from 'react';
+    import type { DOMAttributes, RefObject } from 'react';
     import type { ${synergyTypes.join(',')} } from '@synergy-design-system/components';
 
     /**
@@ -58,6 +58,7 @@ export const runCreateIntrinsicElements = job('React: Creating react intrinsic e
       {
         children?: any;
         key?: any;
+        ref?: RefObject<SynElement | null>;
       } &
       SynEventMap<Events>
     >;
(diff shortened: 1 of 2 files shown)


packages/components/scripts/jobs/react/intrinsic-elements.js
```diff
@@ -91,6 +92,13 @@ export const runCreateIntrinsicElements = job('React: Creating react intrinsic e
     ${componentExports.join('\n')}
 
     declare module 'react' {
+      // #746: Allow synergies custom properties as css variable names
+      // @see https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors
+      // @see https://stackoverflow.com/questions/65878880/typescript-template-literal-as-interface-key/65879197#65879197
+      interface CSSProperties {
+        [key: \`--\${string}\`]: string | undefined;
+      }
```
kirchsuSICKAG, line 100: Is this really enough for css Variables? I'm still getting a typescript error for css variables:
Bildschirmfoto 2025-01-22 um 13 13 52

kirchsuSICKAG (changes requested):

packages/components/scripts/jobs/react/intrinsic-elements.js
kirchsuSICKAG: I just saw that you extend the interface of `CSSProperties` but typescript complains about `CSSStyleDeclaration`. Maybe this is the problem?

kirchsuSICKAG (changes requested):

packages/components/scripts/jobs/react/intrinsic-elements.js
kirchsuSICKAG: I investigated it a bit further and it seems standard html element `style` properties do have another type (`CSSProperties`) than our web components (`CSSStyleDeclaration`):
react-style

kirchsuSICKAG (changes requested):

kirchsuSICKAG (approved):

PR #756, opened 2025-01-27
fix: 🐛 Remove warning message in console for syn-header and syn-combobox in dev mode

### LINKED ISSUE #753: fix: 🐛 syn-combobox and syn-header show warning in console in dev mode

# Version

The bug occurred with the following version of the component library: >= 2.16.0

# Component

The issue arises when using the following component(s): syn-combobox, syn-header

# Environment

# Description

When providing a `value` property initially with the `<syn-combobox>` or just using the syn-header following warning is logged in the console:

Image

# Acceptance criteria

- [x] syn-combobox with value should not log warning message
- [x] syn-header should not log warning message

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR removes the console warning messages of the syn-header and syn-combobox with value.
The warning appears, because changes are done in the `firstUpdated` callback, which lead to updated changes. By waiting for `this.updateComplete` before doing the changes, they are no longer triggering during the update process.

### 🎫 Issues

Closes #753 

## 👩‍💻 Reviewer Notes
Have a look at any syn-header story and check following code for the syn-combobox, that there is no longer this warning message in the console

```html
<syn-combobox value="opt1">
  <syn-option value="opt1">Option 1</syn-option>
  <syn-option value="opt2">Option 2</syn-option>
  <syn-option value="opt3">Option 3</syn-option>
</syn-combobox>
```

Unfortunately there is no way I can test this via unit test 😕
Or at least I have no idea how I could test it

## ✅ DoD

### DIFF
--- a/packages/components/src/components/combobox/combobox.component.ts
+++ b/packages/components/src/components/combobox/combobox.component.ts
@@ -236,7 +236,11 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
 
   firstUpdated() {
     // initially set the displayLabel if the value was set via property initially
-    this.displayLabel = this.value;
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.updateComplete.then(() => {
+      this.displayLabel = this.value;
+    });
     this.formControlController.updateValidity();
   }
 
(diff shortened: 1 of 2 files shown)


schilchSICKAG (approved):
You **could** technically test it, but I don´t think this is necessairy at all.

```ts
// Pseudocode, not working, but you get the idea...
// Some test describe/it stuff
const consoleSpy = sinon.stub(console, 'warn');
const renderMe = await render(html`<syn-side-nav>...`);

await expect(consoleSpy).to.not.have.been.called.with('Lit: Error MEssage here');
```

I would just let it be as its just in dev and I am happy enough we got rid of the message.

PR #759, opened 2025-01-30
chore: 🔧 Extend quality gate for framework wrapper

### LINKED ISSUE #555: chore: 🔧 Extend quality gate for framework wrapper
# Description

Currently only a subset of wrapper components are tested (contact form). We should aim to test all of them (function test, not screenshot comparison).

One solution could be to extend all demo packages.

# Rationale

This helps to find errors faster (like typescript errors) for a each component

# Acceptance criteria
- [ ] angular demo extended
- [ ] vue demo extended
- [ ] react demo extended
- [ ] vanilla demo extended
- [ ] dod contains a task to update demos for new components

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds all current components to our demo projects to make sure setups don´t break when importing components. It also includes two sample tests to get started for deeper testing of this page.

### 🎫 Issues

Closes #555 

## 👩‍💻 Reviewer Notes

Just have a look at the adjusted demos and the new `AllComponents` test. There are no code changes in components or otherwise, so should be fine for a quick release.

## 📑 Test Plan

Check out this branch, run a clean build and move to the e2e demo project. Run the tests via `pnpm start` or the new `pnpm start.ui` commands. Observe the outputs of the new tests, as well as the layout of the demo projects.

## ✅ DoD


schilchSICKAG:
We chose to not add any more tests, leaving it be for the ones already implemented in this PR. We may add regression tests for our components here in the future, but for now its feature complete, even if the ticket tells us otherwise :)

packages/_private/vue-demo/src/AllComponents.vue
```diff
@@ -0,0 +1,46 @@
+<script setup lang="ts">
+import { SynVueTab, SynVueTabGroup, SynVueTabPanel } from '@synergy-design-system/vue';
+import type { SynTabShowEvent } from '@synergy-design-system/components';
+import * as DemoImports from './AllComponentParts/index.js';
+
+const Demos = Object
+  .entries(DemoImports)
+  .map(([name, Component]) => [name.replace('Demo', ''), Component]);
```
kirchsuSICKAG, line 8: What is the reason, that you need to name the components with prefix "Demo" and here remove it again? In all other frameworks, you name it just without "Demo". Is this some vue specific thing?

packages/_private/vanilla-demo/vite.config.ts
```diff
@@ -70,6 +70,10 @@ export const withSpriteSheetGenerator = {
         'home',
         'contact_mail',
         'contact_emergency',
+        'grid_view',
+        'wallpaper',
+        'house',
+        'chat',
```
kirchsuSICKAG, line 76: The icons of the alerts are missing and are therefore not visible:
image

```suggestion
        'grid_view',
        'wallpaper',
        'house',
        'chat',
        'info',
        'check_circle',
        'settings',
        'warning',
        'error',
```

packages/_private/vue-demo/src/AllComponents.vue
schilchSICKAG: Yes, this is indeed Vue specific. Vue needs components to be named with camelcase, otherwise it will show Error Messages :(

packages/_private/vanilla-demo/vite.config.ts
schilchSICKAG: I see, will adjust this

schilchSICKAG: Fixed

kirchsuSICKAG (approved):

PR #764, opened 2025-01-31
fix: 🐛 syn-alert: Not able to import syn-alert when using SSR

### LINKED ISSUE #762: fix: 🐛 syn-alert: Not able to import syn-alert when using SSR

# Version

The bug occurred with the following version of the component library: syn-alert, syn-validate

# Component

The issue arises when using the following component(s): All

# Environment

The issue occurred in the following browser(s): node.js (SSR)

# Description

When rendering Synergy on the server side, a side effect in [`alert.component.ts`](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/src/components/alert/alert.component.ts#L25) will fail on SSR targets as `document.createElement` is always called, resulting in an error.

> ⚠️ As `<syn-validate>´also uses `<syn-alert>` under the hood, it is also not able to be used in SSR environments until this bug is fixed.

# Acceptance criteria

- [ ] The creation of the toast stack is only done on client targets.
- [ ] `<syn-alert>` works in our [Angular SSR Demo](https://github.com/synergy-design-system/ssr-demos/blob/main/packages/angular/src/AllComponentParts/index.ts#L2)
- [ ] `<syn-validate>` works in our [Angular SSR Demo](https://github.com/synergy-design-system/ssr-demos/blob/main/packages/angular/src/AllComponentParts/index.ts#L33)
- [ ] Provide PR for Shoelace.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds a small adjustment to `<syn-alert>` to make it possible to use it in ssr environments. When importing the component, it originally outright created the `toastStack`, leading to break the import in server environments.

I opted to make it available as `static private` members of `SynAlert` itself, leading to the same result but making sure it will not crash when imported and creating the toast stack only when the toast method is called.

### 🎫 Issues

Closes #762 

## 👩‍💻 Reviewer Notes

Just have a look at the code changes, they should be pretty self explanationary.

## 📑 Test Plan

Play around with your local version and [enable the imports for Alert and Validate](https://github.com/synergy-design-system/ssr-demos/blob/main/packages/angular/src/AllComponentParts/index.ts#L2) in our angular ssr demo application.

> Tip: This is easiest when using the build in synergy and copying the contents of `alert.component.js` that was created in the dist directory, starting from the comment `// src/components/alert/alert.component.ts`.

It should work fine now without issuing an error message when loading the Angular SSR demo.

## ✅ DoD

### DIFF
--- a/packages/components/scripts/vendorism/custom/alert.vendorism.js
+++ b/packages/components/scripts/vendorism/custom/alert.vendorism.js
@@ -1,4 +1,4 @@
-import { replaceSections } from '../replace-section.js';
+import { addSectionAfter, replaceSections } from '../replace-section.js';
 import { removeSections } from '../remove-section.js';
 
 const FILES_TO_TRANSFORM = [
@@ -31,7 +31,7 @@ const transformStyles = (path, originalContent) => {
 };
 
 const transformComponent = (path, originalContent) => {
-  const content = replaceSections([
+  const contentWithoutCountdown = replaceSections([
     // Begin remove countdown
     [
       "@property({ type: String, reflect: true }) countdown?: 'rtl' | 'ltr';",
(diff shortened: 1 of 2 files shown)


kirchsuSICKAG (approved):

schilchSICKAG:
I will close this PR as it will automatically land with the update to Shoelace 2.20.0 done in https://github.com/synergy-design-system/synergy-design-system/pull/769

PR #765, opened 2025-01-31
fix: 🐛 syn-tab-panel: Not able to import syn-tab-panel when using SSR 

### LINKED ISSUE #763: fix: 🐛 syn-tab-panel: Not able to import syn-alert when using SSR

# Version

The bug occurred with the following version of the component library: All

# Component

The issue arises when using the following component(s): syn-tab-panel

# Environment

The issue occurred in the following browser(s): node.js (SSR)

# Description

When rendering Synergy on the server side, a side effect in [`scrollend-polyfill.ts`](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/src/internal/scrollend-polyfill.ts#L37) will fail on SSR targets as ` window` is always queried for support, resulting in an error.

# Acceptance criteria

- [ ] Remove the polyfill if it is not needed on latest major versions.
- [ ] If still needed, guard the call to the window call so it returns `true` on SSR (or does not run at all).
- [ ] `<syn-tab-panel>` works in our [Angular SSR Demo](https://github.com/synergy-design-system/ssr-demos/blob/main/packages/angular/src/AllComponentParts/index.ts#L29)
- [ ] Provide PR for Shoelace.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adjusts the `scrollend-polyfill` so it only runs in the browser. With this change `<syn-tab-group>` can be rendered without errors in ssr environments like Angular.

### 🎫 Issues

Closes #763 

## 👩‍💻 Reviewer Notes

This is just a small change. Unfortunately, we cannot remove the polyfill as Safari has no support for `scrollend` support. This feature is used in the Arrows left and right in the tab group that acts as a previous/before navigation on the front and the back of the items. The event is used to disable the "Next Arrow" when the user has scrolled to the last item horizontally.

## 📑 Test Plan

Until this is released, check out the ssr Demo and locate `packages/angular/node_modules/@synergy-design-system/components/dist/components/tab-group/tab-group.component.js`. Locate the chunk that imports `SynTabGroup` (name may vary) and replace the section that starts with `var debounce` until `// src/components/tab-group/tab-group.component.ts` with the code from the disted bundle. The only difference should be an IIFE that is called and checks if window is defined.

Clean the `.angular` directory in the SSR demo, enable the Tabs story by including it in `AllComponents/index.ts` and run. You should not receive any errors. When scrolling to the last element (maybe you have to add one or two to the component), Safari should still disable the right arrow button after a short timeout.

## ✅ DoD


kirchsuSICKAG (approved):

schilchSICKAG:
I will close this PR as it will automatically land with the update to Shoelace 2.20.0 done in https://github.com/synergy-design-system/synergy-design-system/pull/769

PR #801, opened 2025-03-10
fix: 🐛 setGlobalDefaultSettings does overwrite explicitly set default property values with new default size

### LINKED ISSUE #799: fix: 🐛 setGlobalDefaultSettings does overwrite explicitly set size="medium" with new default size

# Version

The bug occurred with the following version of the component library: 2.22.5

# Component

The issue arises when using the following component(s): setGlobalDefaultSettings

# Environment

# Description

The mechanism of the `setGlobalDefaultSettings` does honor explicitly set "size" property on element and does not overwrite them with the new default. But this behavior is only applies to `size="small"` and `size="large"` and is ignored for `size="medium".

Image

See following codepen for reproduction:

https://codepen.io/kirchsuSICKAG/pen/pvowdgV

# Acceptance criteria

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR fixes the issue, that explicitly set properties with their default value (e.g. size="medium" for the syn-button `<syn-button size="medium" />`) is not honored, but overwritten with the new default value.

### 🎫 Issues
Closes #799 

## 👩‍💻 Reviewer Notes

I just found out that the `requestUpdate` callback is called for explicitly set properties, even if the property is set to its default value. If a property is not explicitly set, this callback is not called. This behaves the same for plain webcomponents and for all 3 web frameworks.
Via this mechanism we can find out, if just the default from internal was used or if it was explicitly set.

## 📑 Test Plan

- check out for plain web component and all 3 web frameworks, if the behavior is now correct!

- example code angular: 

```html
<syn-button>default</syn-button>
<syn-button size="small">small</syn-button>
<syn-button size="medium">medium</syn-button>
<syn-button size="large">large</syn-button>
```

- example code vue:

```html
  <SynVueButton>default</SynVueButton>
  <SynVueButton size="small">small</SynVueButton>
  <SynVueButton size="medium">medium</SynVueButton>
  <SynVueButton size="large">large</SynVueButton>
```

- example code react:

```html
    {/* New react way */}
    <syn-button>default</syn-button>
    <syn-button size="small">small</syn-button>
    <syn-button size="medium">medium</syn-button>
    <syn-button size="large">large</syn-button>

    {/* Old react way */}
    <SynButton>default</SynButton>
    <SynButton size="small">small</SynButton>
    <SynButton size="medium">medium</SynButton>
    <SynButton size="large">large</SynButton>
```

## ✅ DoD

### DIFF
--- a/packages/components/src/utilities/defaultSettings/decorator.ts
+++ b/packages/components/src/utilities/defaultSettings/decorator.ts
@@ -21,13 +21,13 @@ type Constructor<T = object> = new (...args: any[]) => T;
  */
 export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
   return <T extends Constructor<SynergyElement>>(Proto: T): T => class extends Proto {
-    private _globalSettingsSetupComplete = false;
+    #globalSettingsSetupComplete = false;
 
     #initialGlobalSettingEmptyProperties = new Map<string, unknown>();
 
-    private _initialProperties: Array<string> = [];
+    #initialProperties: Array<string> = [];
 
-    private _systemDefaultSettings: Record<string, unknown>;
+    #systemDefaultSettings: Record<string, unknown>;
 
     private _isInitialized = false;
 
@@ -41,7 +41,7 @@ export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
       super(...args);
       this._isInitialized = true;
-      this._systemDefaultSettings = extractDefaultSettingsForElement(name, 'initial');
+      this.#systemDefaultSettings = extractDefaultSettingsForElement(name, 'initial');
     }
 
     /**
@@ -74,30 +74,30 @@ export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
         // element initialization is not ready
         !this._isInitialized
         // setup is already complete
-        || this._globalSettingsSetupComplete
+        || this.#globalSettingsSetupComplete
         || !propName
         // skip properties, which do not belong to global settings
-        || !(propName in this._systemDefaultSettings)
+        || !(propName in this.#systemDefaultSettings)
         // skip properties, which are already in the initial properties
-        || this._initialProperties?.includes(propName as string)) {
+        || this.#initialProperties?.includes(propName as string)) {
         return;
       }
 
       // Only explicit set properties / attributes on the element will trigger a requestUpdate
       // (e.g. <syn-button size="medium" /> ). If the default is used (e.g. <syn-button />),
       // the property will not appear in the _initialProperties array.
-      this._initialProperties.push(propName as string);
+      this.#initialProperties.push(propName as string);
     }
 
     protected willUpdate(changedProps: PropertyValues): void {
       super.willUpdate(changedProps);
 
       // Skip after the first run
-      if (this._globalSettingsSetupComplete) {
+      if (this.#globalSettingsSetupComplete) {
         return;
       }
 
-      this._globalSettingsSetupComplete = true;
+      this.#globalSettingsSetupComplete = true;
 
       // Get the default settings
       const defaults = extractDefaultSettingsForElement(name);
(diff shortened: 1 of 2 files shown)


packages/components/src/utilities/defaultSettings/decorator.ts
```diff
@@ -21,10 +21,12 @@ type Constructor<T = object> = new (...args: any[]) => T;
  */
 export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
   return <T extends Constructor<SynergyElement>>(Proto: T): T => class extends Proto {
-    #globalSettingsSetupComplete = false;
```
schilchSICKAG, line 24: Is there a specific reason this was exchanged from a "real" private variable to a ts one? Initially, I did not want to bleed the implementation to the outside

packages/components/src/utilities/defaultSettings/decorator.ts
```diff
@@ -52,15 +54,30 @@ export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
       removeGlobalEventNotification(this);
     }
 
+    requestUpdate(propName?: PropertyKey, oldValue?: unknown, options?: PropertyDeclaration) {
+      super.requestUpdate(propName, oldValue, options);
+
+      if (this._globalSettingsSetupComplete || this._initialProperties === undefined) {
```
schilchSICKAG, line 60: `this._initialProperties` is never undefined as we provide an empty array for constructing it. Could be removed. Elements that do not use this decorator won´t run the logic anyways.

packages/components/src/utilities/defaultSettings/decorator.ts
```diff
@@ -52,15 +54,30 @@ export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
...
+        return;
+      }
+
+      if (!this._initialProperties.includes(propName as string) && oldValue !== undefined) {
```
schilchSICKAG, line 64: I really should have RTFM. Very cice solution! Only thing that I am seeing is that all initial properties are now collected. Do you think it would make sense to filter them via `extractDefaultSettingsForElement`?

packages/components/src/utilities/defaultSettings/defaultSettings.test.ts
```diff
@@ -41,6 +41,12 @@ describe('GlobalSettings', () => {
       expect(button.size).to.equal('small');
     });
 
+    it('should render the button with its original size when a size=medium property is provided', async () => {
```
schilchSICKAG, line 44: ❤️

schilchSICKAG (approved):
Thats just great! So simple and yet so powerful. Should have been in the first version. Thanks for tackling this! Just some (very) minor stuff and I think we are good to go. If you don´t see value in my comments, just go ahead

packages/components/src/utilities/defaultSettings/decorator.ts
schilchSICKAG: Ah, makes sense. Thanks for the explanation.

packages/components/src/utilities/defaultSettings/decorator.ts
kirchsuSICKAG: yeah this could be a good idea 👍

packages/components/src/utilities/defaultSettings/decorator.ts
kirchsuSICKAG: For some reason I got following error message 
image

I think this is because the `requestUpdate` method is called before e.g. the SynButton constructor and before the enableDefaultSettings Decorator constructor is called. With the changes I did now, it seems like it is enough to only do the _isInitialized as fake private member

packages/components/src/utilities/defaultSettings/decorator.ts
```diff
@@ -25,11 +25,25 @@ export function enableDefaultSettings(name: ComponentNamesWithDefaultValues) {
...
+
+    private _isInitialized = false;
+
     // eslint-disable-next-line class-methods-use-this
     get __originalDecoratedClassName() {
       return name;
     }
 
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    constructor(...args: any[]) {
```
schilchSICKAG, line 40: Thats just a thing of taste and you don´t have to do this, but I think you could also use `...args: unknown[]` and get rid of the eslint disable here.

schilchSICKAG (approved):

PR #803, opened 2025-03-11
feat: ✨ Allow to override autofill styles for syn-input

### LINKED ISSUE #800: feat: ✨ Allow to override autofill styles for syn-input

# Component

I suggest to add the feature to the following component: syn-input

# Story

As a developer using the `<syn-input>` in a `<form>`, I want to make sure that I am able to override the internal styles used for the [autofill](https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill) functionality.

This must be provided for example via new css variables as interface (e.g. `--syn-input-autofill-text-color` and `--syn-input-autofill-shadow`). Unfortunately, this does not seem work out of the box with css parts.

# Acceptance criteria

- [ ] Naming for the new needed css variables is decided.
- [ ] The new tokens are added to [the Synergy custom overrides](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/src/components/input/input.custom.styles.ts), with their baseline set to [Shoelace defaults](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/src/components/input/input.styles.ts#L120)
- [ ] Optional: Document the new tokens in the component using two new `@cssproperty` statements.
- [ ] To be decided: Add a story to show this feature (Sidenode: I do not think this is needed, but we should discuss in the team).

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the possibility to adjust the default coloring of `<syn-input>` when used with user data autofilling enabled.

### 🎫 Issues

Closes #800 

## 👩‍💻 Reviewer Notes

Autofill has some common requirements:

1. You need to have it wrapped in a form
2. You have to have at least two fields for autocomplete to work (e.g. E-Mail and Name, Street...)
3. Your browser must be setup to automatically add your contact information (default for Safari, configurable for the other browsers)

## 📑 Test Plan

Start Storybook and add the following story to `input.stories.ts`:

```typescript
export const FormDemo: Story = {
  render: () => html`
    <form method="post">
      <syn-input name="name" type="text" label="Name"></syn-input>
      <syn-input name="street" type="text" label="Street"></syn-input>
      <syn-input name="email" type="email" label="E-Mail"></syn-input>
      <syn-button>Submit me</syn-button>
    </form>
    <style>
    syn-input {
      --syn-input-autofill-shadow: none;
      --syn-input-autofill-readonly-shadow: none;
      --syn-input-autofill-text-fill-color: red;
      --syn-input-autofill-caret-color: yellow;
    }
    </style>
  `,
};
```

This will add a sample demo form that has already overridden the defaults to none for the border and red for the text. Open in your browser. The autofill symbol should be visible. Let your contact data autofill and notice the text is red and the background yellow (which is the default and the reason for the shadow that is inserted!). Add the `readonly` attribute to the Street field and notice how it changes when you adjust the variable.

Please also read the new documentation in the docs that describes the tokens that where used.

## ✅ DoD

### DIFF
--- a/packages/components/scripts/vendorism/custom/input.vendorism.js
+++ b/packages/components/scripts/vendorism/custom/input.vendorism.js
@@ -229,18 +229,6 @@ const transformStyles = (path, originalContent) => {
   content = content.replaceAll('.input--no-spin-buttons ', '');
 
   // #800: Make sure we allow to override autofill vis css variables
-  content = addSectionAfter(
-    content,
-    `:host {
-    display: block;`,
-    `
-    --syn-input-autofill-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-background-color-hover) inset;
-    --syn-input-autofill-readonly-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-readonly-background-color) inset;
-    --syn-input-autofill-text-fill-color: var(--syn-color-primary-500);
-    --syn-input-autofill-caret-color: var(--syn-input-color);
-    `,
-  );
-
   content = replaceSections([
     [
       'box-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-background-color-hover) inset !important;',
@@ -259,6 +247,18 @@ const transformStyles = (path, originalContent) => {
       'box-shadow: var(--syn-input-autofill-readonly-shadow) !important;',
     ],
   ], content);
+
+  content = addSectionAfter(
+    content,
+    `:host {
+    display: block;`,
+    `
+    --syn-input-autofill-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-background-color-hover) inset;
+    --syn-input-autofill-readonly-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-readonly-background-color) inset;
+    --syn-input-autofill-text-fill-color: var(--syn-color-primary-500);
+    --syn-input-autofill-caret-color: var(--syn-input-color);
+    `,
+  );
 
   return {
     content,
--- a/packages/components/scripts/vendorism/custom/input.vendorism.js
+++ b/packages/components/scripts/vendorism/custom/input.vendorism.js
@@ -14,9 +14,18 @@ const FILES_TO_TRANSFORM = [
  * @returns
  */
 const transformComponent = (path, originalContent) => {
+  let content = removeSections([
+    ['/** Draws a filled', 'filled = false;'],
+    ['/** Draws a pill-style', 'pill = false;'],
+    ["'input--pill':", ','],
+  ], originalContent);
+
+  // Replace filled with readonly
+  content = content.replaceAll('filled', 'readonly');
+
   // #800: Add documentation for the new css variable tokens
-  let content = addSectionBefore(
-    originalContent,
+  content = addSectionBefore(
+    content,
     ` */
 export`,
     ` *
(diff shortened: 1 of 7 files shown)


packages/components/src/components/input/input.styles.ts
```diff
@@ -12,6 +12,11 @@ export default css`
 	/* stylelint-disable */
   :host {
     display: block;
+
+    --syn-input-autofill-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-background-color-hover) inset;
+    --syn-input-autofill-readonly-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-readonly-background-color) inset;
+    --syn-input-autofill-text-fill-color: var(--syn-color-primary-500);
+    --syn-input-autofill-caret-color: var(--syn-input-autofill-caret-color);
```
kirchsuSICKAG, line 19: Currently the caret-color does not have a color, if the user does not set the `--syn-input-autofill-caret-color` variable.

The value of the variable should be the same as before the changes

```suggestion
    --syn-input-autofill-caret-color: var(--syn-input-color);
```

packages/components/src/components/input/input.component.ts
```diff
@@ -69,6 +69,11 @@ import { enableDefaultSettings } from '../../utilities/defaultSettings/decorator
  * @csspart decrement-number-stepper - The decrement number stepper button.
  * @csspart increment-number-stepper - The increment number stepper button.
  * @csspart divider - The divider between the increment and decrement number stepper buttons.
+ *
+ * @cssproperty --syn-input-autofill-shadow - The shadow to apply when the input is autoreadonly.
+ * @cssproperty --syn-input-autofill-readonly-shadow - The shadow to apply when the input is readonly and autoreadonly.
+ * @cssproperty --syn-input-autofill-text-fill-color - The text fill color to apply when the input is autoreadonly.
+ * @cssproperty --syn-input-autofill-caret-color - The caret color to apply when the input is autoreadonly.
```
kirchsuSICKAG, line 76: For me the "autoreadonly" sounds strange. Is this the general term for the autofill?

kirchsuSICKAG (changes requested):

packages/components/src/components/input/input.component.ts
schilchSICKAG: Not at all :). My last commit was not taken into account, which was also the reason for the wrong token in styles.

packages/components/scripts/vendorism/custom/input.vendorism.js
```diff
@@ -214,6 +228,38 @@ const transformStyles = (path, originalContent) => {
...
+  // #800: Make sure we allow to override autofill vis css variables
+  content = addSectionAfter(
+    content,
+    `:host {
+    display: block;`,
+    `
+    --syn-input-autofill-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-background-color-hover) inset;
+    --syn-input-autofill-readonly-shadow: 0 0 0 var(--syn-input-height-large) var(--syn-input-readonly-background-color) inset;
+    --syn-input-autofill-text-fill-color: var(--syn-color-primary-500);
+    --syn-input-autofill-caret-color: var(--syn-input-color);
```
kirchsuSICKAG, line 240: I'm not sure where this is happening, but it seems our vendorism overwrites the var(--syn-input-color) at some point to var(--syn-input-autofill-caret-color); 😅

packages/components/src/components/input/input.styles.ts
kirchsuSICKAG: It seems like the vendorism overwrites the var(--syn-input-color);  with var(--syn-input-autofill-caret-color); (see my other comment in the input.vendorism.ts)

kirchsuSICKAG (changes requested):

packages/components/src/components/input/input.component.ts
schilchSICKAG: Fixed

packages/components/src/components/input/input.styles.ts
schilchSICKAG: Fixed in latest

kirchsuSICKAG (approved):

PR #816, opened 2025-03-19
fix: 🐛 Combobox should always show the text content of the option instead of the value

### LINKED ISSUE #797: fix: 🐛 setting the value property of the combobox results in the displayed option value instead of the text content

# Version

The bug occurred with the following version of the component library: 2.22.5

# Component

The issue arises when using the following component(s): syn-combobox

# Environment

# Description
Setting the `value` property of the syn-combobox leads to displaying the "value" of the selected syn-option instead of the text content. If selecting an option of the combobox via mouse or keyboard, the text content is shown. The behavior should be the same for both cases

https://codepen.io/kirchsuSICKAG/pen/ByaWMKL?editors=1110

See following image of the codepen. The first combobox is selected via the `value` property of the combobox, the second via selection with the mouse

Image

# Acceptance criteria

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR fixes the problem, that if the combobox value was set via property or attribute, the value of the option was shown instead of the text content.

### 🎫 Issues
Closes #797 

## 📑 Test Plan
Use following story code, to check out everything is now working fine if
 1. setting the value via attribute initially
 2. setting a value, which option is initially not available, but added dynamically
 3. setting a value, programmatically after some time

```js
export const ValueViaProperty: Story = {
  parameters: {
    docs: {
      description: {
        story: generateStoryDescription('combobox', 'label'),
      },
    },
  },
  render: () => html`

    <syn-combobox id="initial" value="option-3">
      <syn-option value="option-1">Option 1</syn-option>
      <syn-option value="option-2">Option 2</syn-option>
      <syn-option value="option-3">Option 3</syn-option>
    </syn-combobox>

    <syn-combobox id="dynamic-option" value="option-4">
      <syn-option value="option-1">Option 1</syn-option>
      <syn-option value="option-2">Option 2</syn-option>
      <syn-option value="option-3">Option 3</syn-option>
    </syn-combobox>

    <syn-combobox id="dynamic-value" value="option-1">
      <syn-option value="option-1">Option 1</syn-option>
      <syn-option value="option-2">Option 2</syn-option>
      <syn-option value="option-3">Option 3</syn-option>
    </syn-combobox>

    <script type="module">
      const dynamicOptionCombo = document.querySelector('#dynamic-option'); 
      const option = document.createElement('syn-option');
      option.value = 'option-4';
      option.textContent = 'Option 4';
        
      const dynamicValueCombo = document.querySelector('#dynamic-value');  

      setTimeout(() => {
        dynamicOptionCombo.appendChild(option);
        dynamicValueCombo.value = 'option-2'
      }, 2000);
    </script>
  `,
};
```

## ✅ DoD


packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -819,6 +821,17 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
 
       this.createComboboxOptionsFromQuery(this.value);
 
+      // check if the value has a corresponding option via value or text content
```
schilchSICKAG, line 824: This seems to be a copy&paste from the valueChange decorator. What do you think about combining the two into a private method?

schilchSICKAG (changes requested):
Other than the copy&paste of the logic I have nothing negative to add.

On a side-note, we already began [adding regression tests](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/_private/e2e-demo-test/src/AllComponents.spec.ts#L57) to our e2e version. Maybe we can add one for this case, too? This would make sure that it also works correctly in all other frameworks forever after :)

kirchsuSICKAG:
@schilchSICKAG 
I just saw, that if the value was set via option.value, the list of the combobox would not open. Therefore I adapted the filter function of the combobox, to also check for the value of the option if the textContent does not include the queryString.

You can use the first combobox code example to check this out.
Before:

combobox_before

After:

combobox_after

schilchSICKAG (approved):

Issue #747, opened 2025-01-21
fix: 🐛 closable syn-alert can be closed on whole vertical area without visual indication


kirchsuSICKAG:
# Checklist (for the issue creator)

- [ ] I have verified the bug exists in the latest version of `@synergy-design-system/components` (e.g. via `pnpm install @synergy-design-system/components`).
- [ ] I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...).
- [ ] I would be willing to provide a PR for this issue.

# Version

The bug occurred with the following version of the component library: 2.18.6

# Component

The issue arises when using the following component(s): syn-alert

# Environment

The issue occurred with the following operating system(s):
The issue occurred in the following browser(s):

# Description

The clickable close area of a closable syn-alert is the whole vertical area of the alert.
There is nothing that suggests that it can be closed in the area under the close icon (hover color on icon), but it will be closed none the less, when clicking there. See gif:

![Image](https://github.com/user-attachments/assets/5ad671a3-c190-4847-833f-c2c030e558d1)

Can be fixed with following css:

```
<style>
  syn-alert::part(close-button){
    align-self: start;
  }
 </style> 
```

# Acceptance criteria

- [ ] All DoD items successfully completed
- [ ] Mirror issue to Shoelace

- [ ] either do a hover effect on the whole closable click area OR (recommended) make only the button area clickable

PR #830, opened 2025-03-28
fix: 🐛  Upgrade packages to latest versions

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR updates our dependencies and dev dependencies to the latest versions.
It also updates Shoelace to version 2.20.1.

### 🎫 Issues
Closes #258 

### NOTE
Unfortunately I needed to set also the webkit and chrome browsers to "concurrency: 1". Otherwise they would fail in our CI. Locally I have no problems here.. Seems to be related to the issue, why firefox was set to concurrency:1 (https://github.com/modernweb-dev/web/issues/2374).  

### Changes from the shoelace update:
- Fixed a bug that prevented <sl-tab-group> to be activated properly when rendered in another <sl-tab-group> [#2367](https://github.com/shoelace-style/shoelace/issues/2367)
    --> we already fixed this on our side and just needed to remove our custom vendorism
- Fixed a bug that in <sl-dropdown> that prevented tab from working properly in some cases [#2371](https://github.com/shoelace-style/shoelace/issues/2371)
- Fixed the guard on popover to allow virtual elements [#2399](https://github.com/shoelace-style/shoelace/issues/2399)
- Fixed the close button in <sl-alert> so clicking above/below it doesn’t inadvertently close it [#2375](https://github.com/shoelace-style/shoelace/issues/2375)
     --> we already fixed this on our side and just needed to remove our custom vendorism
- Fixed accessibility issues for elements that are closed while having slotted focused children. [#2383](https://github.com/shoelace-style/shoelace/issues/2383)
       --> we already fixed this on our side and just needed to remove our custom vendorism

## 📑 Test Plan

Check out that everything still works fine

## ✅ DoD


packages/components/.eslintrc
```diff
@@ -19,7 +19,8 @@
       "files": ["./src/**/*.test.ts"],
       "rules": {
         "import/no-extraneous-dependencies": "off",
-        "@typescript-eslint/no-unused-expressions": "off"
+        "@typescript-eslint/no-unused-expressions": "off",
+        "@typescript-eslint/await-thenable": "off"
```
schilchSICKAG, line 23: Where is this from? What led to errors? If this is more common, does it make sense to add it to the lint package?

packages/components/src/components/alert/alert.custom.styles.ts
```diff
@@ -63,7 +63,6 @@ export default css`
     color: var(--syn-typography-color-text);
     margin-block: calc(var(--syn-spacing-x-small) - var(--syn-panel-border-width));
     margin-inline-end: var(--syn-spacing-x-small);
-    padding-inline-end: 0;
```
schilchSICKAG, line 66: Where is this coming from? Should this be there? Are there any visual changes that come from this change?

packages/components/src/components/alert/alert.styles.ts
```diff
@@ -61,7 +61,8 @@ export default css`
     display: flex;
     align-items: center;
     font-size: var(--syn-font-size-medium);
-    padding-inline-end: var(--syn-spacing-medium);
+    margin-inline-end: var(--syn-spacing-medium);
+    align-self: center;
```
schilchSICKAG, line 65: Where does this come from? Did we check this does not lead to breaking changes?

packages/components/web-test-runner.config.js
```diff
@@ -33,13 +33,21 @@ export default {
     })
   ],
   browsers: [
-    playwrightLauncher({ product: 'chromium' }),
+
```
schilchSICKAG, line 36: This means the changes for concurrency are always active, making local tests run slow, too? In this case (and if it still works concurrently in dev), I would suggest to do something like this:

```javascript
browsers: [
  playwrightLauncher({
    product: 'chromium',
    concurrency: process.env.CI ? 1 : undefined,
  });
]
```

This would make sure that we are still concurrent in local dev, but work fine in CI. What do you think?

schilchSICKAG (changes requested):

packages/components/.eslintrc
kirchsuSICKAG: This comes from the components test.
For some reason it will no longer allow to await the `expect`s there

image

schilchSICKAG: Interesting. Is `expect` async? I think it should be, but he does not seem to notice it, so there may be something broken in the chai types?

packages/components/src/components/alert/alert.styles.ts
kirchsuSICKAG: This comes from our ticket #747 
I created a shoelace PR to fix it for them. I needed to use "margin" instead of "padding" otherwise the clickable area was bigger than it should be for the close icon of the closable alert

packages/components/src/components/alert/alert.custom.styles.ts
kirchsuSICKAG: see answer to other question

packages/components/src/components/alert/alert.styles.ts
schilchSICKAG: Valid, we still adjust the alignment, so we need it in both files.

packages/components/src/components/alert/alert.custom.styles.ts
schilchSICKAG: Valid, we still adjust the alignment, so we need it in both files.

packages/components/.eslintrc
schilchSICKAG: I think for the time being it is ok to turn the rule off. However, chai expect was never async, so its a usage problem here instead of a linter error. We should have a look into this in the future as the test system is very brittle and produces false positives. This is not in scope, so resolving this issue.

schilchSICKAG (approved):

PR #836, opened 2025-04-07
feat: ✨ syn-range should align vertically with other input elements

### LINKED ISSUE #809: feat:  ✨ syn-range should align vertically with other input elements

# Version

The bug occurred with the following version of the component library: 2.19.1

# Component

The issue arises when using the following component(s): syn-range

# Environment

# Description
When putting several synergy input element with a syn-range vertically under each other, they do not align horizontally if the knob is either left or right. The thumb of the syn-range goes over the vertical line:

Image

Also see following codepen: https://codepen.io/kirchsuSICKAG/pen/emOxroR

Also we should add an example for how to align labels to the outer line (see second synergy range)

# Acceptance criteria

- [ ] syn-range should be vertically aligned to other inputs
- [ ] story for how to align labels
- [ ] compare pros and cons of solution variants
- [ ] fix knob border over prefix/suffix
- [ ] Check Design Ticket: https://github.com/synergy-design-system/design/issues/174
- [ ] Check Figma Design: https://www.figma.com/design/bZFqk9urD3NlghGUKrkKCR/branch/pfA24ov3VBO3O8YDOrlIg4/Synergy-Digital-Design-System?node-id=32676-156120&t=xexsLz6Ql35HGFJO-11

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adjusts the layout of `<syn-range>`. This is done to prevent the thumb from being drawn outside of the visible area.

### 🎫 Issues

Closes #809 

## 👩‍💻 Reviewer Notes

I just had to add some internal spacing, both negative and positive ones, depending on the element. I added a new internal token named `--half-thumb-size` as the half size of the thumb is now used on at least 3 different places in the code. All original occurrences have been moved to this new token.

## 📑 Test Plan

Have a look at the storybook stories for a visual representation. Notice that the track is still starting on the start of the element, but the know will now happily move to the edge of the visual space when hitting both the left and right boundaries instead of moving over them. This also works when having a prefix or suffix.

## ✅ DoD

### DIFF
--- a/packages/components/src/components/range/range.component.ts
+++ b/packages/components/src/components/range/range.component.ts
@@ -527,7 +527,7 @@ export default class SynRange extends SynergyElement implements SynergyFormContr
     thumb.setAttribute('aria-valuetext', this.tooltipFormatter(value));
     const pos = (value - this.min) / (this.max - this.min);
     // eslint-disable-next-line no-param-reassign
-    thumb.style.insetInlineStart = `calc( ${100 * pos}% - var(--full-thumb-size) / 2 )`;
+    thumb.style.insetInlineStart = `calc(${100 * pos}% - var(--half-thumb-size))`;
     this.#updateTooltip(thumb);
   }
 
--- a/packages/components/src/components/range/range.styles.ts
+++ b/packages/components/src/components/range/range.styles.ts
@@ -16,6 +16,13 @@ export default css`
 
     /* This is needed to get the full with of the element, including the border */
     --full-thumb-size: calc(var(--thumb-size) + (var(--syn-focus-ring-width) * 2));
+
+    /*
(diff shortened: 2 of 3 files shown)


reichst1SICKAG (approved):
Design is good. Chromatic and storybook looks fine

kirchsuSICKAG (approved):

schilchSICKAG:
> Because of the white border we have around the thumb, the syn-range is not completely aligned to the other input element. Now the question is, do we need to take this into account for the calculation?
 
> @schilchSICKAG What do you think?

After having a look at this, we think this is not that problematic at all. We will release with this version. If there are problems with those two white pixels, we are free to adjust this later on.

PR #837, opened 2025-04-07
feat: ✨ improve delayed loading of system icons

### LINKED ISSUE #806: feat: ✨ improve delayed loading of system icons

# Version

The bug occurred with the following version of the component library: 2.22.6

# Component

The issue arises when using the following component(s): syn-icon

# Environment

# Description

It is possible, that in applications, which have a high load our system icons are loaded delayed, as a fetch call is done and the order for this is behind the loading of other stuff. This can e.g. lead that in the header the SICK logo is not yet visible, while the other parts of the header are.

We should adapt the handling of loading system icons, as it is not necessary for them to do a fetch, as we already have the SVG's for them

# Acceptance criteria

- system icons are no longer fetched

### LINKED ISSUE #827: docs: 📚 Remove long loading icon page
# Description
This icon page (https://synergy-design-system.github.io/?path=/docs/components-syn-icon-default-icons--docs), which lists all our icons in alphabetic order, needs a very long time for loading and makes this open tab is no longer usable.
As we now have the icon search page, which is also more performant, this overview is no longer needed and should be removed.

# Acceptance criteria

- [ ] remove the default icons folder with the icons overview
- [ ] add a hint / link to the icon search in the documentation of our syn-icon component

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds some speed boosts for displaying system icons (e.g. `<syn-icon name="menu" library="system"></syn-icon>`. It does this by preventing the `fetch` request that always queued the icon, which could lead to system icons being loaded well after the rest of applications.

It also removes the legacy icon overview that we initially created and that was deprecated by the [Icon Search](https://synergy-design-system.github.io/?path=/docs/icon-search--docs).

### 🎫 Issues

Closes #806
Closes #827

## 👩‍💻 Reviewer Notes

Just have a look at the code. Basically, I stay (nearly) completely with the original code by instanciating the needed `Response` object by myself. I also had to alter the system library to make this work. Note that this file never was part of the vendoring process. This is because the assets package interferes with this anyways.

## 📑 Test Plan

Have a look at storybook. Experiment with the demos. When using a new `<syn-icon library="system" name="menu"></syn-icon>`, you should not see a `fetch` request anymore, but no async actions at all (Technically, they are still async, but as I create the `Resonse` synchronously you will not notice).

## ✅ DoD


packages/docs/stories/components/icon-default-icons.stories.ts
kirchsuSICKAG, line 1: Could you also add an information with the link to the icon search, to the icon documentation (`icon.stories.ts`)?
Maybe add something like

```
## How to find the correct icon? 
Use our [icon search](https://synergy-design-system.github.io/?path=/docs/icon-search--docs) to find a matching icon for your need. 
```

kirchsuSICKAG (changes requested):

packages/docs/stories/components/icon-default-icons.stories.ts
schilchSICKAG: Thats a good idea. Just added this one. Please have another look and tell me if this is the solution we want to use.

PR #842, opened 2025-04-11
feat: ✨ syn-select should allow numeric types as value

### LINKED ISSUE #805: feat: ✨ syn-select should allow numeric types as value

# Component

I suggest to add the feature to the following component: syn-select, syn-combobox, syn-option (unsure if needed)

# Story

As a developer using the Synergy Design System in my application, I want to use my datastores model values with `<syn-select>` directly. Currently, this is not possible when using non string types for the `value` property.

Example:

```typescript
const store = {
  valueAsNumber: 1,
  valueAsString: '1',
};

<!-- Angular Example -->
<syn-select [value]=store.valueAsNumber></syn-select>
<syn-select [value]=store.valueAsString></syn-select>
```

While the string version works, the number version fails the check if the given option is selected, resulting in an empty dropdown list.

# Acceptance criteria

- [x] `<syn-select>` allows a value of type `Number` without ts errors
- [x] An option that has a numeric `value` property is correctly displayed as the `selected` item in a `<syn-select>` with a `Number` value.
- [ ] Shoelace PR
- [x] Check multi select explicitly

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the ability for `<syn-select>` and `<syn-option>` to use type save `Number` values as their corresponding `value` properties.

### 🎫 Issues

Closes #805 

## 👩‍💻 Reviewer Notes

I added the needed test cases in our e2e tests as a (very basic) integration test toolchain. I initially opted for a `Array.map` for the vanilla html solution, However, lits `html` helper will automatically convert all numbers passed in such a loop to a string, so I had to use the regressions functionality.

I also created a new internal package `demo-utilities` that holds shared data and types needed for all framework demos. This should lessen the duplicated code in the future. Initially, I refactored the select tests to make sure it works. Please have a look if this is fine for you the way I created it.

## 📑 Test Plan

See the framework demos for a simple demo. It does not add new items, but I made sure this also works.

## ✅ DoD


kirchsuSICKAG:
Looks nice to me. I like the demo-utilities package ❤️

kirchsuSICKAG (approved):

PR #844, opened 2025-04-14
feat: ✨ Let users configure delimeters for syn-select and syn-range

### LINKED ISSUE #540: feat: ✨ Let users configure delimiters for syn-select

# Story

As a developer using the Synergy Design System, I would like to set a default delimiter for the `value` attribute of `syn-select`so that it is easier to migrate my current application.

---

This issue was brought up from our users migrating to Synergy from our previous system. For current applications is is easier to adjust the value attribute as adjusting the data used as a `syn-option` value for each available values as they then are able to just adjust the view layer, not the logic.

# Acceptance criteria

- [x] `syn-select` allows to set the [delimiter](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/src/components/select/select.component.ts#L119) used together with the `multiple` attribute
- [ ] ~~`syn-range` allows to set the delimiter used together with the `multiple` attribute~~ (**We opted to not include this as it is too complex and there is not a big value in it**)
- [x] `syn-option` uses the parent's delimiter
- [x] Developers are able to set the delimiter via property
- [x] DefaultSettings include the delimiter so it can be set on a global level

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the possibility to configure the delimeter used in `<syn-range>` and `<syn-select>` that is used for both when `multiple` is set. The default is defined as `" "`. It also includes the adjustments needed to make this configurable via `defaultSettings`.

### 🎫 Issues

Closes #540 

## 👩‍💻 Reviewer Notes

I had to include a new private `state` called `delimeter` in the `<syn-option>`, too. This is set when either the `delimeter` of `<syn-select>` changes or it receives new items via `slotChange`.

Angular seems not to take the initial settings into account. After saving the select component in the editor, it works. Probably, there is some timing issue. I was able to mitigate this in testing via setting it explicitly, but this must be further investigated.

## ✅ DoD

### DIFF
--- a/packages/_private/vue-demo/src/AllComponentParts/DemoSelect.vue
+++ b/packages/_private/vue-demo/src/AllComponentParts/DemoSelect.vue
@@ -46,4 +46,16 @@ onMounted(async () => {
       <SynVueOption v-for="item in numericItems" :value="item.id" :key="item.id"> {{ item.label }}</SynVueOption>
     </SynVueSelect>
   </div>
+
+  <syn-select
+    data-testid="select-540-delimeter"
+    delimeter="|"
+    help-text="This select uses a custom delimeter"
+    label="Multiple with custom delimeter"
+    multiple
+    value="1|2"
+  >
+    <SynVueOption v-for="level in levels" :value="level.value" :key="level.value"> {{ level.label }}</SynVueOption>
+  </syn-select>
+
 </template>
--- a/packages/components/src/components/option/option.component.ts
+++ b/packages/components/src/components/option/option.component.ts
@@ -47,6 +47,10 @@ export default class SynOption extends SynergyElement {
 
   @query('.option__label') defaultSlot: HTMLSlotElement;
 
+  // the delimiter used to separate multiple values in a select
+  // This is provided by the wrapping syn-select
+  @state() delimeter = ' ';
+
   @state() current = false; // the user has keyed into the option, but hasn't selected it yet (shows a highlight)
   @state() selected = false; // the option is selected and has aria-selected="true"
   @state() hasHover = false; // we need this because Safari doesn't honor :hover styles while dragging
@@ -117,9 +121,12 @@ export default class SynOption extends SynergyElement {
       this.value = String(this.value);
     }
 
-    if (this.value.includes(' ')) {
-      console.error(`Option values cannot include a space. All spaces have been replaced with underscores.`, this);
-      this.value = this.value.replace(/ /g, '_');
+    const { delimeter } = this;
+
+    if (this.value.includes(delimeter)) {
+      console.error(`Option values cannot include "${delimeter}". All occurrences of "${delimeter}" have been replaced with "_".`, this);
+      const regex = new RegExp(delimeter, 'g');
+      this.value = this.value.replace(regex, '_');
     }
   }
 
--- a/packages/components/src/components/range/range.component.ts
+++ b/packages/components/src/components/range/range.component.ts
@@ -5,6 +5,7 @@ import { ifDefined } from 'lit/directives/if-defined.js';
 import { property, query, queryAll } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
 import type { SynergyFormControl } from '../../internal/synergy-element.js';
+import { watch } from '../../internal/watch.js';
 import { defaultValue } from '../../internal/default-value.js';
 import { FormControlController, customErrorValidityState, validValidityState } from '../../internal/form.js';
 import { HasSlotController } from '../../internal/slot.js';
@@ -113,14 +114,23 @@ export default class SynRange extends SynergyElement implements SynergyFormContr
   /** The preferred placement of the range's tooltip. Use "none" to disable the tooltip */
   @property({ attribute: 'tooltip-placement', type: String }) tooltipPlacement: 'top' | 'bottom' | 'none' = 'top';
 
+  /**
+   * The delimiter to use when setting the value when `multiple` is enabled.
+   * The default is a space, but you can set it to a comma or other character.
+   * @example <syn-select delimeter="|" value="option-1|option-2"></syn-select>
+   */
+  @property() delimeter = ' ';
+
   /** The current values of the input (in ascending order) as a string of space separated values */
   @property({ type: String })
   set value(value: string | null) {
-    this.#value = value ? value.split(' ').map(n => +n).sort(numericSort) : [];
+    this.#value = value
+      ? value.split(this.delimeter).map(Number).sort(numericSort)
+      : [];
   }
 
   get value() {
-    return this.#value.slice().sort(numericSort).join(' ');
+    return this.#value.slice().sort(numericSort).join(this.delimeter);
   }
 
   /**
@@ -134,7 +144,7 @@ export default class SynRange extends SynergyElement implements SynergyFormContr
     const oldValue = this.#value;
     this.#value = Array.isArray(value) ? value.slice().sort(numericSort) : value || [];
     if (arraysDiffer(oldValue, this.#value)) {
-      this.requestUpdate('value', oldValue.join(' '));
+      this.requestUpdate('value', oldValue.join(this.delimeter));
     }
   }
 
@@ -348,6 +358,11 @@ export default class SynRange extends SynergyElement implements SynergyFormContr
   /** Gets the validation message */
   public get validationMessage() {
     return this.#validationError;
+  }
+
+  @watch('delimeter')
+  handleDelimeterChange() {
+    this.value = this.#value.join(this.delimeter);
   }
 
   #onClickTrack(event: PointerEvent, focusThumb = true) {
--- a/packages/components/src/components/select/select.component.ts
+++ b/packages/components/src/components/select/select.component.ts
@@ -112,6 +112,13 @@ export default class SynSelect extends SynergyElement implements SynergyFormCont
   @state() selectedOptions: SynOption[] = [];
   @state() private valueHasChanged: boolean = false;
 
+  /**
+   * The delimiter to use when setting the value when `multiple` is enabled.
+   * The default is a space, but you can set it to a comma or other character.
+   * @example <syn-select delimeter="|" value="option-1|option-2"></syn-select>
+   */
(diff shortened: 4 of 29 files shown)


packages/components/src/components/option/option.component.ts
```diff
@@ -117,9 +121,12 @@ export default class SynOption extends SynergyElement {
...
     }
 
-    if (this.value.includes(' ')) {
-      console.error(`Option values cannot include a space. All spaces have been replaced with underscores.`, this);
-      this.value = this.value.replace(/ /g, '_');
+    const { delimiter } = this;
+
+    if (this.value.includes(delimiter)) {
+      console.error(`Option values cannot include "${delimiter}". All occurrences of "${delimiter}" have been replaced with "_".`, this);
+      const regex = new RegExp(delimiter, 'g');
```
kirchsuSICKAG, line 128: I think we need to do this a little bit more complex. As if someone is using a special character of regex (e.g. `.` which means any character , or `$` which means end of string, ...)
the replacement of the value will lead to strange behavior. So we need to escape the special regex characters with their escaped versions.

Reproductions to see the issue.
## Example 1 with `.`:

```html
  <syn-select value="test 1" delimiter=".">
    <syn-option value="test 1">opt 1</syn-option>
    <syn-option value="test. 2">opt 2</syn-option>
    <syn-option value=" test 3">opt 3</syn-option>
  </syn-select>
```
This results into following option value, because every single character is exchanged with `_`:
image

## Example 2 with `$`:

```html
  <syn-select value="test 1" delimiter="$">
    <syn-option value="test 1">opt 1</syn-option>
    <syn-option value="test$ 2">opt 2</syn-option>
    <syn-option value=" test 3">opt 3</syn-option>
  </syn-select>
  ```
  
image

## Possible fix
We could escape the special regex characters. According to copilot this could be the fix:

```js 
   const escapedDelimiter = delimiter.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
      const regex = new RegExp(escapedDelimiter, 'g');
```

Fo both of my examples this works

packages/components/src/components/select/select.component.ts
```diff
@@ -676,6 +692,13 @@ export default class SynSelect extends SynergyElement implements SynergyFormCont
     this.formControlController.emitInvalidEvent(event);
   }
 
+  @watch('delimiter')
+  handleDelimiterChange() {
+    this.getAllOptions().forEach(option => {
+      option.delimiter = this.delimiter;
+    });
```
kirchsuSICKAG, line 699: It it not enough to do this just in the watcher of the `delimiter`.
We also need to do this in the `handleDefaultSlotChange` otherwise it is not working for dynamically added options.

image

```html
  <syn-select value="test 1" delimiter="--">
    <syn-option value="test 1">opt 1</syn-option>
    <syn-option value="test 2">opt 2</syn-option>
    <syn-option value=" test 3">opt 3</syn-option>
  </syn-select>
  <script type="module">
    const select = document.querySelector('syn-select');
    const newOption = document.createElement('syn-option');
    newOption.value = 'test 4';
    newOption.innerText = 'opt 4';
    select.appendChild(newOption);
  </script>
```

packages/components/src/components/select/select.component.ts
kirchsuSICKAG, line 1: I think we need to add the `delimiter` to the watcher of  `handleValueChange`
I would expect, that the value selection changes if I change the `delimiter`.

so instead of:

```js
@watch(['defaultValue', 'value'], { waitUntilFirstUpdate: true })
  handleValueChange() {
  ```
  we have: 
  
  ```js
    @watch(['defaultValue', 'value', 'delimiter'], { waitUntilFirstUpdate: true })
  handleValueChange() {
  ```
  
  ## Current behavior
delimiter_old

## Behavior with `delimiter` in watcher
delimiter_new

File: packages/components/src/components/range/range.component.ts [hunk 3]
```diff
@@ -350,6 +360,11 @@ export default class SynRange extends SynergyElement implements SynergyFormContr
     return this.#validationError;
   }
 
+  @watch('delimiter')
+  handleDelimiterChange() {
+    this.value = this.#value.join(this.delimiter);
```
kirchsuSICKAG, line 365: The delimiter seems not to work initially. The value need to change before it takes action

```html
  <syn-range value="20.40" delimiter="."></syn-range>
```

range_delimiter

kirchsuSICKAG (changes requested):

packages/components/src/components/option/option.component.ts
schilchSICKAG: Ha, thats a good one. Did not even think about this, good catch :). Unfortunately, I cannot get this to work via vendorism, so will create this as a utility function.

packages/components/src/components/range/range.component.ts
schilchSICKAG: Not sure that I understand this one. If you initial value is 2040 its not a valid value for limiting. Rendering the same code in Storybook (`<syn-range value="20.40" delimiter="."></syn-range>`) give two knobs initially. I think we should have a look together.

packages/components/scripts/vendorism/custom/select.vendorism.js
```diff
@@ -121,6 +124,67 @@ const transformComponent = (path, originalContent) => {
...
+    'const val = this.valueHasChanged ? this.value : this.defaultValue;',
+    `
+    let newValue = val;
+    if (this.multiple) {
+      if (!Array.isArray(this.defaultValue)) {
+        newValue = typeof this.defaultValue === 'string' ? this.defaultValue.split(this.delimiter) : [this.defaultValue].filter(Boolean);
+      }
+    } else {
+      newValue = Array.isArray(this.defaultValue) ? this.defaultValue.join(this.delimiter) : this.defaultValue;
+    }
```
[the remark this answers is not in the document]
schilchSICKAG: Funny thing is the code works with the defaultValue set as it is currently set, but not with the changes you are requesting. Maybe we will have to look into this again.

packages/components/src/components/select/select.component.ts
schilchSICKAG: Fixed in latest

packages/components/src/components/select/select.component.ts
schilchSICKAG: Seems legit, just added it.

packages/components/src/components/range/range.component.ts
schilchSICKAG: Ignore my previous comment, I see where the problem is coming from. Works in frameworks, but not in native locally :(.

schilchSICKAG: This will never work correctly because of the way we use our `value` `setters` and `getters`. The value of `this.delimiter` is always the original default one, as the `setter` runs **before** the initialization of the property in the `@property` decorator. Therefore, we would have needed something like this to make it work:

```typescript
set value() {
  const delimiter = this.getAttribute('delimiter') || this.delimiter;
  // do splitting here...
}
```

However, this may clash when the attribute and property are out of sync. If not doing it like this, the following code will give different results:

```html
<!-- working example -->
<syn-range delimiter="|" value="10|20">/syn-range>

<!-- broken example -->
<syn-range value="10|20" delimiter="|">/syn-range>
```

Properties are not set in the same time, but the DOM order of the property matters. As this is quite unstable, I opted to remove it.

packages/_private/vue-demo/src/AllComponentParts/DemoSelect.vue
```diff
@@ -46,4 +46,16 @@ onMounted(async () => {
...
+  <syn-select
+    data-testid="select-540-delimiter"
+    delimiter="|"
+    help-text="This select uses a custom delimiter"
+    label="Multiple with custom delimiter"
+    multiple
+    value="1|2"
+  >
+    <SynVueOption v-for="level in levels" :value="level.value" :key="level.value"> {{ level.label }}</SynVueOption>
+  </syn-select>
```
kirchsuSICKAG, line 59: The custom-element (`syn-select`)  is used here for the select instead of the vue wrapper (`SynVueSelect`)

packages/components/src/components/select/select.component.ts
```diff
@@ -533,7 +540,18 @@ export default class SynSelect extends SynergyElement implements SynergyFormCont
...
+
+    let newValue = val;
+    if (this.multiple) {
+      if (!Array.isArray(this.defaultValue)) {
+        newValue = typeof this.defaultValue === 'string' ? this.defaultValue.split(this.delimiter) : [this.defaultValue].filter(Boolean);
+      }
+    } else {
+      newValue = Array.isArray(this.defaultValue) ? this.defaultValue.join(this.delimiter) : this.defaultValue;
+    }
+    const value = Array.isArray(newValue) ? newValue : [newValue];
```
kirchsuSICKAG, line 554: Do we want this to stay or do we want to remove it again? As all other frameworks work without this and angular seems to do strange stuff and does not work correctly anyways?

packages/_private/vue-demo/src/AllComponentParts/DemoSelect.vue
schilchSICKAG: Good catch :)

packages/docs/stories/components/range.stories.ts
```diff
@@ -363,6 +363,7 @@ export const MultiThumb: Story = {
   },
   render: () => html`
     <syn-range max="100" min="0" value="30 70"></syn-range>
+    <syn-range max="100" min="0" value="20|40" delimiter="|"></syn-range>
```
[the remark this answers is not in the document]
schilchSICKAG: Removed

packages/components/src/components/select/select.component.ts
schilchSICKAG: I think I will remove this, but we should create a new ticket that will (hopefully) lead us to a solution on this issues.

schilchSICKAG: New ticket created at https://github.com/synergy-design-system/synergy-design-system/issues/847

kirchsuSICKAG (approved):

PR #852, opened 2025-04-25
fix: 🐛 syn-input has some paddings which are not aligned with design

### LINKED ISSUE #817: fix: 🐛 input without prefix has additional padding to suffix

# Version
2.22.6

# Component

syn-input

# Environment

Angular

# Description

According to Figma, the value of a syn-input should be padded with spacing-small to the suffix. 
This works if a prefix and suffix exist. But when only a suffix exists, the value is padded with additional 8px

# Acceptance criteria

- [ ] suffix is always padded with spacing/small according to figma

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR fixes some non-alignment of inner paddings with design.
Non alignment appears if:
- having only a suffix slot
- having clearable property set
- having password-toggle set

### 🎫 Issues
Closes #817.

## 📑 Test Plan
Example code for testing:

```html
  <syn-input label="Correct right padding" value="aaaasadasdjasdasdasdasdasdasdasdasdasdasdsdsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdf">
    <span slot="prefix">prefix</span>
    <span slot="suffix">suffix</span>
  </syn-input>
  <syn-input label="Incorrect right padding" value="aaaasadasdjasdasdasdasdasdasdasdasdasdasdsdsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdf">
    <span slot="suffix">suffix</span>
  </syn-input>
<syn-input clearable label="Incorrect right padding" value="aaaasadasdjasdasdasdasdasdasdasdasdasdasdsdsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdf">
  </syn-input>

<syn-input password-toggle label="Incorrect right padding" value="aaaasadasdjasdasdasdasdasdasdasdasdasdasdsdsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdfsdf">
  </syn-input>
 
```

## ✅ DoD

### DIFF
--- a/packages/components/src/components/input/input.custom.styles.ts
+++ b/packages/components/src/components/input/input.custom.styles.ts
@@ -143,6 +143,16 @@ export default css`
     padding-right: var(--syn-spacing-small);
   }
 
+  .form-control--has-prefix .input__control {
+    padding-left: 0;
+  }
+
+  .form-control--has-suffix .input__control,
+  .input:has(.input__clear) .input__control,
+  .input:has(.input__password-toggle) .input__control {
+    padding-right: 0;
+  }
+
   :host([type='number']) .input--large:not(.input--no-spin-buttons) .input__clear,
   :host([type='number']) .input--large:not(.input--no-spin-buttons) .input__password-toggle {
     padding-right: var(--syn-spacing-medium);
--- a/packages/components/src/components/input/input.custom.styles.ts
+++ b/packages/components/src/components/input/input.custom.styles.ts
@@ -143,10 +143,12 @@ export default css`
     padding-right: var(--syn-spacing-small);
   }
 
+  /* Fixes wrong paddings on some suffix special cases: https://github.com/synergy-design-system/synergy-design-system/issues/817  */
   .form-control--has-prefix .input__control {
     padding-left: 0;
   }
 
+  /* Fixes wrong paddings on some suffix special cases: https://github.com/synergy-design-system/synergy-design-system/issues/817  */
   .form-control--has-suffix .input__control,
   .input:has(.input__clear) .input__control,
   .input:has(.input__password-toggle) .input__control {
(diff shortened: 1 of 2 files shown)


packages/components/src/styles/form-control.custom.styles.ts
```diff
@@ -24,18 +24,6 @@ export default css`
     margin-top: var(--syn-spacing-x-small);
   }
 
-  .form-control--small.form-control--has-prefix .input__control  {
```
schilchSICKAG, line 27: This may have side effects for the other input types relying on this. Did you test `syn-select`, `syn-range` and others for compatibility?

packages/components/src/components/input/input.custom.styles.ts
```diff
@@ -143,6 +143,16 @@ export default css`
     padding-right: var(--syn-spacing-small);
   }
 
+  .form-control--has-prefix .input__control {
```
schilchSICKAG, line 146: I know why you added this here, but may it be better to at least add comments why this is here?

schilchSICKAG (changes requested):

packages/components/src/styles/form-control.custom.styles.ts
kirchsuSICKAG: The only component, where this has side-effects is the sy-range. It is the only component where the `form-control--has-prefix` class is also used.
I think it was wrong at the first place, that this styling was added to the general form-control.custom.style.ts as this is an input specific styling.
The syn-range now looks different and I think this is correct

packages/components/src/components/input/input.custom.styles.ts
kirchsuSICKAG: done

schilchSICKAG (approved):

Issue #811, opened 2025-03-18
feat: ✨ syn-details exports content_wrapper as CSS part


schilchSICKAG:
# Checklist (for the issue creator)

- [x] I have looked at the documentation site and was not able to find the required feature.
- [x] I have looked into the Figma boards and was not able to find the required feature.
- [x] I would be willing to provide a PR for this feature.

# Component

I suggest to add the feature to the following component: syn-details

# Story

As a developer using the Synergy Design System, I would like to be able to override styles for the div wrapping the `<syn-detail>`s content so that I am able to customize spacing and the overflow behavior.

# Acceptance criteria

- [x] All DoD items successfully completed
- [x] A new part "details-body" is added to the HTML output and documentation

PR #853, opened 2025-04-28
feat: ✨ number-input - allow clamping via property

### LINKED ISSUE #417: feat: ✨ number-input - allow clamping via property
# Story

As a user of the Synergy Design System, I would like to have the value clamped in a number-input. If the user inputs a value which is out of range should be set to the min or max value. Then I do not need to check what the min or max value is.

# Acceptance criteria

- [x] Have a story which shows how this can be achieved
- [x] Input of a value > max value is set to max
- [x] Input of a value < min value is set to min

### LINKED ISSUE #812: feat: ✨ Invalid syn-input when step is set

# Version

The bug occurred with the following version of the component library: 2.24.0

# Component

The issue arises when using the following component(s): syn-input

# Description

syn-input becomes invalid when step is set, see:
https://codepen.io/weissti1SICKAG/pen/JojLbMd?editors=1110

# Acceptance criteria

### LINKED ISSUE #818: fix: 🐛

# Version

The bug occurred with the following version of the component library: 2.22.6

# Component

The issue arises when using the following component(s): syn-input

# Environment

# Description

`<syn-input type="number" min="0" max="43" step="10" value=20 />` will not allow to step higher than 40 with the spinner-button and the increase-button will never be disabled at the same time. Expected beahvior would be to allow a final increase-step from 40 to 43 and then have the increase-button disabled. The issue becomes even more propblematic, when the input is combined / synced with a syn-range, that allows to increase to the actual max-value of 43. The input is then set to 43, the increase-button is disabled but (see other bug) the field is invalid. 

# Acceptance criteria

### LINKED ISSUE #838: feat: ✨ number-input: min-/max- fraction digits
# Component

I suggest to add the feature to the following component: syn-input

# Story

As a user of the Synergy Design System, I would like to show customized number of fraction digits so that the displayed precision fits to my application.

# Acceptance criteria

- [x] add properties: min-/max-fraction-digits
- [x] default for min should be 0
- [x] default for max should be ♾️
- [x] display fraction digits according to the set properties

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the first iteration of `<syn-input type="number" />`s new [configurable numeric input behavior](https://github.com/orgs/synergy-design-system/discussions/839). As the first of a list of [changes tackling this topic](https://github.com/synergy-design-system/synergy-design-system/issues?q=state%3Aopen%20label%3A%22syn-input%2Fnumber%22), it includes the needed infrastructure for further development, as well as the number clamping and step alignment features as proposed in #417, #818 and #811.

### 🎫 Issues

Closes #417
Closes #818
Closes #812
Closes #838 

## 👩‍💻 Reviewer Notes

Please have a look especially on documentation, functionality of `defaultSettings` and event emission. Demos have been updated with native and modern strategies and tests have been written for all available frameworks.

## 📑 Test Plan

Check out the demos of how to use the new feature (e.g.. `YOUR_FRAMEWORK/AllComponents/Input`). Play around with the changes. Also try to set completely invalid values.

Also note we are **only** validating the user input. If you set something via property, it will NOT clamp when using the `modern` mechanism. This is done to still show the value that is available in a store at least.

The noStepAlign and noStepValidation features are now also implemented and can be tested by using the following example code (e.g. via a temporary Story):

```html
export const NumericStrategy: Story = {
  render: () => html`
    <form>
      <syn-input label="Native" type="number" min="0" max="10" value="0" step="3"></syn-input>
      <br />
      <syn-input label="modern" type="number" min="0" max="10" value="0" step="3" numeric-strategy="modern"></syn-input>
      <button type="submit">hello</button>
    </form>
    <script>
      document.addEventListener('syn-clamp', console.log);
      document.addEventListener('syn-input', console.log);
      document.addEventListener('syn-change', console.log);
      document.querySelector('form').addEventListener('syn-invalid', console.log, { capture: true });
    </script>
  `,
};
```

The first steps for number clamping are also available. This means that the new properties `min-fraction-digits`, `max-fraction-digits` and `numberFormatterOptions` are now available.

This can be tested with the following code:

```html
<!-- Using a minimum via prop -->
<syn-input variant="modern" min-fraction-digits="4"></syn-input>

<!-- Fallback: Use the digits provided via step -->
<syn-input variant="modern" step="0.1234"></syn-input>
```

## ✅ DoD


klaarseSICKAG:
Could we discuss alternative namings to numericStrategy="modern" ?
What about just name it to what it does? "clamping"
From a users perspective the documentation " 'modern': Uses the modern implementation. " is not especially helpful to figure out what it exactly does ...

kirchsuSICKAG (changes requested):

packages/components/scripts/vendorism/custom/input.vendorism.js
```diff
@@ -182,6 +187,128 @@ export`,
...
+  @query('[part=increment-number-stepper]') incrementButton: HTMLButtonElement;`,
+    ],
+    // Add the numeric strategy property
+    [
+      "@property() inputmode: 'none' | 'text' | 'decimal' | 'numeric' | 'tel' | 'search' | 'email' | 'url';",
+      `
+  #numericStrategy: NumericStrategy = nativeNumericStrategy;
+
+  /**
+   * Defines the strategy for numeric inputs.
```
klaarseSICKAG, line 221:
```suggestion
   * Defines the strategy for handling numbers in the numeric input.
```

packages/components/scripts/vendorism/custom/input.vendorism.js
```diff
@@ -182,6 +187,128 @@ export`,
...
+   * This is used to determine how the input behaves when the user interacts with it.
+   * Includes the following configuration options:
+   * 1. autoClamp: If true, the input will clamp the value to the min and max attributes.
+   * 
+   * You may provide this as one of the following values:
+   * - 'native': Uses the native browser implementation.
+   * - 'modern': Uses the modern implementation.
```
klaarseSICKAG, line 228:
```suggestion
   * - 'modern': Uses a more intuitive implementation: values are clamped to the nearest min or max value, the value is not aligned to 0 + n*step, the validation doesn't depend on the value being on a step number (n*step)
```

klaarseSICKAG (commented):
Nice work!

packages/components/src/components/input/input.component.ts
```diff
@@ -90,6 +98,8 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
   private readonly localize = new LocalizeController(this);
 
   @query('.input__control') input: HTMLInputElement;
+  @query('[part=decrement-number-stepper]') decrementButton: HTMLButtonElement;
+  @query('[part=increment-number-stepper]') incrementButton: HTMLButtonElement;
```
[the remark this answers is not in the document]
schilchSICKAG: You are right, I will remove them. Seems to be a leftover

packages/components/scripts/vendorism/custom/input.vendorism.js
schilchSICKAG: This is nice. I am not able to include this in the first version, though. The other features (noStepAlign etc) are not yet implemented. Will adjust this as we implement the new feature flags

packages/components/scripts/vendorism/custom/input.vendorism.js
schilchSICKAG: Adjusted, thanks for the input!

packages/components/scripts/vendorism/custom/input.vendorism.js
schilchSICKAG: Adjusted, thanks for the input!

packages/components/src/components/input/input.component.ts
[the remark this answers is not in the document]
schilchSICKAG: Removed, thanks for the input!

packages/components/src/components/input/input.component.ts
```diff
@@ -218,6 +228,47 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+
+  /**
+   * Defines the strategy for numeric inputs.
+   * This is used to determine how the input behaves when the user interacts with it.
+   * Includes the following configuration options:
+   * 1. autoClamp: If true, the input will clamp the value to the min and max attributes.
+   * 
+   * You may provide this as one of the following values:
+   * - 'native': Uses the native browser implementation.
+   * - 'modern': Uses the modern implementation.
```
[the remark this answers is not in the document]
schilchSICKAG: After internal discussion, we will leave this as is for the time being.

packages/components/src/components/input/input.component.ts
```diff
@@ -344,6 +433,20 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
   }
 
   private handleKeyDown(event: KeyboardEvent) {
+    if (this.#numericStrategy.noStepAlign) {
```
kirchsuSICKAG, line 436: Here should also be checked if type="number". 
Otherwise following happens if doing this and using up-down arrows:

```html
      <syn-input label="modern"  value="1dfgdg" numeric-strategy="modern"></syn-input>
```

input-stepalign

kirchsuSICKAG (changes requested):

packages/components/src/components/input/input.component.ts
```diff
@@ -344,6 +433,20 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+        event.preventDefault();
+        event.stopPropagation();
+      }
+
+      if (key === 'ArrowUp' && !this.isIncrementDisabled()) {
+        this.handleStepUp();
+      } else if (key === 'ArrowDown' && !this.isDecrementDisabled()) {
+        this.handleStepDown();
+      }
+      return;
```
kirchsuSICKAG, line 448: The return should be removed. 
Otherwise the `enter` handling does not take action for this strategy (input is submitted on enter in a form)

kirchsuSICKAG (changes requested):

packages/components/src/components/input/input.component.ts
```diff
@@ -437,6 +540,19 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+    if (this.#numericStrategy.noStepAlign) {
+      const { max, step, valueAsNumber } = this;
+
+      const stepToUse = step === 'any' ? 1 : typeof step === 'number' ? step : parseFloat(step);
+      const maxToUse = max === undefined || max === null ? Number.POSITIVE_INFINITY : typeof max === 'string' ? parseFloat(max) : max;
+      const nextValue = Math.min(valueAsNumber + stepToUse, maxToUse);
+
+      // If the input is empty, nextValue will be NaN.
+      // The following check will ensure that the input is set to the max value instead.
+      this.input.value = Number.isNaN(nextValue) ? maxToUse.toString() : nextValue.toString();
```
kirchsuSICKAG, line 552: Why using the max number?
The native input uses  `0 + step size` as next value. This also feels more intuitive for me as the max number

packages/components/src/components/input/input.component.ts
```diff
@@ -445,6 +561,19 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+    if (this.#numericStrategy.noStepAlign) {
+      const { min, step, valueAsNumber } = this;
+
+      const stepToUse = step === 'any' ? 1 : typeof step === 'number' ? step : parseFloat(step);
+      const minToUse = min === undefined || min === null ? Number.NEGATIVE_INFINITY : typeof min === 'string' ? parseFloat(min) : min;
+      const nextValue = Math.max(valueAsNumber - stepToUse, minToUse);
+
+      // If the input is empty, nextValue will be NaN.
+      // The following check will ensure that the input is set to the min value instead.
+      this.input.value = Number.isNaN(nextValue) ? minToUse.toString() : nextValue.toString();
```
kirchsuSICKAG, line 573: Why using the min number?
The native input uses  `0 - step size` as next value. This also feels more intuitive for me as the min number

kirchsuSICKAG (changes requested):

reichst1SICKAG:
We also need some more Docs with the docs-token to explain the story in storybook and also add this in Figma before we release this ;)

schilchSICKAG:
> We also need some more Docs with the docs-token to explain the story in storybook and also add this in Figma before we release this ;)

Ah sorry, this is just a test story we use for dev until the feature is ready. It will be removed when we are ready ironing our the issues still left.

packages/components/src/components/input/input.component.ts
schilchSICKAG: Fixed

packages/components/src/components/input/input.component.ts
schilchSICKAG: This is now moved into the check if the keys are the correct ones. I don´t want it to have any other side effects, so would like to leave the return statement.

packages/components/src/components/input/input.component.ts
```diff
@@ -544,7 +673,7 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
               maxlength=${ifDefined(this.maxlength)}
               min=${ifDefined(this.min)}
               max=${ifDefined(this.max)}
-              step=${ifDefined(this.step as number)}
+              step=${ifDefined(!this.#numericStrategy.noStepValidation ? this.step as number : undefined)}
```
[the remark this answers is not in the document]
schilchSICKAG: Fixed in latest

packages/components/src/components/input/input.component.ts
schilchSICKAG: This is fixed in the latest version

packages/components/src/components/input/input.component.ts
schilchSICKAG: This is fixed in the latest version

packages/components/src/components/input/input.component.ts
```diff
@@ -218,6 +226,51 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+   *
+   * Includes the following configuration options:
+   *
+   * - **autoClamp**: If true, the input will clamp the value to the min and max attributes.
+   * 
+   * You may provide this as one of the following values:
+   *
+   * - 'native': Uses the native browser implementation.
+   * - 'modern': Uses a more intuitive implementation: Values are clamped to the nearest min or max value.
+   * - An object that matches the `NumericStrategy` type. Note this can only be set via `property`, not as an `attribute`!
```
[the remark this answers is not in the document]
schilchSICKAG: Done

reichst1SICKAG:
> > Could we discuss alternative namings to numericStrategy="modern" ?
> 
> I am open to suggestions for the naming. It has to be something that can be used as a general term, because clamping alone is just one of the features of the `number-strategy` that we will implement, see the linked discussion for the reasoning behind the names.
> 
> > What about just name it to what it does? "clamping"
> 
> This won´t fit, because we will add other settings into this as well to prevent property pollution. You would need various new properties only for the new settings to be applied. Note boolean properties cannot be used because they cannot be set to true per default, making a sring or object a requirement.
> 
> > From a users perspective the documentation " 'modern': Uses the modern implementation. " is not especially helpful to figure out what it exactly does ...
> 
> This is described in the accompanying ADR. You can also see this pointed out in the documentation.
> 
> We can for sure discuss this, but the API proposal was done way back before and I personally think is not something we want to change at this time.

I have asked ChatGTP and here are it's suggestions:

```
enhanced	Signals “all the same under the hood, but with extra polish and features”
advanced	Implies a step-up for power users or richer use cases
smart		Connotes automation or “it just works” behavior (e.g. auto-formatting)
rich		Suggests richer interactivity beyond the native arrows
pro		A bit more marketing-y—“for professional use”
plus		Friendly, informal, and indicates “more” without being too heavy-handed
custom		Emphasizes “your styling, your logic, your way” (but is fairly generic)
friendly	Conveys an easier, more approachable UX
streamlined	Focus on a smoother, friction-free interaction
interactive	Stresses interactivity over the browser default
```

I like enhanced or smart

schilchSICKAG:
> > > Could we discuss alternative namings to numericStrategy="modern" ?
> > 
> > 
> > I am open to suggestions for the naming. It has to be something that can be used as a general term, because clamping alone is just one of the features of the `number-strategy` that we will implement, see the linked discussion for the reasoning behind the names.
> > > What about just name it to what it does? "clamping"
> > 
> > 
> > This won´t fit, because we will add other settings into this as well to prevent property pollution. You would need various new properties only for the new settings to be applied. Note boolean properties cannot be used because they cannot be set to true per default, making a sring or object a requirement.
> > > From a users perspective the documentation " 'modern': Uses the modern implementation. " is not especially helpful to figure out what it exactly does ...
> > 
> > 
> > This is described in the accompanying ADR. You can also see this pointed out in the documentation.
> > We can for sure discuss this, but the API proposal was done way back before and I personally think is not something we want to change at this time.
> 
> I have asked ChatGTP and here are it's suggestions:
> 
> ```
> enhanced	Signals “all the same under the hood, but with extra polish and features”
> advanced	Implies a step-up for power users or richer use cases
> smart		Connotes automation or “it just works” behavior (e.g. auto-formatting)
> rich		Suggests richer interactivity beyond the native arrows
> pro		A bit more marketing-y—“for professional use”
> plus		Friendly, informal, and indicates “more” without being too heavy-handed
> custom		Emphasizes “your styling, your logic, your way” (but is fairly generic)
> friendly	Conveys an easier, more approachable UX
> streamlined	Focus on a smoother, friction-free interaction
> interactive	Stresses interactivity over the browser default
> ```
> 
> I like enhanced or smart

The problem with all of those (including modern) is they are subjective. From what I heard from the community they prefer modern, but in about 5 years, what once looked modern is is "antiquated"? :). We could also call it `synergy`, as it may be the new default value for Synergy Version 3 anyhow, moving away from the default of `native`. As said, I am open on changing this. Lets get some more feedback from the community in the first beta phase and see how the naming and functionality is received.

packages/components/src/components/input/input.component.ts
```diff
@@ -344,6 +436,20 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
 
   private handleKeyDown(event: KeyboardEvent) {
+    if (this.#numericStrategy.noStepAlign && this.type === 'number') {
+      const { key } = event;
+      if (key === 'ArrowUp' || key === 'ArrowDown') {
+        event.preventDefault();
+        event.stopPropagation();
+
+        if (key === 'ArrowUp') {
+          this.handleStepUp();
```
kirchsuSICKAG, line 446: Currently there is a bit strange behavior with the modern strategy and events.

1. when using keybord up and down, no longer a syn-change & syn-input is emitted but only syn-input
2. being at the max number of the input, using up-key results in emitting syn-input& syn-clamping event, everytime i use up-arrow. For me I would expect nothing happens here


input-events

packages/components/src/components/input/input.component.ts
```diff
@@ -445,6 +595,49 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+
+      // Needed because the input could be empty. In this case, valueAsNumber is NaN
+      const usedInitialValue = Number.isNaN(valueAsNumber) ? 0 : valueAsNumber;
+      const usedMin = typeof min === 'string' ? parseFloat(min) : min;
+      const usedMax = typeof max === 'string' ? parseFloat(max) : max;
+      const usedStep = (typeof step === 'undefined' || step === null || step === 'any') ? 1 : typeof step === 'number' ? step : parseFloat(step);
+
+      let wantedNextValue = usedInitialValue - usedStep;
+      // Floating point issues...
+      wantedNextValue = Math.abs(wantedNextValue) < Number.EPSILON ? 0 : wantedNextValue;
```
kirchsuSICKAG, line 609: Shouldn't this line resolve this issue or is this another problem?: 

floating-point-issue

packages/components/src/events/syn-clamped.ts
```diff
@@ -0,0 +1,20 @@
+export type SynClampedEvent = CustomEvent<{
+  /**
+   * The "direction" the value was clamped to
+   */
+  clampedTo: 'min' | 'max';
+
+  /**
+   * The last value value that the user has input BEFORE clamping
```
klaarseSICKAG: value value

packages/components/src/components/input/strategies.ts
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Clamp the inputs value to its min and max values when reaching it.
+   *
+   * When set to false, this will follow the HTML5 spec:
+   * - Allows overflow of min and max via user input or spinners
+   * - Will emit an invalid event
+   * - Will emit an input event
+   * - Will emit a change event
+   *
+   * When set to true, this will follow synergy specs:
```
klaarseSICKAG, line 14: Captial S for Synergy

packages/components/src/events/syn-clamped.ts
schilchSICKAG: Fixed

packages/components/src/components/input/strategies.ts
schilchSICKAG: Fixed

packages/components/src/components/input/input.component.ts
```diff
@@ -472,6 +685,43 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+    
+    // Check if there are any options that where provided via formatter options
+    if (typeof numberFormatterOptions === 'object') {
+      return true;
+    }
+
+    // As a last fallback, see if the step has a decimal value
+    // If it has, we should format according to the steps amount of fraction digits
+    const stepToUse = step === 'any' || !step ? 1 : +step;
+    const stepFractionDigits = stepToUse.toString().split('.')[1]?.length || 0;
```
kirchsuSICKAG, line 720: Is a split of `split('.')` always correct? Isn't it dependent on the locale (browser / OS)?

packages/components/src/components/input/formatter.ts
```diff
@@ -0,0 +1,71 @@
...
+    minimumFractionDigits: providedMinFractionDigits,
+    ...otherOptions
+  } = numberFormatterOptions;
+
+  // There may be times where the user sets something like this:
+  // <syn-input min-fraction-digits="0" step="0.3">
+  // In this case, we need to use the step value to determine the fraction digits
+  // to use for the formatter.
+  const stepToUse = step === 'any' || !step ? 1 : +step;
+  const stepFractionDigits = stepToUse.toString().split('.')[1]?.length || 0;
```
kirchsuSICKAG, line 25: Is a split of `split('.')` always correct? Isn't it dependent on the locale (browser / OS)?

kirchsuSICKAG (changes requested):

packages/components/src/components/input/formatter.ts
```diff
@@ -0,0 +1,71 @@
...
+    maximumFractionDigits = providedMinFractionDigits;
+  } else if (providedMaxFractionDigits && !providedMinFractionDigits) {
+    // If only max is set, use the max both min and max
+    minimumFractionDigits = providedMaxFractionDigits;
+    maximumFractionDigits = providedMaxFractionDigits;
+  } else if (providedMaxFractionDigits && providedMinFractionDigits) {
+    // If both are set, use the provided values, but make sure to sort them first
+    minimumFractionDigits = Math.min(providedMinFractionDigits, providedMaxFractionDigits);
+    maximumFractionDigits = Math.max(providedMinFractionDigits, providedMaxFractionDigits);
+  }
```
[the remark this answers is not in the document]
schilchSICKAG: Fixed in latest

packages/components/src/components/input/input.component.ts
schilchSICKAG: The lit number bindings will not accept a value of `0,8`, but will use the provided attribute value as a number directly. Therefore, if we would like the users to be able to provide it like `<syn-input step="0,8">` we would have to do a lot more. For this reason, I opted to just leave the step be as a `Number`.

packages/components/src/components/input/input.component.ts
```diff
@@ -472,6 +685,43 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
     this.formControlController.updateValidity();
   }
 
+  #formatNumber(value: number) {
+    return formatNumber(value, this.step, {
+      maximumFractionDigits: this.maxFractionDigits,
+      minimumFractionDigits: this.minFractionDigits,
+      ...this.numberFormatterOptions,
+    });
+  }
```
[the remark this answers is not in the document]
schilchSICKAG: Ha, good catch. I think this will have to be adjusted as most frameworks may set via `property` and the formatting would therefore maybe not work.

packages/components/src/components/input/formatter.ts
kirchsuSICKAG: you're right 👍

packages/components/src/components/input/input.component.ts
kirchsuSICKAG: you're right 👍

packages/components/src/components/input/input.component.ts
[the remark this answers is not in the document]
schilchSICKAG: I have updated the functionality to apply auto-format when changing via `attribute`. For the `property`, I don´t see an easy way as we would have to adjust the whole logic for this to work properly.

packages/components/src/components/input/input.component.ts
```diff
@@ -218,6 +227,84 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+  /**
+   * Optional options that should be passed to the `NumberFormatter` when formatting the value.
+   * This is used to format the number when the input type is `number`.
+   * Note this can only be set via `property`, not as an `attribute`!
+   */
+  @property({
+    attribute: false,
+    reflect: false,
+    type: Object,
+  }) numberFormatterOptions: Intl.NumberFormatOptions;
```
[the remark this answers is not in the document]
schilchSICKAG: As discussed we will currently leave this as is.

packages/components/src/components/input/input.component.ts
```diff
@@ -305,11 +412,77 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+        // If neither min nor max are set, we set the value to 0
+        const { max, min } = this;
+
+        if (max !== undefined && max !== null) {
+          initialNextValue = typeof max === 'string' ? parseFloat(max) : +max;
+        } else if (min !== undefined && min !== null) {
+          initialNextValue = typeof min === 'string' ? parseFloat(min) : +min;
+        } else {
+          initialNextValue = 0;
+        }
```
kirchsuSICKAG, line 466: Don't we also need this check and setting invalid values to "min", "max" or "0" in the `attributeChangedCallback`?

kirchsuSICKAG (changes requested):

packages/components/src/components/input/input.component.ts
```diff
@@ -305,11 +412,77 @@ export default class SynInput extends SynergyElement implements SynergyFormContr
...
+
+    return {
+      eventObj,
+      shouldClamp: !!eventObj,
+      nextValue,
+    };
+  }
+
   private handleChange() {
+    if (this.type === 'number' && (this.#isNumberFormattingEnabled() || this.#numericStrategy.autoClamp)) {
```
kirchsuSICKAG, line 450: Currently we will do the formatting not only if the number formatting is enabled, but also when the auto-clamping strategy is enabled. Is this correct or do we only want to format and do the invalid handling stuff on number formatting enabled?

Also because the check was augmented with `this.#isNumberFormattingEnabled()`, the clamp event is now also fired if `autoClamp` is NOT enabled, but e.g. a min-fraction-digit is set:

image

Example code: 

```html
  <syn-input
      label="Native (min and max)"
      max="10"
      min="0"
      numeric-strategy="native"
      type="number"
      value="0"
      step="0.3"
      min-fraction-digits="3"
    ></syn-input>
   
    <script type="module">
      const inputs = document.querySelectorAll('syn-input');
      inputs.forEach((input) => {
        console.log(input);
        input.addEventListener('syn-clamp', (event) => {
          console.log('syn-clamp event:', event.detail);
          console.log(input);
        });
      });
      </script>
```

kirchsuSICKAG (changes requested):

packages/components/src/components/input/input.component.ts
schilchSICKAG: I see what I did there :). Fixed in latest

packages/components/src/components/input/input.component.ts
schilchSICKAG: As discussed, I opted to remove the check of `setAttribute` as it is inconsistent with setting the `value` via `property`. Only user-input will get formatted from now on.

PR #855, opened 2025-04-30
feat: ✨ Add variant="sticky" to syn-side-nav

### LINKED ISSUE #581: feat: ✨ Create template for small-fixed width syn-side-nav navigation
# Component

I suggest to add the feature to the following component: syn-side-nav

# Story

As a User of the Synergy Design System, I would like to add a Template for a fixed-width syn-side-nav that the Synergy Users see what is possible with our Components.

# Acceptance criteria

- [ ] Design Ticket: https://github.com/synergy-design-system/design/issues/142

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR:
- adds a new property `variant` to the syn-side-nav
Variant has now 3 possible values:
            - `fixed`: current `<syn-side-nav />`. This will also be the default
            - `rail`:  current `<syn-side-nav rail />`. The `rail` attribute is deprecated with this PR
            - `sticky`:  new variant type with a sticky side-nav
- exchanges all usages of `rail` attribute with `variant="rail"`
- deprecates the `rail` attribute (with hints for the deprecation)

### 🎫 Issues
Closes #581.

## 👩‍💻 Reviewer Notes
- I got errors in the test for `<syn-side-nav variant="sticky"/>` with "ResizeObserver loop completed with undelivered notifications.". This error does not occur for  `<syn-side-nav variant="sticky" open/>`.   The resize observer callback in the nav-items retriggers a render cycle, because of the "this.showPrefixOnly". Via "requestAnimationFrame" we get rid of it. See also: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#observation_errors. 

- have a look at the new english and german translations, if they fit

## 📑 Test Plan
- Have a look at the new variant property and check if everything works fine for the "old" variants `rail` and `fixed``
- have a look at the new variant `sticky` if everything works as expected

## ✅ DoD


reichst1SICKAG:
`Deprecate "rail" attribute and introduce "variant" attribute`

-> This will lead to breaking changes and a new major version right?

reichst1SICKAG:
I would opt to rename the variant `fixed` to `full` or something similar.

1. For me `fixed` is to similar to `sticky` and relates more to a position (scrolling) than a version. 
Also might be confusing with css position: fixed/sticky ? 
In the ticket #581  we wrote [...] add a property for a fixed-width syn-side-nav [...]

2. 'full' sounds also better when the navigation is opened via burger-icon or as the rail.open version. 
So we have `full`, `rail`, and `sticky/fixed-width` versions.

reichst1SICKAG:
Is `sticky` really a variant or is it rather an additional option for `rail`?

kirchsuSICKAG:
> Is `sticky` really a variant or is it rather an additional option for `rail`?
> 
> ```
> 
> `full`
> 	open      -> full
> 	closed    -> invisible
> `rail`
> 	open	-> full+open
> 	closed	-> rail
> 
> `sticky`
> 	open 	-> rail+open   + extra footer icon
> 	closed	-> rail+closed + extra footer icon
> ```

@reichst1SICKAG Yes it needs to be a different options as:
- otherwise we would need another property for showing the toggle icon in the footer and if another variant will get asked for some day, we get a property hell
- also the sticky variant has another handling how it opens than the rail variant. The rail opens on hover and closes when going out of it, while the sticky just stays open or close on hover and going out

packages/components/src/components/side-nav/side-nav.component.ts
```diff
@@ -327,29 +345,63 @@ export default class SynSideNav extends SynergyElement {
     }
   }
 
+  // eslint-disable-next-line complexity
+  protected override willUpdate(changedProperties: PropertyValues) {
+    super.willUpdate(changedProperties);
+
+    // TODO: this can be removed in synergy version 3.0
+    if (changedProperties.has('rail')) {
+      if (this.rail && process.env.NODE_ENV !== 'production') {
```
schilchSICKAG, line 354: This won´t work because we already create a build for the components via `pnpm build`. When creating the build, it will set `process.env.NODE_ENV` to `production` internally. This means this works in Storybook (because we don´t set the flag there and don´t use the production build), but won´t work in projects using the package directly (e.g. try to add it to one of our demo projects). Did not think about this when I proposed using it. I think we will have to find another way - or don´t check at all for a precondition but always display the error.

packages/components/src/components/side-nav/side-nav.component.ts
```diff
@@ -364,8 +416,15 @@ export default class SynSideNav extends SynergyElement {
...
           <footer class="side-nav__footer" part="footer-container" slot="footer">  
 
-            ${hasFooter ? html`<syn-divider part="footer-divider" class="side-nav__footer-divider"></syn-divider>` : ''}
-            <slot name="footer" part="footer" ></slot> 
+            ${showFooterDivider ? html`<syn-divider part="footer-divider" class="side-nav__footer-divider"></syn-divider>` : ''}
+            <slot name="footer" part="footer" ></slot>
+            ${this.variant === 'sticky'
+              ? html`<syn-nav-item part="toggle-nav-item" class="side-nav__toggle-nav-item" @click=${this.toggleOpenState} ?divider=${hasFooter}>
+                      <syn-icon slot="prefix" library="system" name="sticky_sidebar" part="toggle-icon"></syn-icon>
+                      ${this.open ? this.localize.term('sideNavHide') : this.localize.term('sideNavShow')}
```
schilchSICKAG, line 424: I notice a small visual irregularity because the animation when closing will start and the text will get changed at the same time. Would it be possible to delay the switch of the localization (e.g. via animationEnd)? I am not sure about this, but we could use a state variable (e.g. this.navHideLabel or something like this?) and just toggle it after the animation is done to make it a bit smoother.

Notice that the text is not (visually) available at all when the item is closed, so it should not matter that much. Not sure if I am a bit picky here, though ;)

schilchSICKAG (changes requested):
I think this works really well. Some very minor nitpicks and possible improvements and we should be good to go.

packages/components/src/components/side-nav/side-nav.component.ts
kirchsuSICKAG: You are absolutely right. I tested it in our demo and it doesn't work 😕 I searched again a bit for a solution of lit itself to show warnings or be able to hook up into their warning mechanism, but I couldn't find anything. Therefore I removed the check for production now

packages/components/src/components/side-nav/side-nav.component.ts
kirchsuSICKAG: is fixed now

packages/components/src/components/side-nav/side-nav.styles.ts
```diff
@@ -8,21 +8,6 @@ export default css`
     display: block;
   }
 
-  /* Side nav should have a width, so it behave correct in an e.g. display flex context */
```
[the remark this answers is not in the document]
kirchsuSICKAG: Unfortunately it is not that easy.
for the fixed / default navigation it would not be enough to add the position: relative. And adding a display:block to it would result in moving the other content somewhere else:

side-nav-default

The sticky variant also has problems. There a position: relative would be enough, but the mobile handling would no longer work:

side-nav-sticky-relative

As this would take some more time to do it correctly, I would let it as it is and maybe do it at a later point.

@schilchSICKAG what do you mean?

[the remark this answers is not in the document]
schilchSICKAG: I think in this case we should leave it as it is currently is. It was just a thought, so I am fine with the current solution, too.

packages/docs/stories/templates/appshell.stories.tsx
```diff
@@ -211,6 +211,71 @@ export const RailNavigationMobile: Story = {
   },
 };
 
+export const StickyNavigationDesktop: Story = {
+  name: 'Sticky Navigation (Desktop)',
+  parameters: {
+    docs: {
+      description: {
+        // TODO: add new sticky description as soon as doc token is available
+        story: generateStoryDescription('application-shell', 'rail-navigation-desktop', 'templates'),
```
schilchSICKAG, line 220: I think this uses the wrong documentation token

schilchSICKAG (changes requested):
One last round and we should be good to go 🚀

packages/tokens/src/figma-tokens/$themes.json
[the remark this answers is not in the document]
kirchsuSICKAG: Good question. I have no idea, where this comes from. @reichst1SICKAG do you know, where this comes from or if this is problematic?

packages/docs/stories/templates/appshell.stories.tsx
kirchsuSICKAG: Updated to the new doc token

packages/components/src/components/side-nav/side-nav.component.ts
```diff
@@ -327,9 +356,41 @@ export default class SynSideNav extends SynergyElement {
     }
   }
 
+  // eslint-disable-next-line complexity
+  protected override willUpdate(changedProperties: PropertyValues) {
```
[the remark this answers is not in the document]
kirchsuSICKAG: your right, this could be shortened, but unfortunately we need to do it a bit different. We also need to go into the first if-statement if `rail === false` so we can set the `variant` property accordingly if someone is just using the rail property to set / unset the side-nav to another variant. I adopted the statement a bit. Is it better like this @schilchSICKAG?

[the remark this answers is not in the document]
schilchSICKAG: Yes, I think its fine this way :)

packages/tokens/src/figma-tokens/$themes.json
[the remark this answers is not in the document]
reichst1SICKAG: This is an internal tokens studio setting. Will have no impact

packages/components/src/components/side-nav/side-nav.component.ts
```diff
@@ -9,12 +9,14 @@ import componentStyles from '../../styles/component.styles.js';
 import styles from './side-nav.styles.js';
 import SynDrawer from '../drawer/drawer.component.js';
 import SynDivider from '../divider/divider.component.js';
+import SynIcon from '../icon/icon.js';
```
schilchSICKAG, line 12: We should import from `icon.component` here instead to make sure the `dependency` function of the item works correctly. Otherwise, it would always define the element.

schilchSICKAG (changes requested):

schilchSICKAG (approved):

reichst1SICKAG (approved):

PR #865, opened 2025-05-16
fix: 🐛 syn-side-nav does not correctly reappear after double-click on burger menu

### LINKED ISSUE #587: fix: 🐛 syn-side-nav does not correctly reappear after double-click on burger menu

# Version

The bug occurred with the following version of the component library: 
2.4.3

# Component

The issue arises when using the following component(s): 
syn-side-nav
(in combination with "burger menu button" in syn-header)

# Environment

The issue occurred with the following operating system(s): tested on windows, linux
The issue occurred in the following browser(s): tested in chrome 127 and firefox 129

# Description

With the sidebar visible, quickly double clicking the "burger menu button" results in wrong behaviour of the sidebar.
The "burger menu" indicates that the sidebar should be visible, and the sidebar "takes space" on the website, but no syn-nav-items are shown.
This can be seen and reproduced on all the stories that have a "burger menu" on https://synergy-design-system.github.io/?path=/docs/components-syn-side-nav--docs

nav-bug-example

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue with `<syn-header>` that lead to animation issues when clicking the burger menu icon too fast.

### 🎫 Issues

Closes #587

## 👩‍💻 Reviewer Notes

I just added a small status variable to check if the animation of the linked `<syn-side-nav>` is already finished. If it is not, the click on the item will do nothing.

I was also thinking about adding something like a `cursor: wait;` to the item in this case, but did not want to make the private property reactive, which would have caused rerenders on finish. Would be possible by using `@state isSideNavAnimating` instead of a private variable.

## 📑 Test Plan

Just click the burger menu icon as fast as you can :). It should break in our current implementation, but ignore the subsequent clicks and not toggle the icon when the animation takes place.

## ✅ DoD

### DIFF
--- a/packages/components/src/components/header/header.component.ts
+++ b/packages/components/src/components/header/header.component.ts
@@ -61,6 +61,11 @@ export default class SynHeader extends SynergyElement {
    */
   private mutationObserver: MutationObserver;
 
+  /*
+   * #587: If the side nav is animating, we should not allow to trigger the burger menu
+   */
+  private isSideNavAnimating = false;
+
   /**
    * The headers label. If you need to display HTML, use the `label` slot instead.
    */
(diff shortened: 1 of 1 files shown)


kirchsuSICKAG (approved):
Nice one :)

PR #870, opened 2025-05-21
feat: ✨ Make angular ngModel event changeable

### LINKED ISSUE #808: feat: ✨ make angular ngModel event changeable

# Story

As a user of the synergy angular wrapper, I want to have the possibility to change the trigger event of the `ngModel`. This is also possible with normal anugular component with ngModel via the ngModelOptions 
property like following:

```js
<input [(ngModel)]="myValue" [ngModelOptions]="{updateOn: 'blur'}" />
```

As our angular wrapper implement their own ngModel the solution via ngModelOptions does not work. We need to adapt the wrappers like following:

```js
  private modelSignal = new AbortController();

  @Input()
  set ngModelUpdateOn(v: 'syn-input' | 'syn-change' | 'syn-blur' | keyof HTMLElementEventMap) {
    this.modelSignal.abort();
    this.modelSignal = new AbortController();
    const option = v || 'syn-input';
    this.nativeElement.addEventListener(option, () => {
      this.valueChange.emit(this.value);
    }, {
      signal: this.modelSignal.signal,
    });
  }
  get ngModelUpdateOn(): { updateOn: string } {
    return this.ngModelUpdateOn;
  }

```

This allows to change the trigger event of the ngModel like following: 

```html
<syn-input [(value)]="myValue"  ngModelUpdateOn="syn-change"></syn-input>
```

# Acceptance criteria

- [ ] The ngModel trigger event is changeable

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds the ability to change the event that triggers the "valueChange" event / the ngModel value

### 🎫 Issues
Closes #808 

## 📑 Test Plan

Check out the new `NgModelUpdateOn` tab in the allcomponents page in angular.

## ✅ DoD

### DIFF
--- a/packages/_private/angular-demo/src/AllComponentParts/NgModelUpdateOn.ts
+++ b/packages/_private/angular-demo/src/AllComponentParts/NgModelUpdateOn.ts
@@ -1,0 +1,29 @@
+import { Component } from '@angular/core';
+import { SynCheckboxComponent, SynInputComponent, SynRangeComponent } from '@synergy-design-system/angular';
+
+@Component({
+  selector: 'demo-ng-model-update-on',
+  standalone: true,
+  imports: [
+    SynInputComponent,
+    SynCheckboxComponent,
+    SynRangeComponent,
+  ],
+  template: `
+    <syn-input data-testid="ng-model-input" [(value)]="inputValue" ></syn-input>
+    <syn-input data-testid="ng-model-input-changed" [(value)]="inputValue" ngModelUpdateOn="syn-change"></syn-input>
+    <syn-checkbox data-testid="ng-model-checkbox" [(checked)]="checkboxValue"></syn-checkbox>
+    <syn-checkbox data-testid="ng-model-checkbox-changed" [(checked)]="checkboxValue" ngModelUpdateOn="syn-blur"></syn-checkbox>
+    <syn-range data-testid="ng-model-range" [(value)]="rangeValue"></syn-range>
+    <syn-range data-testid="ng-model-range-changed" [(value)]="rangeValue" ngModelUpdateOn="syn-blur"></syn-range>
+
+    <div data-testid="ng-model-result-input">{{inputValue}}</div>
+    <div data-testid="ng-model-result-checkbox">{{checkboxValue}}</div>
+    <div data-testid="ng-model-result-range">{{rangeValue}}</div>
+  `,
+})
+export class NgModelUpdateOn {
+  inputValue = 'Hello';
+  rangeValue = '50';
+  checkboxValue = false;
+}
--- a/packages/_private/e2e-demo-test/src/Angular.spec.ts
+++ b/packages/_private/e2e-demo-test/src/Angular.spec.ts
@@ -1,0 +1,88 @@
+import { expect, test } from '@playwright/test';
+import { AllComponentsPage } from './PageObjects/index.js';
+import {
+  fillInputWithoutBlur,
+  getFrameworkPort,
+} from './helpers.js';
+
+const angularPort = getFrameworkPort('angular');
+
+test.describe('<Angular /> on port 5176', () => {
+  // #808 feature ngModelUpdateOn
+  test.describe('ngModelUpdateOn feature', () => {
+    test('Default values for ngModelUpdateOn', async ({ page }) => {
+      const AllComponents = new AllComponentsPage(page, angularPort);
+      await AllComponents.loadInitialPage();
+
+      await AllComponents.activateItem('ngModelUpdateOnLink');
+      await expect(AllComponents.getLocator('ngModelUpdateOnContent')).toBeVisible();
+
+      const input = await AllComponents.getLocator('ngModelUpdateOnInput');
+      const inputResult = await AllComponents.getLocator('ngModelUpdateOnInputResult');
+      const checkbox = await AllComponents.getLocator('ngModelUpdateOnCheckbox');
+      const checkboxResult = await AllComponents.getLocator('ngModelUpdateOnCheckboxResult');
+      const range = await AllComponents.getLocator('ngModelUpdateOnRange');
+      const rangeResult = await AllComponents.getLocator('ngModelUpdateOnRangeResult');
+
+      await expect(inputResult).toHaveText('Hello');
+      await expect(checkboxResult).toHaveText('false');
+      await expect(rangeResult).toHaveText('50');
+
+      // Fill the input without triggering the syn-change event.
+      // Only the syn-input event is triggered
+      await fillInputWithoutBlur(input, 'World');
+      // Check directly for the input result as doing an action with another element,
+      // will result in a blur of the syn-input and therefore a syn-change event
+      await expect(inputResult, 'NgModel value of syn-input should have changed on syn-input event').toHaveText('World');
+
+      await checkbox.click();
+      await expect(checkboxResult, 'NgModel value of syn-checkbox should have changed on syn-input event').toHaveText('true');
+
+      await range.dragTo(range, {
+        targetPosition: { x: 0, y: 18 },
+      });
+      await expect(rangeResult, 'NgModel value of syn-range should have changed on syn-change event').toHaveText('0');
+    });
+
+    test('Non-Default values for ngModelUpdateOn', async ({ page }) => {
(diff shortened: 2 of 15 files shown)


packages/components/scripts/jobs/angular/createComponents.js
```diff
@@ -105,12 +94,38 @@ const componentsCustomization = {
...
+    this.nativeElement.addEventListener(option, () => {
+      ${changeEmitter}
+    }, {
+      signal: this.modelSignal.signal,
+    });
+  }
+  get ngModelUpdateOn(): keyof HTMLElementEventMap {
+    return this.ngModelUpdateOn;
+  }`;
+}
```
schilchSICKAG, line 120: The linter is unhappy :). We are missing a semicolon here.

packages/components/scripts/jobs/angular/createComponents.js
```diff
@@ -156,17 +173,21 @@ export const runCreateComponents = job('Angular: Creating components', async (me
         
       public nativeElement: ${component.name};
       private _ngZone: NgZone;
+      ${getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix) ? 'private modelSignal = new AbortController();' : ''}
```
schilchSICKAG, line 176: Just an idea: We could have less complexity by adjusting the following lines:

```javascript
${getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix) ? 'private modelSignal = new AbortController();' : ''}
${getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix) ? `this.ngModelUpdateOn = '${twoWayBindingEvent}'` : ''}
${getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix) ? getNgModelUpdateOnInput(component.tagNameWithoutPrefix) : ''}
```

```javascript
// Somewhere above
const getTwoWayBindingData = component => {
  if (!getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix)) {
    return {
      controller: '',
      modelUpdater: '',
      method: '',
    };
  }

  const twoWayBindingEvent = getEventAttributeForTwoWayBinding(component.tagNameWithoutPrefix);
  return {
    controller: "private modelSignal = new AbortController();' : ''",
    modelUpdater: "this.ngModelUpdateOn = '${twoWayBindingEvent}'",
    method: getNgModelUpdateOnInput(component.tagNameWithoutPrefix),
  };
};
```

You could then use it like this:

```javascript
// runCreateComponents
// Old
const twoWayBindingEvent = getEventAttributeForTwoWayBinding(component.tagNameWithoutPrefix);

// New
const { controller, modelUpdater, method } = getTwoWayBindingData(component);

// Old
${getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix) ? 'private modelSignal = new AbortController();' : ''}
// New
${controller}

// Old
${getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix) ? `this.ngModelUpdateOn = '${twoWayBindingEvent}'` : ''}
// New
${modelUpdater}

// Old
${getIsTwoWayBindingEnabledFor(component.tagNameWithoutPrefix) ? getNgModelUpdateOnInput(component.tagNameWithoutPrefix) : ''}
// New
${method}
```

This would lead to less complexity in the output. However, this is just an idea, we would also have to come up with a better naming.

packages/_private/angular-demo/src/AllComponentParts/NgModelUpdateOn.ts
schilchSICKAG, line 1: I would be a bit happier if this would not be part of "AllComponents". The reason for this is:

- The amount of items is not sync anymore
- Its not clear on first view that this is the case
- There is no clean separation between framework specific and agnostic stuff

Recommendation:

I think its quite nice to have those framework specific tests. However, I would tend to add a new page in the main sidenav and the tests from there. Visual example is attached. What do you think?



You could theoretically still reuse the AllComponents, however this would have to be refactored so the stories that are loaded are a property rather than a direct import like currently.

In this case, you could:

```typescript
// Remove me from all-components.components.ts
import * as DemoImports from '../AllComponentParts/index.js';
const Demos = Object.entries(DemoImports);
const activeDemo = Demos.at(0)?.at(0);

// Add me to the component:
Demos: <T> = {};
```

Then, where rendering all-components you could add the imports removed from the top of the file and provide it as [Demos]="Demos". This would allow the same structure, but make the separation more clear.

What do you think?

packages/_private/angular-demo/src/AllComponentParts/NgModelUpdateOn.ts
```diff
@@ -0,0 +1,29 @@
...
+@Component({
+  selector: 'demo-ng-model-update-on',
+  standalone: true,
+  imports: [
+    SynInputComponent,
+    SynCheckboxComponent,
+    SynRangeComponent,
+  ],
+  template: `
+    <syn-input data-testid="ng-model-input" [(value)]="inputValue" ></syn-input>
```
schilchSICKAG, line 13: We should add the issue-id to the test-ids as we did on the other tests.

packages/_private/e2e-demo-test/src/Angular.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+import {
+  fillInputWithoutBlur,
+  getFrameworkPort,
+} from './helpers.js';
+
+const angularPort = getFrameworkPort('angular');
+
+test.describe('<Angular /> on port 5176', () => {
+  // #808 feature ngModelUpdateOn
+  test.describe('ngModelUpdateOn feature', () => {
```
schilchSICKAG, line 12: Maybe we should use our standardized describe, like `test.describe('Feature#808: ngModelUpdateOn feature', () => {`

packages/_private/e2e-demo-test/src/Angular.spec.ts
```diff
@@ -0,0 +1,94 @@
+import { expect, test } from '@playwright/test';
+import { AllComponentsPage } from './PageObjects/index.js';
+import {
+  fillInputWithoutBlur,
+  getFrameworkPort,
+} from './helpers.js';
+
+const angularPort = getFrameworkPort('angular');
+
+test.describe('<Angular /> on port 5176', () => {
```
schilchSICKAG, line 10: You have received the port via `getFrameworkPort`, why not use it here: `test.describe(`<Angular /> on port ${angularPort}`, () => {`

packages/_private/e2e-demo-test/src/helpers.ts
```diff
@@ -31,6 +32,21 @@ export const fillField = async (locator: Locator, value: string, selector = 'inp
...
 
+/**
+ * Set the locators inner inputs value.
+ * The elements is NOT blurred after the value is set, so only a syn-input is triggered.
+ *
+ * @param locator The original form locator
+ * @param value The value to set
+ * @param selector The selector to use in the shadow root
+ */
+export const fillFieldWithoutBlur = async (locator: Locator, value: string, selector = 'input') => {
```
schilchSICKAG, line 43: I think we could get rid of those methods by adjusting `fillField`:

```typescript
export const fillField = async (locator: Locator, value: string, selector = 'input')
export const fillField = async (locator: Locator, value: string, selector = 'input', blur = true)
```

schilchSICKAG (changes requested):

packages/_private/angular-demo/src/AllComponentParts/NgModelUpdateOn.ts
kirchsuSICKAG: I created a demo-template, where the demos can be added via router data

packages/_private/e2e-demo-test/src/helpers.ts
kirchsuSICKAG: fixed

packages/components/scripts/jobs/angular/createComponents.js
kirchsuSICKAG: fixed

packages/components/scripts/jobs/angular/createComponents.js
kirchsuSICKAG: fixed

schilchSICKAG (approved):
Nice! Thanks for the updates, I think its much cleaner now.

PR #886, opened 2025-06-02
chore: 🔧 Integrate turbosnap

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR enables turbosnap

### 🎫 Issues
Closes https://github.com/synergy-design-system/Internal/issues/65 

## 👩‍💻 Reviewer Notes

I needed to remove the components import in the `preview.ts` file and add the single components import in each story.
This is needed as any change of an import in the `preview.ts` file will result in a FULL rebuild, instead of only the stories, which are affected by a change.

I added exceptions for following packages, so they won't trigger any full rebuilds on changes:
- packages/_private/**
- packages/angular/**
- packages/react/**
- packages/vue/**

I added explicit files / folders so they will always trigger a full rebuild on change: 
- packages/tokens/src/figma-tokens/semantic/**
- packages/tokens/src/figma-tokens/globals.json 
- packages/tokens/src/figma-tokens/color/**
- packages/assets/src/**/*.svg
- packages/styles/src/**/*.css

Please have a look at these exceptions for triggering and not triggering a full rebuild, if they make sense for you or if maybe something is missing.

For a complete list of checks I did, have a look at this issue: https://github.com/synergy-design-system/Internal/issues/65

## 📑 Test Plan
Have a look at the exceptions for triggering (**externals**) and not triggering (**untraced**) a full rebuild, if they make sense for you or if maybe something is missing.

## ✅ DoD


packages/docs/chromatic.config.json
```diff
@@ -1,6 +1,20 @@
 {
+  "$schema": "https://www.chromatic.com/config-file.schema.json",
+  "projectId": "Project:64f819d70a69cb8728e06daf",
   "onlyChanged": true,
-  "projectId": "Project:64d9e6c38cfad73a2412a13d",
   "storybookBaseDir": "packages/docs",
-  "zip": true
+  "zip": true,
+  "untraced": [
```
schilchSICKAG, line 7: Won´t turbosnap in CI mode read the `chromatic.config.json`? I ask this because we are duplicating settings here between `quality-check.yml` and this configuration. It would be nicer if there was just one source of truth for the untraced files etc. Think this is possible?

schilchSICKAG (changes requested):
Solid PR. The only concerns I have are

- we are duplicating the settings from quality-check and the chromatic config. I am not sure if we will remember this in the future. I also would have assumed the build should pick up the config automatically.
- For the imports and their requirements, it seems to work fine. However, what we could also try is to use our autoloader that we [inherited from Shoelace](https://shoelace.style/getting-started/installation/#cdn-installation-easiest). Don´t know if this works and is configurable enough, but it would still only need to require the autoloader.js. However, if the dependencies are explicit for the stories, I am fine, too and you may ignore this.

packages/docs/chromatic.config.json
kirchsuSICKAG: No, it shouldn't. As this chromatic.config.json is located at the `packages/docs/` folder and the github action is triggering on the root of our monorepo, he would not find the config. If we want to be able to test chomatic things locally (like pushing local build to chromatic), we need the config. 
What I could do is referencing the config file in the github action, then we should not need to have all the `untraced` and `externals` configs doubled

schilchSICKAG: Thats fine for me. Then just lets reference the config file to keep DRY

PR #900, opened 2025-06-11
chore: 🔧 Fine tune chromatic diffThreshold

### LINKED ISSUE #889: chore: 🔧  Fine tune threshold for pixel comparison in Chromatic
# Description

Currently some changes are not detected by chromatic as change and therefore does not appear as a change, which needs to be verified.
E.g. changing the width of the syn-divider from 1px to 3px result in detected changes for
- combobox
- divider
- select
- ...

__but there are false negatives for
- side-nav
- appshell stories

Large screeshots results in large number of pixels that need to be changed because the diffThreshold is a percentage value.
Making the diffThreshold customized to the size of the screenshot could be fix the issue.

possible idea (adjust the generateScreenshotStory function):

```js
export const generateScreenshotStory = (
  stories: { [key: string]: StoryObj },
  options: number | screenshotStoryOptions = 150,
): StoryObj => {
  // …
  const guessedFullHeightOfStory = Object.keys(stories).length * heightPx;
  const usedThreshold = FORMEL_HIER 
```

# Rational
Even small changes should be trigger detection of visual differences

# Acceptance criteria

- [ ] current false negatives should be detected as change
- [ ] false positives are avoided
- [ ] formular is created which takes the story size into account and applied to each story
- [ ] avoid overload of chromatic quota

kirchsuSICKAG:
# Pull Request

### 🎫 Issues
Closes #889.

## 👩‍💻 Reviewer Notes
In this PR I also fixed some screenshot, which where not shown correctly  or complete. Also I found some missing screenshots of stories, which I added. Therefore we have these changes to accept.
I reviewed them and for me they all look good. But it would be nice if someone could review and accept them.

## ✅ DoD


packages/docs/.storybook/preview.ts
```diff
@@ -81,7 +81,7 @@ const preview: Preview = {
       },
     },
     chromatic: {
-      diffThreshold: 0.50, // Original value is set to 0.63
+      diffThreshold: 0.063, // Original value is set to 0.063
```
schilchSICKAG, line 84: When we use the default, is there a reason we leave it in there at all?

schilchSICKAG (approved):

PR #909, opened 2025-06-30
chore: 🔧 Support figma code connect

### LINKED ISSUE #906: chore: 🔧 Support figma code connect

# Story

As a user using the Synergy Design System, I would like to have the possibility to see example source code for my figma prototypes so that I am able to copy my design as snippets for real projects.

# Acceptance criteria

- [x] A first component (button) is available via Code connect
- [x] The button supports all properties, attributes and slots
- [ ] The button allows to select between web-components, angular, vue and react syntax
- [x] An estimate can be made about automating this process.

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds [figma code connect](https://www.figma.com/code-connect-docs/html/) for all of our components, as well as our icons. This enables live example validation in Figmas dev mode, as well as creating sample applications from figma designs via [Figmas MCP server](https://help.figma.com/hc/en-us/articles/32132100833559-Guide-to-the-Dev-Mode-MCP-Server#:~:text=The%20Dev%20Mode%20MCP%20Server%20brings%20Figma%20directly,with%20data%20sources%20using%20the%20Model%20Context%20Protocol.).

### 🎫 Issues

Closes #906

## 👩‍💻 Reviewer Notes

I initially tried to automate things quite a bit. However, code connect uses a very static approach in its `connect` method. For example, properties cannot be shared across different files.

A set of findings can be seen at the code connect namespace (`packages/components/code-connect/README.md`).

### Implementation Details:

I opted to split the code connect calls into components and icons, as each of our icons has to be a connected file. This is because we do not have a `syn-icon` component in Figma. Therefore, we have roughly 2500 connected components.

Those connectors will automatically get updated via `fetch-assets` in the assets package, but have to be published to figma manually.

### Strategic view

Basically, to get more out of code connect, we would have to rename a couple of layers im Figma, which may (or may not) be breaking changes. This would make it possible to create even better results. I already tried it with Figmas MCP server and the first results where quite nice (e.g. creating our Form template as code was working quite well).

We will have to decide if we want to follow this path. If we want to, here are some considerations:

### Final observations:

- Connecting files is quite easy when you get the hang on it.
- To make your prototypes work, you will have to connect the other files, too! This means a team needs to connect its own components to get the most out of it. The synergy component connection will be transferred, though.
- If we go this path, we may have to adjust the code-connect examples whenever something in figma changes (e.g. when renaming a layer it might not work anymore as Figma Code Connect works on the layer and instance names).
- If we do not want to follow this, we might just call `pnpm exec figma connect unpublish` any time to remove all traces. However, this has to be done in this branch!

## 📑 Test Plan

Just follow the steps in the readme to connect

## ✅ DoD


packages/components/code-connect/components/syn-badge.figma.ts
```diff
@@ -0,0 +1,38 @@
...
+    defaultSlot: figma.string('*value'),
+    variant: figma.enum('variant', {
+      danger: 'danger',
+      neutral: 'neutral',
+      primary: 'primary',
+      success: 'success',
+      text: 'text',
+      'text-inverted': 'text-inverted',
+      warning: 'warning',
+    }),
```
kirchsuSICKAG, line 36: I think here is something mixed with `color` and `variant` property of figma.
image
 I have no idea why we have the color property in figma, but the `variant` property does only include:
 - default
 - success
 - neutral
 - danger
 - warning
 
 and our dev components does only have: 
 - 'primary'
- 'success'
- 'neutral'
- 'warning'
- 'danger'

So doesn't it need to look like this: 

```js
    variant: figma.enum('variant', {
      danger: 'danger',
      default: 'primary',
      neutral: 'neutral',
      success: 'success',
      warning: 'warning',
    }),
```

packages/components/code-connect/components/syn-breadcrumb.figma.ts
```diff
@@ -0,0 +1,35 @@
+import figma, { html } from '@figma/code-connect/html';
+import type SynBreadcrumb from '../../src/components/breadcrumb/breadcrumb.js';
+import type { ValidProperties } from '../core/types';
+
+/**
+ * @todos
+ * - default slot should be visible whenever one of the details is open, not just the first one
```
kirchsuSICKAG, line 7: copy paste from `syn-accordion.figma.ts`

packages/components/code-connect/components/syn-combobox.figma.ts
```diff
@@ -0,0 +1,85 @@
...
+  ],
+  links: [
+    {
+      name: 'Documentation',
+      url: 'https://synergy-design-system.github.io/?path=/docs/components-syn-combobox--docs',
+    },
+  ],
+  props: {
+    clearable: figma.boolean('clearable'),
+    defaultSlot: figma.children('_helper/dropdown/syn-combobox'),
```
kirchsuSICKAG, line 60: There seems to be a problem with the `_helper/dropdown/syn-combobox`: 

image

packages/components/code-connect/components/syn-details.figma.ts
```diff
@@ -0,0 +1,67 @@
...
+      <h3>${subheadline}</h3>
+      ${content}
+    </div>
+  `,
+  imports: [],
+  props: {
+    content: figma.enum('content', {
+      text: figma.string('↳ content'),
+    }),
+    subheadline: figma.boolean('subheadline', {
```
kirchsuSICKAG, line 21: Why does the content need a `figma.enum` but the subheadline needs a `figma.boolean` although both have toggles in figma? 

image

packages/components/code-connect/components/syn-dialog.figma.ts
```diff
@@ -0,0 +1,81 @@
+import figma, { html } from '@figma/code-connect/html';
+import type SynDialog from '../../src/components/dialog/dialog.js';
+import type { ValidProperties } from '../core/types';
+
+/**
+ * @todos
+ */
+
+// Footer helper
```
kirchsuSICKAG, line 9: There seems to be a problem with the `footer helper` and `header helper`

image

packages/components/code-connect/components/syn-input.figma.ts
```diff
@@ -0,0 +1,78 @@
...
+      readonly=${readonly}
+      placeholder=${placeholder}
+      type=${type}
+    >
+      <span slot="prefix">
+        ${prefixSlot}
+      </span>
+      <span slot="suffix">
+        ${suffixSlot}
+      </span>
```
kirchsuSICKAG, line 36: The suffix and prefix slot are not shown correctly: 

image

packages/components/code-connect/components/syn-optgroup.figma.ts
```diff
@@ -0,0 +1,45 @@
...
+    suffixSlot,
+  }) => html`
+    <syn-optgroup
+      disabled=${disabled}
+      style=${style}
+    >
+      <span slot="prefix">${prefixSlot}</span>
+      ${defaultSlot}
+      <span slot="suffix">${suffixSlot}</span>
+    </syn-optgroup>
```
kirchsuSICKAG, line 21: The code of the optgroup is not really correct. 
Normally in dev it would look like this: 
image

but it is shown as this in figma: 
Bildschirmfoto 2025-07-08 um 08 31 09

The `defaultSlot` should be mapped to the `label` prop instead

```suggestion
    <syn-optgroup
      disabled=${disabled}
      style=${style}
      label= ${defaultSlot}
    >
      <span slot="prefix">${prefixSlot}</span>
      <span slot="suffix">${suffixSlot}</span>
    </syn-optgroup>
```

packages/components/code-connect/components/syn-radio.figma.ts
```diff
@@ -0,0 +1,37 @@
...
+    'import { SynRadio } from "@synergy-design-system/components";',
+  ],
+  links: [
+    {
+      name: 'Documentation',
+      url: 'https://synergy-design-system.github.io/?path=/docs/components-syn-radio--docs',
+    },
+  ],
+  props: {
+    defaultSlot: figma.instance('↳ <slot (default)>'),
```
kirchsuSICKAG, line 29: The slot value is not shown correctly: 

image

packages/components/code-connect/components/syn-range.figma.ts
```diff
@@ -0,0 +1,71 @@
...
+      label=${label}
+      size=${size}
+      value=${value}
+    >
+      <span slot="prefix">
+        ${prefixSlot}
+      </span>
+      <span slot="suffix">
+        ${suffixSlot}
+      </span>
```
kirchsuSICKAG, line 29: The prefix and suffix slot are not shown correctly: 
image

packages/components/code-connect/components/syn-range-tick.figma.ts
```diff
@@ -0,0 +1,56 @@
...
+  links: [
+    {
+      name: 'Documentation',
+      url: 'https://synergy-design-system.github.io/?path=/docs/components-syn-range-tick--docs',
+    },
+  ],
+  props: {
+    defaultSlot: figma.boolean('*label', {
+      true: figma.instance('<slot (default)>'),
+    }),
```
kirchsuSICKAG, line 53: The slot content is not shown: 

image

packages/components/code-connect/components/syn-select.figma.ts
```diff
@@ -0,0 +1,88 @@
...
+      multiple=${multiple}
+      open=${open}
+      placeholder=${placeholder}
+      size=${size}
+      value=${value}
+    >
+      <span slot="prefix">
+        ${prefixSlot}
+      </span>
+      ${defaultSlot}
```
kirchsuSICKAG, line 45: The default slot is not shown correct:

packages/components/code-connect/components/syn-side-nav.figma.ts
```diff
@@ -0,0 +1,37 @@
...
+  example: ({
+    defaultSlot,
+    open,
+    variant,
+  }) => html`
+    <syn-side-nav
+      open=${open}
+      variant=${variant}
+    >
+      ${defaultSlot}
```
kirchsuSICKAG, line 16: The slot content is not shown correct:

image

packages/components/code-connect/components/syn-badge.figma.ts
schilchSICKAG: You are right. However, I removed `default` alltogether as it is the default in the component, so it does not pollute the DOM to much.

packages/components/code-connect/components/syn-switch.figma.ts
```diff
@@ -0,0 +1,47 @@
...
+    helpText,
+    size,
+  }) => html`
+    <syn-switch
+      checked=${checked}
+      disabled=${disabled}
+      help-text=${helpText}
+      size=${size}
+    >
+      ${defaultSlot}
```
kirchsuSICKAG, line 20: Slot content is not shown correct: 

image

packages/components/code-connect/components/syn-tab-group.figma.ts
```diff
@@ -0,0 +1,40 @@
...
+    defaultSlot,
+    placement,
+    sharp,
+  }) => html`
+    <syn-tab-group
+      contained=${contained}
+      placement=${placement}
+      sharp=${sharp}
+    >
+      ${defaultSlot}
```
kirchsuSICKAG, line 18: Is it possible to add a `slot="nav"` in the tab-group to the syn-tab? I guess not, only if we always add the `slot="nav"` to the syn-tab, right? 

image

packages/components/code-connect/components/syn-table-cell.figma.ts
```diff
@@ -0,0 +1,60 @@
...
+// Synergy Web Component Connection
+figma.connect('https://www.figma.com/design/bZFqk9urD3NlghGUKrkKCR/Synergy-Digital-Design-System?node-id=22460-32928', {
+  example: ({
+    className,
+    defaultSlot,
+  }) => html`
+    <td
+      class=${className}
+    >
+      ${defaultSlot}
```
kirchsuSICKAG, line 12: The slot content is not shown correct:
image

packages/components/code-connect/components/syn-tag.figma.ts
```diff
@@ -0,0 +1,42 @@
...
+    iconSlot,
+    removable,
+    size,
+  }) => html`
+    <syn-tag
+      removable=${removable}
+      size=${size}
+    >
+      ${iconSlot}
+      ${defaultSlot}
```
kirchsuSICKAG, line 18: The slot content is not shown correct:

image

packages/components/code-connect/components/syn-details.figma.ts
schilchSICKAG: Whoever created the API in Figma chose an `enum` for the content of the helper, see attached screenshot:

Bildschirmfoto 2025-07-08 um 08 53 10

Because of this, `boolean` will not work as the value is technically an enum, even if it just has one option and really does nothing. Therefore I prepared the API to use the provided enum value. I think it would be better to adjust this on the Figma side as we cannot do anything against this in code. Note this is also part of the helper component, not the web-component itself.

packages/components/code-connect/components/syn-dialog.figma.ts
schilchSICKAG: Both are working fine, it is the same API deep render stuff as in combobox.

packages/components/code-connect/components/syn-breadcrumb.figma.ts
schilchSICKAG: I removed the comment.

packages/components/code-connect/components/syn-tooltip.figma.ts
```diff
@@ -0,0 +1,28 @@
...
+
+// Synergy Web Component Connection
+figma.connect('https://www.figma.com/design/bZFqk9urD3NlghGUKrkKCR/Synergy-Digital-Design-System?node-id=16910-25970', {
+  example: ({
+    content,
+  }) => html`
+    <syn-tooltip
+      content=${content}
+    >
+      Content
```
kirchsuSICKAG, line 13: The `Content` text should be removed, as it is set via the `content` property

```suggestion
```

kirchsuSICKAG (changes requested):

packages/components/code-connect/components/syn-optgroup.figma.ts
schilchSICKAG: Ah, now I get it :). The problem with syn-optgoup is twofold:

1. There is now way to map a string, to boolean, so I would have to use `<span slot="label">${defaultSlot}</span>` in there. I now opted to use the `label` property like suggested.
2. The component does not take children as our `<syn-optgroup>` does in code. This means the output will always be incorrect when it is used. There is no way to fix this other than recreate the component in figma from scratch to allow a list of children that we can render. I created a todo comment for this.

packages/components/code-connect/components/syn-tooltip.figma.ts
schilchSICKAG: The question is what we will render here. Content was the wrong word, I agree. However, the figma component does not allow nesting, which our HTML version needs. Therefore, we will have to add some code sample. I opted to do it like this now:

```html
    <syn-tooltip
      content=${content}
    >
      <syn-button>
        Replace with content that should activate the tooltip.
      </syn-button>
    </syn-tooltip>
```

I think this makes it clear that the component that is wrapped by the tooltip must be set. To make it work properly, the component in Figma has to accept children as a `defaultSlot` that we can read, nothing that I can do from code side otherwise :(

packages/components/code-connect/components/syn-tab-group.figma.ts
schilchSICKAG: Yes, cannot do anything against this as the properties are missing and there is no possibility to add **any** logic to the connected components :(.

kirchsuSICKAG (approved):

PR #914, opened 2025-07-07
feat: ✨ Add restricted options to syn-combobox

### LINKED ISSUE #626: feat: ✨ make syn-combobox restrictable to added options

# Component

I suggest to add the feature to the following component: syn-combobox

# Story

As a user of the Synergy Design System, I would like to be able to have a combobox, which does not allow other user input than the added syn-option's.
If something is typed in, which does not fit to one of the options, it should be discarded.
Have a look at [the combobox of vaadin](https://vaadin.com/docs/latest/components/combo-box) to see the desired behavior

# Acceptance criteria

- [ ] the combobox allows setting a property or has an exported utility function, which restricts the combobox to only allow option values

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds the `restricted` property to the syn-combobox. Additionally the selected state is added to the syn-combobox, not only for `restricted`, but for all.

### 🎫 Issues
Closes #626 

## 👩‍💻 Reviewer Notes
- is the property naming  (`restricted`) okay?
- feels the behavior correct?

## 📑 Test Plan

Have a look at the changes and how the new behavior feels like as well as the new stories.

## ✅ DoD

### DIFF
--- a/packages/_private/e2e-demo-test/src/AllComponents/SynCombobox.spec.ts
+++ b/packages/_private/e2e-demo-test/src/AllComponents/SynCombobox.spec.ts
@@ -1,10 +1,11 @@
 import { expect, test } from '@playwright/test';
 import {
+  type SynChangeEvent,
   type SynCombobox,
   type SynSelect,
 } from '@synergy-design-system/components';
 import { AllComponentsPage } from '../PageObjects/index.js';
-import { createTestCases, fillField } from '../helpers.js';
+import { createTestCases, fillField, hasEvent } from '../helpers.js';
 
 test.describe('<SynCombobox />', () => {
   createTestCases(({ name, port }) => {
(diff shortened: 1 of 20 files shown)


packages/_private/e2e-demo-test/src/AllComponents/SynCombobox.spec.ts
```diff
@@ -173,5 +174,50 @@ test.describe('<SynCombobox />', () => {
...
+
+        await fillField(combobox, 'lo', '.combobox__display-input', true);
+        await waitForChange;
+
+        // Check that the displayed value is the text content of the option
+        const displayedValue = await combobox.evaluate((ele: SynCombobox) => ele.displayLabel);
+
+        expect(displayedValue).toEqual('ipsum');
+      });
+    }); // regression#813
```
schilchSICKAG, line 221: I think this should be // feature #626 instead of regression.

packages/_private/e2e-demo-test/src/AllComponents/SynCombobox.spec.ts
```diff
@@ -173,5 +174,50 @@ test.describe('<SynCombobox />', () => {
         expect(resetDisplayedValue).toEqual('Option 1');
       });
     }); // regression#813
+
+    test.describe(`Regression#626: ${name}`, () => {
+      test('should reset the value of a restricted combobox to empty value', async ({ page }) => {
```
schilchSICKAG, line 179: Do you think it makes sense to add a test for remote loading of data, too? Unsure about this, but I think there where errors with combobox when the data loaded was not on the page ininitially

packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -195,6 +199,12 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
   /** The combobox's required attribute. */
   @property({ reflect: true, type: Boolean }) required = false;
 
+  /**
+   * When set to `true`, restricts the combobox to only allow selection from the available options.
+   * Users will not be able to enter custom values that are not present in the list.
+   */
+  @property({ reflect: true, type: Boolean }) restricted = false;
```
schilchSICKAG, line 206: Just a small note about this: We will not be able to set this as a default setting because of the design of boolean properties in html when using `boolean` attributes.

The reason for this is, that every call of this is functionally equivalent:

```html
<!-- all of those variants set restricted to true -->
<syn-select restricted>
<syn-select restricted="something">
<syn-select restricted="restricted">
<syn-select restricted="true">
<syn-select restricted="false">

<!-- Only possibility to set it to false! -->
<syn-select>
```

Because we are `reflecting` the property as an attribute, we could use defaultSettings to set it via `SynCombobox.restricted = true`. However, you would never be able to unset it. If a framework chooses to use the provide attribute for rendering instead of the property, the system will never be able to unset `restricted`.

schilchSICKAG (approved):

packages/_private/e2e-demo-test/src/AllComponents/SynCombobox.spec.ts
kirchsuSICKAG: done

PR #916, opened 2025-07-09
fix: 🐛 syn-validate - Incorrect state of form element if setting customValidationMessage via custom event

### LINKED ISSUE #915: fix: 🐛 syn-validate - Async customValidationMessage change of syn-validate can result in incorrect invalid state of input element

# Version

The bug occurred with the following version of the component library: 2.38.0

# Component

The issue arises when using the following component(s): syn-validate

# Environment

# Description
Setting the customValidationMessage property of syn-validate async in a microtask (e.g. in react via useLayoutEffect), the invalid state of the input element is not set correctly because of timing issues. 
The element is not highlighted with a red border. See:

Image

Example reproduction code: 

```html
<syn-validate on="revalidate">
  <syn-input></syn-input>
</syn-validate>
  
<script type="module">
  const input = document.querySelector('syn-input');
  const validate = document.querySelector('syn-validate');
  input.addEventListener('syn-change', () => {
    validate.customValidationMessage = 'Invalid value';
    input.dispatchEvent(new CustomEvent('revalidate'));
  });
</script>
```

# Acceptance criteria

- [ ] invalid state is shown correctly

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR fixes the problem, that potentially the form element is not shown with red border, if a custom event for the `on` property is used.

### 🎫 Issues

Closes #915 

## 📑 Test Plan

Have a look at the demo pages and tests.

## ✅ DoD

### DIFF
--- a/packages/components/src/components/validate/validate.component.ts
+++ b/packages/components/src/components/validate/validate.component.ts
@@ -270,24 +270,14 @@ export default class SynValidate extends SynergyElement {
     input.focus();
   }
 
-  /**
-   * Triggers a validation run, showing the validation message if needed.
-   */
   // eslint-disable-next-line complexity
-  private validate = (e: Event) => {
-    // Make sure to stop the validate component from going into an endless cycle of triggering
-    if (isInvalidEvent(e.type) && this.variant === 'native' && this.isInternalTriggeredInvalid === true) {
-      this.isInternalTriggeredInvalid = false;
-      return;
-    }
-
-    // Make sure to always prevent the invalid event when not using native validation
-    if (isInvalidEvent(e.type) && this.variant !== 'native') {
-      e.preventDefault();
-      e.stopPropagation();
-    }
-
+  private handleValidate = async (e: Event) => {
     const input = e.currentTarget as HTMLInputElement;
+    if (input instanceof SynergyElement) {
+      // When using a synergy element, we need to wait for it to be ready!
+      // This is needed as the validity state of the element may not be set yet.
+      await input.updateComplete;
+    }
     this.isValid = input.validity?.valid;
 
     // When we are using eager, make sure to skip focus on the first mount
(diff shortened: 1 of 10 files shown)


packages/components/src/components/validate/validate.component.ts
```diff
@@ -270,24 +270,14 @@ export default class SynValidate extends SynergyElement {
...
-      return;
-    }
-
-    // Make sure to always prevent the invalid event when not using native validation
-    if (isInvalidEvent(e.type) && this.variant !== 'native') {
-      e.preventDefault();
-      e.stopPropagation();
-    }
-
+  private handleValidate = async (e: Event) => {
```
schilchSICKAG, line 274: Whats the reason for this new method? Does `validate` not handle validation?

packages/components/src/components/validate/validate.component.ts
```diff
@@ -316,6 +306,27 @@ export default class SynValidate extends SynergyElement {
...
+    }
+
+    // Make sure to always prevent the invalid event when not using native validation
+    if (isInvalidEvent(e.type) && this.variant !== 'native') {
+      e.preventDefault();
+      e.stopPropagation();
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.handleValidate(e);
```
schilchSICKAG, line 327: Maybe it would be cleaner to have `validate` itself be async? As the validate event is internal, it should not be a problem to bind it the same way as before, should it?

schilchSICKAG (commented):
Not sure why the validate functionality was split, otherwise it looks fine.

packages/components/src/components/validate/validate.component.ts
schilchSICKAG: Ok, so basically the problem is that syn validate will now always be of type "Promise<void"> or something like this? I don´t think this is a major problem and would suggest to try to just silence the linter for this line :)

kirchsuSICKAG: Not exactly sure about what it means, but if it will result in a problem we can just fix it afterwards. I`ve undone it

schilchSICKAG (approved):

PR #1034, opened 2025-09-30
fix: 🐛 Synergy element which use syn-popup may break if used in a stacking context

### LINKED ISSUE #849: fix: 🐛 Synergy element which use syn-popup may have a layout break if used in position:relative or position:sticky parents

# Version

The bug occurred with the following version of the component library: 2.30.0

# Component

The issue arises when using the following component(s): syn-select, syn-dropdown, syn-tooltip

# Environment

# Description
If having a synergy element, which uses the syn-popup under the hood (e.g. syn-select, syn-dropdown, syn-tooltip), inside a parent element with e.g. position: sticky or position: relative, it is possible, that the layout will break and the synergy element renders BEHIND other stuff.

See following codepen for the broken layout:
https://codepen.io/kirchsuSICKAG/pen/zxxwdOj

and a gif of the problem:

Image

# Possible fix
Currently our elements use the `@floating-ui/dom` library for the positioning of elements. They have a hint in their documentation about this problem (https://floating-ui.com/docs/misc#clipping) and suggest portals as solution for this problem.

Since 2024 all major browsers support the [Popover api](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API). In webawesome they mixed the `@floating-ui/dom` with the popover api, which results in the problem no longer occurring!

## Following changes need to be done:
- the `hoist` of other elements and the `strategy` attribute  of popup are no longer needed!
- a check if the popover api is supported: ` var SUPPORTS_POPOVER = globalThis?.HTMLElement?.prototype.hasOwnProperty("popover");`
- in method `start` (line 284) of popup.component.ts add `this.popup.showPopover();`
- in method `stop` (line 297) of popup.component.ts add `this.popup.hidePopover();`
- in method `reposition` (line 411) use this code for the strategy
```typescript
    const getOffsetParent =
      SUPPORTS_POPOVER
        ? (element: Element) => platform.getOffsetParent(element, offsetParent)
        : platform.getOffsetParent;

    computePosition(this.anchorEl, this.popup, {
      placement: this.placement,
      middleware,
      strategy: SUPPORTS_POPOVER ? "absolute" : "fixed",
```
- in the `render` method change the popup part to: 
```html
      <div
        part="popup"
        popover="manual"
        class=${classMap({
          popup: true,
          'popup--active': this.active,
          'popup--fixed': !SUPPORTS_POPOVER,
          'popup--has-arrow': this.arrow
        })}
      >
```
- in the `popup.style.ts` add this styling to undo the browser internal popup styles:
```
  .popup {
    /* Clear UA styles for [popover] */
    :where(&) {
      inset: unset;
      padding: unset;
      margin: unset;
      width: unset;
      height: unset;
      color: unset;
      background: unset;
      border: unset;
      overflow: unset;
    }
  }
```

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds a fix for using syn-tooltip in stacking context as outlined in #849 by following the recommendations and using popover api under the hood now.

### 🎫 Issues

Closes #849 

## 👩‍💻 Reviewer Notes

The changes seem to work fine by themself. I also deprecated all hoist and strategy properties of the components, but did not alter the interface. However, popover will now be checked and used if available in the browser environment. Only if it is not available it will use the `strategy` as a fallback. In this case, nothing will change for the user.

## 📑 Test Plan

Have a look at the created stories and affected components (Dropdown, Range, Select, Combobox...).

## ✅ DoD


packages/components/src/components/tooltip/tooltip.component.ts
```diff
@@ -86,7 +86,7 @@ export default class SynTooltip extends SynergyElement {
   @property({ type: Boolean, reflect: true }) disabled = false;
 
   /** The distance in pixels from which to offset the tooltip away from its target. */
-  @property({ type: Number }) distance = 13;
+  @property({ type: Number }) distance = 8;
```
kirchsuSICKAG, line 89: This change is unintended and should be reverted --> needs to be done in the custom vendorism script. See the comment there

packages/components/scripts/vendorism/custom/tooltip.vendorism.js
```diff
@@ -10,11 +12,19 @@ const FILES_TO_TRANSFORM = [
...
   // Get a 4px spacing between nupsi and anchor
-  const content = originalContent.replace(
+  let content = originalContent.replace(
     '@property({ type: Number }) distance = 8;',
     '@property({ type: Number }) distance = 13;',
   );
 
+  // #849: Mark hoist as deprecated
+  content = addSectionAfter(
+    originalContent,
```
kirchsuSICKAG, line 22: The `content` needs to be used there instead of the `originalContent`. Otherwise the first change with the new distance value is not taking action.

```suggestion
    content,
```

kirchsuSICKAG:
The deprecations should also be added to the `html-validate` package in the file `packages/components/code-quality/html-validate/scripts/synergy-element-rules`

kirchsuSICKAG (approved):

PR #1040, opened 2025-10-08
fix: 🐛 Subsequently changed delimiter not taking into account for option value

### LINKED ISSUE #1036: fix: 🐛 Delimiter for Select is not always working
# Version

The bug occurred with the following version of the component library:
    "@synergy-design-system/angular": "2.34.3",
    "@synergy-design-system/assets": "~1.15.1",
    "@synergy-design-system/components": "2.34.3",
    "@synergy-design-system/styles": "~1.7.1",
    "@synergy-design-system/tokens": "~2.20.0",

# Component

The issue arises when using the following component(s):
syn-select
syn-option

# Environment

The issue occurred with the following operating system(s):
The issue occurred in the following browser(s):

# Description

# Acceptance criteria

kirchsuSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue where the `<syn-select />` `delimiter` is not taken into account for the select if:

1. it is used in angular and it comes to some timing issues, which result into the options being processed with the default delimiter before the set delimiter is taken into account
2. if the delimiter is changed at a later point

### 🎫 Issues
Closes #1036 

## 👩‍💻 Reviewer Notes
I wanted to add a framework specific angular test with exactly the use case of point 1. But unfortunately it has some very specific circumstances, which it is not easily reproducible with our e2e framework specific test scenarios.

## ✅ DoD

- ~[ ] I have added documentation to the DaVinci migration guide (if need be)~


packages/components/src/components/option/option.component.ts
```diff
@@ -44,6 +44,8 @@ export default class SynOption extends SynergyElement {
   // @ts-expect-error - Controller is currently unused
   private readonly localize = new LocalizeController(this);
 
+  // Needed to store the original value before any delimiter processing
```
schilchSICKAG, line 47: Maybe add the ticket number as a reference here? We did this to most fixes that we have added.

packages/components/scripts/vendorism/custom/option.vendorism.js
```diff
@@ -54,6 +56,22 @@ const transformComponent = (path, originalContent) => {
     ],
   ], originalContent);
 
+  content = addSectionsAfter([
```
schilchSICKAG, line 59: Add a comment why these changes are here, see line 75

packages/_private/vanilla-demo/src/all-components-regressions.ts
```diff
@@ -15,7 +15,20 @@ const getAllComponentsElement = async () => {
   return allComponents;
 };
 
-const appendOptions805 = async (querySelector: string) => {
+const appendOptionsWithSpace = async (querySelector: string) => {
```
schilchSICKAG, line 18: Maybe call this appendOptions1036 to be concise in naming?

schilchSICKAG (approved):
Seams very reasonable, nice find and good solution :). My findings are just some random documentation stuff, you may fix them or leave them be. For me, everything else is fine.

packages/components/src/components/option/option.component.ts
kirchsuSICKAG: done

packages/_private/vanilla-demo/src/all-components-regressions.ts
kirchsuSICKAG: done

PR #1074, opened 2025-11-05
feat: ✨ CD update for syn-menu

### LINKED ISSUE #958: feat: ✨ CD update for syn-menu
# Component

I suggest to add the feature to the following component: syn-menu

# Story

As a developer of the Synergy Design System, I would like to adapt the appearance to match the new CD 2025

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds brand2025 compatibility for `<syn-menu>`, `<syn-menu-item>` and `<syn-menu-label>`. It also adds some new changes to the `brand2025` dark theme.

### 🎫 Issues

Closes #958  

## 👩‍💻 Reviewer Notes

Just follow the regular procedure

## 📑 Test Plan

Just follow the regular procedure

## ✅ DoD


packages/components/src/components/menu-item/menu-item.custom.styles.ts
kirchsuSICKAG: Even if in most cases because of the strange focus behavior of the syn-menu-item, we can't see the hover... 
The hover has in design another color. It should have `--syn-interactive-background-color-hover` instead of the `--syn-color-neutral-1000`, which it currently has

Image

Image

packages/components/src/components/menu-item/menu-item.custom.styles.ts
```diff
@@ -42,6 +42,20 @@ export default css`
     color: var(--syn-color-neutral-0);
   }
 
+  /* Adjust background and text color for focused elements */
+  :host(:focus-visible) .menu-item {
+    background-color: var(--syn-interactive-background-color-focus, var(--syn-color-primary-600));
+    color: var(--syn-interactive-color-focus, var(--syn-color-neutral-0));
```
kirchsuSICKAG: Which interactive token should this be? This token does not exist. It needs to be either `--syn-interactive-emphasis-color-...` or `--syn-interactive-quiet-color-...`

packages/components/src/components/menu-item/menu-item.custom.styles.ts
```diff
@@ -42,6 +42,20 @@ export default css`
...
+  }
+
+  /* stylelint-disable selector-not-notation, plugin/no-unsupported-browser-features */
+  :host(:hover:not([aria-disabled='true'], :focus-visible)) .menu-item,
+  .menu-item--submenu-expanded {
+    background-color: var(--syn-interactive-background-color-hover, var(--syn-color-neutral-100));
+    color: var(--syn-interactive-color-hover, var(--syn-color-neutral-1000));
```
kirchsuSICKAG: Which interactive token should this be? This token does not exist. It needs to be either `--syn-interactive-emphasis-color-...` or `--syn-interactive-quiet-color-...`

packages/components/src/components/menu-item/menu-item.custom.styles.ts
schilchSICKAG: Right and fixed.

packages/components/src/components/menu-item/menu-item.custom.styles.ts
schilchSICKAG: Right and fixed

kirchsuSICKAG (changes requested):

packages/components/src/components/menu-item/menu-item.custom.styles.ts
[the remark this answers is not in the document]
schilchSICKAG: Hmm, got it. Its quite hard to do this right because it acts differently, depending where it is rendered. Sorry for that.

kirchsuSICKAG (approved):

packages/components/src/components/menu-item/menu-item.custom.styles.ts
```diff
@@ -33,15 +49,29 @@ export default css`
...
+  :host(:focus-visible) .menu-item .menu-item__prefix::slotted(syn-icon),
+  :host(:focus-visible) .menu-item .menu-item__suffix::slotted(syn-icon) {
+    color: var(--syn-option-color-active, var(--syn-typography-color-text-inverted));
   }
 
+  /* Adjust background and text color for focused elements */
+  /* stylelint-disable selector-not-notation, plugin/no-unsupported-browser-features */
+  :host(:hover:not([aria-disabled='true'], :focus-visible)) .menu-item,
+  .menu-item--submenu-expanded {
+    background-color: var(--syn-option-background-color-hover, var(--syn-color-primary-1000));
```
kirchsuSICKAG: We should not use the var(--syn-color-primary-1000) here as fallback as it is a new variable and is therefore potentially not there.

packages/components/src/components/menu-item/menu-item.custom.styles.ts
```diff
@@ -7,11 +7,27 @@ export default css`
...
+    border-radius: calc(var(--outline) * 1.5);
+    color: var(--syn-option-color, var(--syn-typography-color-text));
     font-size: var(--syn-font-size-medium);
+    outline: var(--outline) solid var(--syn-panel-background-color);
+    outline-offset: calc(var(--outline) * -1 + 1px);
     padding: var(--syn-spacing-small) var(--syn-spacing-medium);
   }
 
+  :host(:focus-visible) .menu-item {
+    background-color: var(--syn-option-background-color-active, var(--syn-color-primary-1000, var(--syn-color-neutral-1000)));
```
kirchsuSICKAG: I would rather just use 1 fallback token instead of 2, as if the option token is not there the primary-1000 will also not be there (both come with this branch).

The code just looks more complicated

PR #1082, opened 2025-11-11
feat: ✨ CD update for syn-tab, syn-tab-group and syn-tab-panel

### LINKED ISSUE #969: feat: ✨ CD update for syn-tab
# Component

I suggest to add the feature to the following component: syn-tab

# Story

As a developer of the Synergy Design System, I would like to adapt the appearance to match the new CD 2025

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds brand2025 compatibility for `<syn-tab>`, `<syn-tab-group>` and `<syn-tab-panel>`. 

### 🎫 Issues

Closes #969 

## 👩‍💻 Reviewer Notes

There are new icons for the scrollers that had to be included. I noticed there are some other new system icons that we are not using and are **now part of this branch**. As we do not use them, I forwarded this to design as the bundle size grows larger than it has to.

## 📑 Test Plan

Just follow the regular procedure

## ✅ DoD


packages/components/scripts/vendorism/custom/tab-group.vendorism.js
```diff
@@ -159,6 +159,13 @@ const transformComponent = (path, originalContent) => {
       'const precedingTabs = this.tabs.slice(0, this.tabs.indexOf(currentTab));',
     ],
   ], content);
+
+  // #969: Use the new system icons for tabs
+  content = replaceSections([
+    ['name="chevron-right"', 'name="tabs-right"'],
+    ['name="chevron-left"', 'name="tabs-left"'],
```
kirchsuSICKAG: the system-icon `tabs-left` is never used, as we overwrite the `chevron-left` before with `chevron-right`. The result is never using `tabs-left`. 
This is caused by line 25-32.

Imagetab-.

When removing the vendoring and using both system-icon we can remove the rotate styles in the css

packages/components/src/components/tab-group/tab-group.custom.styles.ts
```diff
@@ -5,8 +5,9 @@ export default css`
   /* Write custom CSS here */
   :host {
     --track-width: var(--syn-border-width-small);
-    --track-color: var(--syn-color-neutral-300);
+    --track-color: var(--syn-panel-border-color);
     --indicator-width: var(--syn-border-width-x-large);
+    --indicator-color: var(--syn-interactive-emphasis-color, --syn-color-primary-600);
```
kirchsuSICKAG: The `var()` is missing around the fallback color




```suggestion
    --indicator-color: var(--syn-interactive-emphasis-color, var(--syn-color-primary-600));
```

packages/components/scripts/vendorism/custom/tab-group.vendorism.js
schilchSICKAG: Fixed as discussed. There where dark places in our CSS that could be removed :)

PR #1129, opened 2025-12-08
feat: ✨ add small/large size to syn-alert

### LINKED ISSUE #1119: feat: ✨ add small/large size to syn-alert

# Component

I suggest to add the feature to the following component: syn-alert, syn-validate

# Story

As a user of the Synergy Design System, I would like to have a small and large version of the syn-alert so that it aligns with the other components sizes like input. The existing version is the medium size.

# Acceptance criteria

- [ ] Check syn-validate

schilchSICKAG:
# Pull Request

## 📖 Description

We have added a new property `size` to `<syn-alert>` that can be set to `small`, `medium` (the default) and `large`.
The property can also be used with Synergy `defaultSettings`.

We also adjusted `<syn-validate>` to draw the `size` attribute of its `slotted` `HTMLInputElement`. If a `size` property is found on the rendered input, it will be forwarded to the rendered `<syn-alert>` if you use `variant="inline"`.

### 🎫 Issues

Closes #1119 

## 👩‍💻 Reviewer Notes

I added the new stories and styles. However, I think that the small version and the medium one do differ so slightly its barely noticable.

I also refactored the demo applications to make better reuse of the global settings code as it already diverged. I opted to remove it alltogether for a shared approach in demo-utilities.

Automatic syncing between the input control and the rendered alert was created via a new mutation observer. I initially had this in the original observer, but this created false positives when showing the alert. For this reason, I had to create new logic for this. Tests where also adjusted to accomplish this.

Notice that native HTML Inputs also have a `size` property. However, I opted to exclude this as its an integer that defines the width of an element. We only transfer the size if the rendered element is a `syn-*` element (e.g. instance of SynergyElement).

## 📑 Test Plan

Check the new stories. Also play around with the size attribute. It should work out of the box, even when using `setGlobalDefaults` in the wrappers.

## ✅ DoD


packages/components/src/components/alert/alert.custom.styles.ts
```diff
@@ -43,10 +48,11 @@ export default css`
   .alert__icon,
   .alert__close-button {
     align-items: flex-start;
-    font-size: var(--syn-font-size-x-large);
+    font-size: var(--size-icon-size);
   }
 
   .alert__message {
+    justify-content: center;
```
kirchsuSICKAG: This needs to be `align-self: center;`
Justify-content is not event set into account

Image


```suggestion
    align-self: center;
```

packages/docs/stories/components/validate.stories.ts
kirchsuSICKAG: The height of the "Screenshot" story needs to be adapted. The new story is too big for the old story height.

See:
https://64f819d70a69cb8728e06daf-mdsujyagid.chromatic.com/?path=/story/components-syn-validate--screenshot

Image

PR #1134, opened 2025-12-12
fix: 🐛 syn-option current selection does not match Figma

### LINKED ISSUE #1127: fix: 🐛 syn-option current selection does not match Figma

# Version

The bug occurred with the following version of the component library: 2.69.0

# Component

The issue arises when using the following component(s): syn-option

# Environment

# Description

After gathering feedback with the community, it became clear that the active marker for `<syn-option>` does not quite match the way it is styled in Figma:

- The size of the inset does not match vertically
- There is no possibility to distinguish between keyboard navigation and active items anymore.

# Acceptance criteria

- [x] Refine the keyboard navigation state
- [ ] Make sure the outline matches the wanted dimensions (see Figma)

schilchSICKAG:
# Pull Request

## 📖 Description

This PR provides some minor adjustments for `<syn-option>` and `<syn-menu-item>` hover and focus states to better mimic the effects as detailed in Figma for the SICK 2025 themes.

Both components now use a combination of `background` and `border-radius` to show the `<syn-option>` with an inset highlight color, allowing to better match the wanted spacings.

## Known open Issues:

- [x] `<syn-menu-item>` not yet updated
- [x] `<syn-optgroup>` broken because of changes in the vertical paddings (WIP!)

### 🎫 Issues

Closes #1127 

## 👩‍💻 Reviewer Notes

I was able to achieve this using a background and border-radius for `syn-option`. If the solution is ok like this, I will also try to change `<syn-menu-item>` to use this.

Also, the new solution is closer to Figma for the 2018 theme (the old one used the full with of the items, the new one does not anymore).

I had to calculate the correct position paddings according to the new borders. Also, I think the border-radius might be slightly off for the small version, but we will have to look into this.

> ~~Keep in mind that the solution for `<syn-menu-item>` is NOT YET IN PLACE. I will add this if (and only if) the `<syn-option>` solution is fine for each reviewer!~~

## 📑 Test Plan

Just compare the old and new versions.

## ✅ DoD

### DIFF
--- a/packages/components/src/components/option/option-size.styles.ts
+++ b/packages/components/src/components/option/option-size.styles.ts
@@ -10,7 +10,7 @@ export default css`
    */
   :host([size="small"]) {
     --option-inset-border-horizontal: var(--syn-spacing-2x-small);
-    --option-inset-border-vertical: calc(var(--syn-spacing-2x-small) - 1px);
+    --option-inset-border-vertical: calc(var(--syn-spacing-x-small) - 1px);
     --option-min-height: var(--syn-input-height-small);
     --option-padding: var(--syn-input-spacing-small);
     --option-font-size: var(--syn-input-font-size-small);


packages/components/src/components/option/option-size.styles.ts
```diff
@@ -5,24 +5,32 @@ export default css`
    * Size adjustments for syn-select and other widgets that use syn-option
    * This allows to share the common settings between all components that
    * should be able to set the size of a syn-option tag.
+   *
+   * Also note that we have to adjust the padding to take the border size into account.
    */
   :host([size="small"]) {
+    --option-inset-border-horizontal: var(--syn-spacing-2x-small);
+    --option-inset-border-vertical: calc(var(--syn-spacing-2x-small) - 1px);
```
kirchsuSICKAG: Was it intentional, that `size="small"` has `--syn-spacing-2x-small` for the token `--option-inset-border-vertical`? 
Because both others, medium and large, have `--syn-spacing-x-small`. 

For me small should also have `--syn-spacing-x-small` as otherwise the space to the sides looks too narrow.

## type="small" with  `--syn-spacing-2x-small`

Image

## type="small" with  `--syn-spacing-x-small`

Image

PR #1188, opened 2026-02-17
feat: ✨ syn-header: Highlight the header when it is used as a "sticky" header

### LINKED ISSUE #529: feat: ✨ syn-header: Highlight the header when it is used as a "sticky" header

# Component

I suggest to add the feature to the following component: syn-header

# Story

As a User using an application built with the Synergy Design System, I would like to have a better visual state representation so that I can see the boundary of the `syn-header` when it is fixed to the top.

This could be archived in multiple ways:

1. By creating a new prop `shadow` and attaching a class on it. Only apply this if no `<syn-side-nav>` is found.
2. ~~By using some JavaScript in the component that uses an intersection observer and checks if the item is the page is currently intersecting and applying a class to the `syn-header`.~~

In both cases, it should show a visual highlight when the header is not on its original position anymore, e.g. because the content is scrolling.

# Acceptance criteria

- [x] Define the styling for fixed positions in design and dev
- [x] Define how to add the styling (Will be added via property)
- [ ] Apply the styling
- [ ] Create a new story in Storybook and Figma

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the ability to mark `<syn-header>` as a sticky header via a new `sticky` boolean property.

### 🎫 Issues

Closes #529 

## 👩‍💻 Reviewer Notes

To really see it in action, add the sticky property to one of the template stories or (preferrably) one of the demo projects. The sticky story itself does not include other elements, so it will not be sticky in the overview.

With this update there is also a minor change included for the `page-background` tokens that was left out before the 3.0 release. There are now the following token changes:

--syn-page-background-color: Default neutral-0 (was neutral-50 before)
--syn-page-background-color-muted: New token for muted backgrounds, defaults to neutral-50

## 📑 Test Plan

- [ ] Check functionality in the demos by altering the rendered html
- [ ] Check the functionality in the Storybook AppShell templates

## ✅ DoD


packages/_private/angular-demo/src/app.css
kirchsuSICKAG: Currently we have following css in all our demos for the header: 

```css
syn-header {
  position: sticky;
  top: 0;
  z-index: 50;
}
````

This css is no longer needed and should be removed in favor of the "sticky" attribute in all demos

PR #1193, opened 2026-02-20
feat: ✨ create syn-tag-group [skip chromatic]

### LINKED ISSUE #408: docs: 📚 tag template
# Description

A tag template which shows how several tags are grouped in a parent element

# Rationale

Currently it is only shown how tags are being used in a multi-select. Tags are also used without the select though.  We should also show the usecase where tags are used without the select. A common scenario is one where a clear link deletes all tags at once.

### LINKED ISSUE #1152: feat: ✨ create tag-group as style
# Story

As a user of the Synergy Design System, I would like to have a style for grouping syn-tag's

# Acceptance criteria

- [ ] add styles: definitions for label and layout
- [ ] Template (Search-Filter)

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the new `<syn-tag-group>`, as well as usage examples (templates) in storybook.

### 🎫 Issues

Closes #1152
Closes #408 

## 👩‍💻 Reviewer Notes

Originally we wanted to create styles for the tag-group only. However, we would have had a hard time with custom stylings because not everything lets itself handle as booleans and there is an underlying DOM structure to follow for this component. Because of that, I opted to create a real component out of it.

I also had issues indexing the new component templates into the MCP. I found that for some stories, a mandatory wrapper was missing. The same was true for some other iframe related content that was not indexable (like some stories for combobox). I fixed both problems in this branch and was able to add long missing stories to the mcp index :)

I added the components to all overviews in the framework demos, as well as the htmllint system.

## 📑 Test Plan

Have a look at the component and its behavior, as well as the new form template. Note that the template differs from the Figma-Version. This is because there are multiple things that wen cannot use the way they where intended (e.g. syn-select does not make sense for this filter as the display is fundamentally different). @reichst1SICKAG is aware of this and will refactor the Figma file accordingly.

## ✅ DoD


packages/docs/stories/templates/tag-group.stories.tsx
```diff
@@ -0,0 +1,351 @@
...
+            `)}
+          </div>
+        </syn-details>
+
+        ${currentState.filters.some(filter => filter.selected.length > 0) ? html`
+          <div class="filter-tags">
+            ${currentState.filters
+              .filter(filter => filter.selected.length > 0)
+              .map(filter => html`
+                <syn-tag-group label-position="start" label="${filter.name}">
```
kirchsuSICKAG: Currently the tag-group `label-position` is not switched to `"top"` if the screen size is too small. In figma it switches to `label-position="top"`

kirchsuSICKAG (changes requested):

packages/docs/stories/templates/tag-group.stories.tsx
schilchSICKAG: You are right. Fixed in latest version

kirchsuSICKAG (approved):

PR #1231, opened 2026-03-11
feat: ✨ syn-radio-button

### LINKED ISSUE #1140: feat: ✨ syn-radio-button
# Story

As a user of the Synergy Design System, I would like to use the radio button as an alternative to a normal radio.

see https://shoelace.style/components/radio-button

# Acceptance criteria

- [ ] Add to all demos / e2e

schilchSICKAG:
# Pull Request

## 📖 Description

This release adds the new `<syn-radio-button>` component for single-selection controls with a more prominent, button-like appearance.

Use `<syn-radio>` for classic form input patterns, and `<syn-radio-button>` when the same behavior should feel more like an immediate mode, view, or option switch.

### 🎫 Issues

Closes #1140

## 👩‍💻 Reviewer Notes

This is currently a **wip**:

- [x] Focus styles are not yet available
- [x] Invalid styles are not defined
- [x] Colors are not yet stable (wrong tokens)
- [x] Dark Mode does not work right
- [x] Figma is missing some stories
- [x] Gaps have to be defined (unable to fetch them correctly from Figma) 

=> Wait until further development from @reichst1SICKAG happens.

## 📑 Test Plan

-

## ✅ DoD

- [ ] I have added a changeset, describing my changes (e.g. via `pnpm release.create`).


packages/components/src/components/button/button.styles.ts
```diff
@@ -42,34 +64,48 @@ export default css`
...
   /* When disabled, prevent mouse events from bubbling up from children */
   .button--disabled * {
     pointer-events: none;
   }
 
+  .button__prefix,
+  .button__suffix,
+  .button__caret {
+    font-size: var(--button-affix-size);
+  }
```
kirchsuSICKAG, line 85: The prefix / suffix / caret icon sizes are wrong. 

Current main:

Image


This branch:

Image

Caret current main:

Image

Caret this branch:

Image

packages/components/src/components/radio-button/radio-button.styles.ts
```diff
@@ -1,30 +1,62 @@
...
+  .button.button--readonly:focus-visible {
+    background: var(--syn-input-readonly-background-color-focus);
+    color: var(--syn-readonly-color-text);
+  }
+
+  .button.button--readonly.button--checked {
+    background: var(--syn-readonly-indicator-color);
+    border-color: var(--syn-readonly-border-color);
+    color: var(--syn-color-neutral-0);
   }
```
kirchsuSICKAG, line 91: I think we also need a class "icon-only", so we can adapt the padding of the "button__label" for icon-only use case. 
In design there is not that much padding around the icon:

Image

This branch:

Image

packages/components/src/components/radio-button/radio-button.styles.ts
```diff
@@ -1,30 +1,62 @@
-/* eslint-disable */
 import { css } from 'lit';
-import buttonStyles from '../button/button.styles.js';
 
 export default css`
```
kirchsuSICKAG, line 3: @reichst1SICKAG , @schilchSICKAG 
When I saw the radio-button for the first time a was a bit confused of the appearance.
Is there a reason why the radio-buttons with radio-group look so much different than the button-group wit buttons? 
I would have assumed they look the same, just with the difference that the buttons are getting selected in case of the radio-buttons in radio-group.

Shoelace also has the same appearance for button-group and radio-button.

Is more a general question and if you decided like this I am fine.

kirchsuSICKAG (changes requested):

packages/components/src/components/button/button.styles.ts
schilchSICKAG: Maybe this was tested on a stale target, but at least for me, they show up fine locally and in Chromatic.

packages/components/src/components/radio-button/radio-button.styles.ts
schilchSICKAG: You are right, fixed in latest

packages/components/src/components/radio-button/radio-button.styles.ts
klaarseSICKAG: I agree. Would have expected a similar design.

packages/mcp/metadata/static/components/syn-radio-button/docs.md
```diff
@@ -0,0 +1,361 @@
+## Default
+
+Radios buttons allow the user to select a single option from a group using a button-like control.Radio buttons are designed to be used with radio groups. When a radio button has focus, the arrow keys can be used to change the selected option just like standard radio controls.
```
klaarseSICKAG: the first item in the group cannot be selected with arrow keys

packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -1909,13 +1909,85 @@
       }
     },
     "radio-button": {
-      "description": {
-        "value": "Radios buttons allow the user to select a single option from a group using a button-like control.",
-        "type": "text"
+      "default": {
+        "description": {
+          "value": "Radios buttons allow the user to select a single option from a group using a button-like control.\n\nRadio buttons are designed to be used with radio groups. When a radio button has focus, the arrow keys can be used to change the selected option just like standard radio controls.",
```
klaarseSICKAG:
```suggestion
          "value": "Radio buttons allow the user to select a single option from a group using a button-like control.\n\nRadio buttons are designed to be used with radio groups. When a radio button has focus, the arrow keys can be used to change the selected option just like standard radio controls.",
```

packages/mcp/metadata/static/components/syn-radio-button/docs.md
schilchSICKAG: And it will never be able to. Sorry, I thought I wrote a comment about this. What basically happens is that:

- The default value of the group is set, so the first button is active
- You use keyboard navigation -> Group value changes
- Because of the interactivity between group and button value changes, Storybook resets the docs page, leading to the value of the button being null.

We cannot do anything against that unfortunately without leaving the value prop out of interaction. The docs page is **never** the source of truth.

packages/components/src/components/radio-button/radio-button.styles.ts
schilchSICKAG: This was done and reviewed in design, so its more a question for @reichst1SICKAG and @eileenwegner.

packages/docs/stories/components/radio-button.stories.ts
```diff
@@ -0,0 +1,274 @@
...
+      disable: false,
+    },
+    docs: {
+      description: {
+        story: generateStoryDescription('radio-button', 'default'),
+      },
+    },
+  },
+  render: args => html`
+    <syn-radio-group label="Select an option" name="a" value="1">
```
klaarseSICKAG, line 68: the stories are not consistent with syn-radio. syn-radio stories are without syn-radio-group. instead the syn-radio-group stories make us of syn-radio. I am not sure which is better but at least we should be consistent..

packages/mcp/metadata/static/components/syn-radio-button/docs.md
schilchSICKAG: See https://64f819d70a69cb8728e06daf-zgaxyudqgx.chromatic.com/?path=/story/components-syn-radio-button--default&args=data-syn-button-group__button--focus:!null;checked:!null and watch the arguments in the header to see the problem in action.

packages/docs/stories/components/radio-button.stories.ts
schilchSICKAG: A syn-radio-button cannot exist without a syn-radio-group. Styling and other stuff is always applied from the group down to the radio-button, so it has to be this way.

schilchSICKAG: Further reference: https://shoelace.style/components/radio-button/

packages/docs/stories/components/radio-button.stories.ts
```diff
@@ -0,0 +1,274 @@
...
+    }
+  },
+  render: () => html`
+    <form class="custom-validity">
+      <syn-radio-group label="Select an option" name="a" help-text="This is required" required>
+        <syn-radio-button value="1">Option 1</syn-radio-button>
+        <syn-radio-button value="2">Option 2</syn-radio-button>
+        <syn-radio-button value="3">Option 3</syn-radio-button>
+      </syn-radio-group>
+      <syn-button type="submit" variant="filled">Submit</syn-button>
```
klaarseSICKAG: possible to make a fake submit so there is no real navigation to 404?

packages/mcp/metadata/static/components/syn-radio-button/docs.md
klaarseSICKAG: ah, I remember. if it's only a storybook problem its ok

packages/docs/stories/components/radio-button.stories.ts
schilchSICKAG: Sure. However every story we have is like this, so if we want to be consequent we have to adjust all other 11 stories that we previously created for invalid handling. What shall we do in this case? Just The play function will also not get called so we cannot safely remove the event. We could either:

a) Use a inline deferred script that prevents the form submit event or
b) Leave it as is.

Whats your take @klaarseSICKAG? I think you are right and it would be saver to not submit. Think could be done in approx. an hour or so

packages/mcp/metadata/static/components/syn-radio-button/docs.md
schilchSICKAG: Yes its story only. The problem is, even if I set an empty value property for the syn-radio-button, the radio-group above will get it empty, toggling it from null to the last one to an empty string and make it go out of bounds. This is also only in the first story (as its connected to storybook controls), not on any of the other stories. Very unfortunate, I think the coupling between radio-button/radio-group/radio is unfortunate... Shoelace chose to do the whole heavy lifting in the group, so the "sub-components" are pretty useless in standalone mode :(. Maybe this gets better when switching to WebAwesome, havent had a look yet. But I would expect the radio to be FormAssociated there instead of the group if they did their job right.

packages/docs/stories/components/radio-button.stories.ts
schilchSICKAG: Just fixed this behavior for all invalid stories by applying a new custom decorator that draws the form around them and cancels the submit.

packages/components/src/components/radio-button/radio-button.styles.ts
schilchSICKAG: Clarified with design: This is by intent.

packages/tokens/src/figma-tokens/_docs.json
schilchSICKAG: Fixed

packages/components/src/utilities/defaultSettings/sort.ts
```diff
@@ -10,11 +10,11 @@ const COMPONENT_UPDATE_ORDER: Partial<Record<ComponentNamesWithDefaultValues, {
 }>> = {
   // Container components that depend on base components update last
   SynButtonGroup: {
-    dependsOn: ['SynButton'],
+    dependsOn: ['SynButton', 'SynRadioButton'],
```
kirchsuSICKAG, line 13: I think we can remove this. The syn-radio-button should only be used with the syn-radio-group, shouldn't it?

packages/components/src/components/button/button.styles.ts
kirchsuSICKAG: You are right. Seems like it was a chrome problem on my side again...

packages/components/src/components/radio-button/radio-button.styles.ts
```diff
@@ -1,30 +1,82 @@
...
+   * Override the default unchecked button
+   */
+  .button--text {
+    color: var(--syn-interactive-quiet-color);
+  }
+
+  .button--text:hover:not(.button--disabled):not(.button--readonly) {
+    background: var(--syn-interactive-background-color-hover);
+    color: var(--syn-interactive-quiet-color-hover);
+  }
```
kirchsuSICKAG, line 47: There is only an "pressed / active" state for the checked radio-button but not for the non-checked radio-button.

See gif:

Image

kirchsuSICKAG (changes requested):

packages/components/src/components/radio-button/radio-button.component.ts
kirchsuSICKAG: We also need to update the version before the release

kirchsuSICKAG (changes requested):

packages/components/src/utilities/defaultSettings/sort.ts
schilchSICKAG: I guessed that, too. However, radio-group renders the radio-buttons in a radio-button-group internally. Thats why its here.

packages/components/src/components/radio-button/radio-button.styles.ts
schilchSICKAG: You are right, will fix that

packages/components/src/components/radio-button/radio-button.component.ts
schilchSICKAG: Jepp. This will hopefully be the last commit since we dont know what is released in the meantime

schilchSICKAG: Set to 3.17.0 (next minor)

packages/components/src/utilities/defaultSettings/sort.ts
schilchSICKAG: See https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/src/components/radio-group/radio-group.component.ts#L421

kirchsuSICKAG: oh, i see

kirchsuSICKAG (approved):

PR #1240, opened 2026-03-20
feat: ✨ Pagination component

### LINKED ISSUE #684: feat: ✨ Pagination component

# Component

I suggest to add the feature to the following component:

# Story

As a [PERSONA] of the Synergy Design System, I would like to [REQUIRE] so that [PRODUCT_VALUE].

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds a new `<syn-pagination>` component that can be used to paginate large datasets. It also includes some adjustments for our component e2e demos as outlined in reviewer notes.

### 🎫 Issues

Closes #684 

### The component

When users interact with the pagination, every control is synchronized around a single source of truth: current page, page size, and total items. Clicking first/previous/next/last updates the page and emits a page-changed event, while changing page size recalculates the page so the user stays anchored to roughly the same dataset position instead of being reset unexpectedly. The component intentionally clamps out-of-range values (for example after total items shrink) so navigation never points to invalid pages, which prevents broken states in real data-refresh scenarios.

In full mode, it exposes page-size selection plus an item-range summary; in compact mode, it removes non-essential controls for tighter layouts.

There is one design choice I want to point out: The select and input sizes grow and shrink, depending on the calculated amount of their values (e.g. the select with 5, 10 and 100 will use 3 chars wide select box, if there are 5 maximal pages, the input will have a width of 1). This leads to "hops" when switching the available page sizes, because the page selector input must also be resized. I checked this with design and they dont want to have a fixed size for the select or input, so I think we will have to check if this is a real obstacle.

Icons:

~~Icons are currently NOT part of the default icon set. They seem to not be available there. I will tell design about it.~~

There are four new system icons that where created:

- first-page
- last-page
- next-page
- previous-page

### Header import adjustments

I noticed the header did not follow our pattern for dependencies. The difference is quite subtle, but it was worth the change nevertheless. With this change, the header will wait to be loaded before trying to define the component.

### Demos

I adjusted the demos so they:

- Use the SICK-2025 light theme per default now
- Adjusted the way the multi story stuff is rendered by moving tabs to the right and making them scrollable. With this change, you see more options and the wide screen estate was not used anyway, making it easier on the eye to navigate.

## 📑 Test Plan

Proceed as usual: have a look at Storybook and the framework demos.

## ✅ DoD


klaarseSICKAG:
the responsiveness hides the number entirely
image

schilchSICKAG:
> the responsiveness hides the number entirely image

Good point. We should probably add breaking for smaller screen estates. I don´t have any designs for this, but I think we can talk with @reichst1SICKAG about it.

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  static dependencies = {
+    'syn-divider': SynDivider,
+    'syn-icon-button': SynIconButton,
+    'syn-input': SynInput,
+    'syn-option': SynOption,
+    'syn-select': SynSelect,
+  };
+
+  /**
+   * When true, a divider is displayed between the page navigation controls and the rows-per-page selector.
```
kirchsuSICKAG: The description for the *divider* property does not seem correct for me. As I would understand it from the description, it should be on the position of the red line and not of the green line, where it really is.

Image

Maybe the description can be adapted to something like "When true, a divider is displayed above the pagination."

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  clampPage,
+  getMaxOptionCharCount,
+  getTotalPages,
+  getTotalPagesCharCount,
+} from './utility.js';
+import type { SynInputEvent } from '../../events/events.js';
+import { enableDefaultSettings } from '../../utilities/defaultSettings/decorator.js';
+
+/**
+ * @summary <syn-pagination /> is a component that provides a data-heavy views, combining page navigation, direct page input, and configurable page-size selection in one control.
```
kirchsuSICKAG: I can't quite put my finger on what bothers me, but something about that sentence doesn't feel right for me.
Maybe change it to something like:
"<syn-pagination /> is a component that helps organizing data-heavy views, combining page navigation, direct page input, and configurable page-size selection in one control."

Not sure if *organizing* is the best word her, but I can't think of a better one right now^^

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+ *
+ * @documentation https://synergy-design-system.github.io/?path=/docs/components-syn-pagination--docs
+ * @status stable
+ * @since 3.0.0
+ *
+ * @event syn-pagination-page-changed - Emitted when the current page changes
+ * @event syn-pagination-page-size-changed - Emitted when the page size changes
+ *
+ * @csspart base - The component's base wrapper.
+ * @csspart divider - The divider element separating the page navigation and page size selector.
```
kirchsuSICKAG: Not quite correct description, as it divides the pagination from whatever is above.


```suggestion
 * @csspart divider - The divider element displayed at the top of the pagination component.
```

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+  @property({ reflect: true, type: Boolean }) disabled = false;
+
+  /**
+   * The size of the pagination controls.
+   */
+  @property({ reflect: true }) size: 'small' | 'medium' | 'large' = 'medium';
+
+  /**
+   * The current page number. The default value is 1.
+   * The component will emit a "page-change" event whenever the page changes, allowing you to respond to page changes in your application.
```
kirchsuSICKAG: I would insert the real event name here

```suggestion
   * The component will emit a "syn-pagination-page-changed" event whenever the page changes, allowing you to respond to page changes in your application.
```

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+   */
+  @property({ attribute: 'current-page', reflect: true, type: Number }) currentPage = 1;
+
+  /**
+   * The number of items to display per page. The default value is 25.
+   * The component will emit a "page-size-change" event whenever the page size changes, allowing you to respond to page size changes in your application.
```
kirchsuSICKAG: I would insert the real event name here
```suggestion
   * The component will emit a "syn-pagination-page-size-changed" event whenever the page size changes, allowing you to respond to page size changes in your application.
```

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+        currentPage: newPage,
+        previousPage: currentPage,
+      },
+    });
+
+    this.currentPage = newPage;
+  }
+
+  /**
+   * Called when the page size is changed via the page size input.
```
kirchsuSICKAG: It's changed if the syn-select changes, not the input


```suggestion
   * Called when the page size is changed via the page size select.
```

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+        },
+      });
+    }
+  }
+
+  protected willUpdate(changed: PropertyValues<this>) {
+    super.willUpdate(changed);
+
+    if (changed.has('currentPage') || changed.has('pageSize') || changed.has('totalItems')) {
+      // If the current page exceeds the total number of pages based on the new page size or total items, we need to adjust it to ensure it remains within valid bounds.
```
kirchsuSICKAG: I think we need also to check the *pageSize* and *totalItems* here for invalid values. If the *pageSize* is set programatically via property, I am able to set an invalid page size, which has effect on the text "x-y of xyz items". Similar happens for invalid *totalItems* values (like negative numbers or characters)
See gif:

Image

kirchsuSICKAG (changes requested):

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+      ? this.pageSize
+      : this.pageSizeOptions[0];
+
+    const isFirstPage = this.currentPage === 1;
+    const isLastPage = this.currentPage === totalPages;
+
+    /* eslint-disable @typescript-eslint/unbound-method */
+    return html`
+      ${this.divider ? html`<syn-divider part="divider"></syn-divider>` : nothing}
+      <div class="pagination" part="base" ${ref(this.baseRef)}>
```
kirchsuSICKAG: To be better aligned with a11y, I think we should add *role="navigation"* and an aria-label to the outermost part of the pagination.
I found this in  paginations of other design systems (component libraries) 
- https://www.delldesignsystem.com/components/pagination/?tab=Vanilla
- https://mui.com/material-ui/react-pagination/#accessibility

Otherwise, the screen reader won't recognize that the inner parts of the pagination belong together.

Accessibility Tree WITHOUT *role="navigation"* and aria-label:
Image


Accessibility Tree WITH *role="navigation"* and aria-label:

Image

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+        ${!isCompact ? html`
+          <div class="pagination__page-size-select-wrapper" part="page-size-select-wrapper">
```
kirchsuSICKAG, line 348: Another thing, which came to my mind, as it would be relativ important for this component: do you now if one of our stakeholders need to support rtl? If not, I guess we can ignore it. But I think this is something we should clarify in general, if we want and need to consider this for new components

packages/components/src/components/pagination/pagination.styles.ts
```diff
@@ -0,0 +1,128 @@
...
+    :host {
+      container-type: inline-size;
+    }
+
+    @container (max-width: 400px) {
+      .pagination__navigation {
+        justify-content: center;
+      }
+    }
+  }
```
kirchsuSICKAG, line 127: Because of responsiveness: doesn't it make sense, that at some point to hide the select wrapper, so only the buttons are visible?

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+              @click=${() => this.updateCurrentPage(this.currentPage + 1)}
+              color="primary"
+              ?disabled=${isLastPage || isDisabled}
+              label=${this.localize.term('paginationNextPage')}
+              library="system"
+              name="next-page"
+              size=${this.size}
+            ></syn-icon-button>
+            
+            <syn-icon-button
```
kirchsuSICKAG, line 435: I think we should add more *part* attributes and also add *exportparts* to underlying synergy components.
As we needed to add them several times after requests in the past, when shoelace missed to do a *part* or *exportparts*.

E.g. syn-icon-button's, syn-select, syn-option

kirchsuSICKAG (changes requested):

packages/components/src/components/pagination/pagination.component.ts
```diff
@@ -0,0 +1,345 @@
...
+          <section>
+            <syn-icon-button
+              @click=${() => this.updateCurrentPage(this.currentPage + 1)}
+              color="primary"
+              ?disabled=${isLastPage || isDisabled}
+              label=${this.localize.term('paginationNextPage')}
+              library="system"
+              name="next-page"
+              size=${this.size}
+            ></syn-icon-button>
```
kirchsuSICKAG, line 433: When the start or end syn-icon-buttons are getting disabled for the first time, because the start or end of the pages are reached the first time, a console.warn is thrown in the console. To reproduce it, just use the "first" or "last" icon button, or use the "previous" or "next" icon buttons until you reach the end

Image

kirchsuSICKAG (changes requested):

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: You are right, that was misleading. Fixed.

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: I am not sure if we want to have each button as an exported entity. I think that may be overkill, dont you think?

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: I added a new version, will be there in the next iteration.

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: Thats a good question. I am not sure, but I think we should test it regardless.

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: I meant an input field, but I think it sounds misleading now that I read it with some time between. Adjusted

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: Fixed

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: Fixed

packages/components/src/components/pagination/pagination.styles.ts
schilchSICKAG: We did not do this on purpose because Design has not yet provided us any information about it. I use this to make sure it at least breaks evenly, but there is currently no "real" responsive behavior implemented.

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: I had a `<nav>` element instead of the div here, also providing a aria-label for it. However, this flags multiple errors for a11y:

- Each item needs a unique aria landmark. We would have to provide this via attributes, which we did not do anywhere else
- Each item needs a unique qualitfied id so the screen reader is able to distinguish multiple paginations on one page. This would mean we also have to forward an id, because random generation will have a negative impact (e.g.. react would rerender the whole component with a different id, defeating the purpose of having it in the first place).

Because of this, I opted to use a div instead. Both your examples are rendering into the light dom, which is kind of easier approach as you have anything in place. I would not overcomplicate it by adding two new properties and the needed usage guards.

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: Done

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: Fixed in latest

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: Weird race condition in syn-icon-button, fixed this by first blurring the element. Its because the internal events seem to race :(

packages/components/src/components/pagination/pagination.component.ts
kirchsuSICKAG: Theoretically they could be exported as a group via giving them all the same part name.

Bildschirmfoto 2026-04-21 um 15 08 35


```html
    <syn-pagination divider current-page="2" page-size="25" total-items="500" ></syn-pagination>
    <style>
      syn-pagination::part(navigation-button) {
        background: red;
      }
    </style>
 ```
    
   
Bildschirmfoto 2026-04-21 um 15 06 43

But I would also be fine for me to leave everything as is for now, and if anyone needs an export, we can just do it at this point.

packages/components/src/components/pagination/pagination.component.ts
kirchsuSICKAG: Would be fine for me if we leave it as it is for now

schilchSICKAG: Just added it. We wanted to be complete, thats ok for me.

packages/components/src/components/pagination/pagination.component.ts
schilchSICKAG: Thats indeed a idea, I added a shared part for the buttons, but not single selectors for them for the time being.

kirchsuSICKAG (approved):

PR #1271, opened 2026-04-29
feat: ✨ syn-chart

### LINKED ISSUE #1205: feat: ✨ new component syn-chart
# Story

As a user of the Synergy Design System, I would like to use one single chart component which serves as a wrapper for the whole charting 3rd party library, so I ...
- ... can easily use this library without the needed boiler plate
- ... get the SICK2025 styling out of the box
- ... get proper defaults which fits the average needs
- ... the series colors match the SICK2025 data colors

# Acceptance criteria

- [x] Chart color palettes are available in tokens package
- [x] Make sure chart tokens are known by the mcp
- [x] Define component name (syn-chart? syn-echart?)
- [x] LIT element is usable
- [x] API of the 3rd party library is transparent
- [x] Make chart instance accessible via getInstance so user can work with native API
- [x] Option property for setting the native 3rd party chart config (MVP)
- [x] Palette can be defined (pre-defined and custom)
- [x] Make sure the component is known by the MCP
- [x] Exclude from default export
- [x] Create chart doc section. (e.g. tell which chart types are supported by Synergy)
  - [x] Create structure: each chart type gets an own Storybook folder
  - [x] Document that there is much more possible than supported / tested
  - [x] Document that if a feature is needed which is not yet in supported/verified, contribution is welcome
- [x] Adapt build process to not include syn-chart in default wrapper exports (import syn-chart from @synergy-design-system/components/syn-chart.js)
- [x] Adapt angular build process to not include syn-chart in component module (+ how to add syn-chart --> angular components readme)
- [x] make sure types for React 19 are correct

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds the new chart palette tokens and the new syn-chart component to synergy.

#### tokens
- **fetching of new chart tokens** from figma chart file
- **refactoring of the code**:  
          - to be able to share the script functionalities
          - removed *chalk* dependency
- **new figma json outputs** under *src/figma-charts/*:
          - *chartTokens.json* 
          - *output/sick2025-light.json* | *output/sick2025-dark.json*
- **new build artifacts** under */dist/charts/*:
          - *themes/sick2025_light.css* | *sick2025_dark.css*
          - *js/index.js* | *js/index.d.ts*
          - *scss/_tokens.scss*
- **Token fixes:** Several hardcoded hex values in sick2025-light replaced with semantic token references to **muted** variables
- **Updated test script** (test/test-css-variables.js): now also covers new chart tokens

#### docs

- **New docs page**: *stories/tokens/ChartColors.mdx* - documents all chart color palettes with descriptions for the palettes.
- **New shared component**: *src/shared-components/TokenPalette.tsx* - reusable color palette table component.
- ** New charts section**: 
         - Overview page under *stories/Charting.mdx*
         - Stories for the chart component under *stories/chart.stories.ts* 

#### components
- ** new syn-chart component**:
      - with *echarts* as optional peer dependency
      - chart component is not part of the default bundle
      - currently only line chart is supported
      - typescript workaround for echarts 6.0.0 added
- ** updated build infrastructure** by removing *globby* and *del*

### 🎫 Issues

Closes #1205 

#### Is the folder structure and file naming for the new chart tokens alright? 
Bildschirmfoto 2026-05-13 um 14 57 49

#### how should the naming of the *palette* property be?
- either exactly how the tokens are called: *categorical*, *sequential-01*,  ..., *sequential-07*, *sequential-status-critical*, ... *sequential-status-info*
- or like the underlying base tokens: *categorical*, *sequential-primary*, *sequential-accent*, ..., *sequential-neutral*, *status-critical*, ... , *status-info*
       
#### how to implement the tree shakeability for the syn-chart component?
We want our chart component to:
✅ Support all ECharts features
✅ Remain tree-shakable
✅ Keep bundle size minimal

This introduces a trade-off between:
👉 Bundle size optimization vs. Developer Experience (DX)

##### Option 1: Fully Tree-Shakable (Consumer Registers Features)
The component registers only the minimum required module (CanvasRenderer).
All other ECharts features (charts, components, etc.) must be explicitly registered by the user using use([...]).
We provide re-exports so that consumers do not need to install ECharts separately.

```js
import { use, LineChart, GridComponent, TooltipComponent }  from '@synergy-design-system/components/chart.js';
import '@synergy-design-system/components/chart.js';

use([LineChart, GridComponent, TooltipComponent]);

document.querySelector('syn-chart').option = {
  xAxis: { type: 'category', data: ['Mon', 'Tue', 'Wed'] },
  yAxis: { type: 'value' },
  series: [{ type: 'line', data: [150, 230, 224] }],
};
````

###### Advantages
- **Minimal bundle size**: Only imported features are included
- ** Scales with usage**: Small apps stay small, large apps pay only what they use
- **Maximum flexibility** :Full ECharts API available
- **Future-proof**: New chart types can be added without modifying the component

###### Disadvantages
- **Higher setup effort**: Consumers must call use([...])
- **Error-prone**: Missing registrations lead to empty charts or runtime errors
- **Ordering constraints**: use() must be called before rendering
- **Less beginner-friendly**

##### Option 2: Preconfigured / Convenience Setup

The library provides a preconfigured version where all (or most) ECharts features are registered automatically.
Consumers can use charts without calling use().
```js
import '@synergy-design-system/components/chart.js';

document.querySelector('syn-chart').option = {
  xAxis: { type: 'category', data: ['Mon', 'Tue', 'Wed'] },
  yAxis: { type: 'value' },
  series: [{ type: 'line', data: [150, 230, 224] }],
};
````

###### Advantages
- **Zero setup**: Works out of the box
- **Better dx**: Especially for beginners
- **Fewer runtime errors**: No missing registrations

###### Disadvantages
- **Large bundle size**: All chart types are always included
- **No scaling**: Even simple use cases pay full cost
- **Breaks tree-shaking benefits**
- If user needs a specific chart type, which we do not support, we need to add it

### Steps to do
- [x] Chart color palettes are available in tokens package
- [x] Make sure chart tokens are known by the mcp
- [x] Define component name (syn-chart? syn-echart?)
- [x] LIT element is usable
- [x] API of the 3rd party library is transparent
- [x] Make chart instance accessible via getInstance so user can work with native API
- [x] Option property for setting the native 3rd party chart config (MVP)
- [x] Palette can be defined (pre-defined and custom)
- [x] Make sure the component is known by the MCP
- [x] Exclude from default export
- [x] Create chart doc section. (e.g. tell which chart types are supported by Synergy)
  - [x] Create structure: each chart type gets an own Storybook folder
  - [x] Document that there is much more possible than supported / tested
  - [x] Document that if a feature is needed which is not yet in supported/verified, contribution is welcome
- [x] Adapt build process to not include syn-chart in default wrapper exports (import syn-chart from @synergy-design-system/components/syn-chart.js)
- [x] Adapt angular build process to not include syn-chart in component module (+ how to add syn-chart --> angular components readme)
- [x] make sure types for React 19 are correct

## ✅ DoD

### CODE REVIEW

File: packages/vue/package.json [hunk 1]
```diff
@@ -12,7 +12,10 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/index.js"
     },
-    "./components/*": "./dist/components/*"
+    "./components/*": {
```
schilchSICKAG, line 15: Just thought about if we should add a charts export into this, too, e.g.:

```json
    "./charts": {
      "types": "./dist/chart.d.ts",
      "import": "./dist/chart.js"
    }
```

Not sure if it makes that much sense, though.

packages/_private/demo-utilities/tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "declaration": true,
     "declarationDir": "./dist",
     "sourceMap": false,
-    "strict": true
+    "strict": true,
+    "skipLibCheck": true
```
schilchSICKAG: Why is this needed?

packages/_private/eslint-config-syn/ts.js
```diff
@@ -58,7 +58,7 @@ export const createCustomConfig = ({
       '@stylistic/quotes': ['error', 'single', { avoidEscape: true }],
       '@stylistic/semi': ['error', 'always'],
       '@typescript-eslint/no-shadow': ['error'],
-      '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+      '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_', }],
```
schilchSICKAG: What is this for? Maybe we should add comments when we change such values in the future. I know that I did not do this when initially releasing this, but it would have been good to know why we already had argsIgnorePattern and now even have caughtErrorsIgnorePattern :)

packages/_private/vue-demo/src/AllComponentParts/DemoChart.vue
```diff
@@ -0,0 +1,10 @@
+<script setup lang="ts">
+  import SynVueChart from '@synergy-design-system/vue/components/SynVueChart';
+  import { mockData } from '@synergy-design-system/demo-utilities';
+
```
schilchSICKAG: Remove unneeded lines please

packages/components/scripts/jobs/synergy/adjustComponentsForDefaultSettings.js
```diff
@@ -49,7 +49,7 @@ export const adjustComponentsForDefaultSettings = job('Synergy: Adjusting compon
       try {
         await fs.writeFile(componentFile, newContent);
         return STATE_WRITTEN;
-      } catch (err) {
+      } catch (_err) {
```
schilchSICKAG: Why? :)

packages/components/scripts/jobs/synergy/esbuildComponents.js
```diff
@@ -1,6 +1,6 @@
+import { glob } from 'node:fs/promises';
```
schilchSICKAG, line 1: Nice addition!

packages/components/scripts/jobs/angular/createComponents.js
```diff
@@ -211,11 +206,13 @@ export const runCreateComponents = job('Angular: Creating components', async (me
       ${eventExports}
     `.trim();
 
-    index.push({
-      name: `${component.name}Component`,
-      outputPath: `@synergy-design-system/angular/components/${component.tagNameWithoutPrefix}`,
-    });
-
+    // Remove chart component from the index, as it should not be part of the default exports
+    if (component.tagNameWithoutPrefix !== 'chart') {
```
schilchSICKAG: Maybe we could reuse this more if we would adjust createFrameworkIndex and do the filter there? 

```javascript
  const alphabeticIndex = [...components]
    .sort(sortByComponentName)
    .filter({ name }) => name !== 'chart') // Adapt to taste
    .map(({ name, outputPath }) => {
      // When we use useDefaultExport,
      // make sure to export the default as named export
      // This is for example needed for vue sfc
      const exportStatement = useDefaultExport ? `default as ${name}` : name;
      return `export { ${exportStatement} } from '${outputPath}';`;
    });

```

packages/components/scripts/jobs/react/createWrappers.js
```diff
@@ -65,10 +62,13 @@ export const runCreateWrappers = job('React: Creating Component Wrappers...', as
       ${eventExports}
     `;
 
-    index.push({
-      name: component.name,
-      outputPath: `./components/${component.tagNameWithoutPrefix}.js`,
-    });
+    // Remove chart component from the index, as it should not be part of the default exports
+    if (component.tagNameWithoutPrefix !== 'chart') {
```
schilchSICKAG: See comment on angular createComponent: I think we can do this on another level already present to remove duplicate logic

packages/components/scripts/jobs/vue/createWrappers.js
```diff
@@ -279,10 +277,18 @@ ${exports}
 </template>
 `.trim();
 
-    index.push({
-      name: vueComponentName,
-      outputPath: `./components/${vueComponentName}.vue`,
-    });
+    // Remove chart component from the index, as it should not be part of the default exports
+    if (component.tagNameWithoutPrefix !== 'chart') {
```
schilchSICKAG: See other frameworks. Can be done in createFrameworkIndex

packages/components/README.md
```diff
@@ -209,7 +209,87 @@ document.addEventListener("load", () => {
 
 ---
 
-### 6. Add html autocompletion to VSCode (optional)
+### 6. Using the `syn-chart` component (optional)
+
+For more detailed information about the `syn-chart` see the [Charting Overview page](https://synergy-design-system.github.io/?path=/docs/charting-overview--docs)
```
schilchSICKAG: about the `syn-chart` see => about the `syn-chart` component see

packages/components/README.md
```diff
@@ -209,7 +209,87 @@ document.addEventListener("load", () => {
...
+
+The `<syn-chart>` component allows you to display charts powered by [Apache ECharts](https://echarts.apache.org). Because not every project needs charting functionality, `echarts` is an **optional peer dependency** that must be installed separately.
+
+> **Note:** `syn-chart` is currently **experimental**. Only **line charts** (`series[].type: 'line'`) are supported. Support for additional chart types will be added in future releases.
+
+#### Step 1: Install the `echarts` dependency
```
schilchSICKAG, line 220: Maybe we should add a disclaimer which echarts versions are supported?

packages/docs/stories/charts/chart.stories.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export const Option: Story = {
+  parameters: {
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'option'),
+      },
+    },
+  },
+  render: () => html`
```
schilchSICKAG, line 94: Just a tip: You can create dynamically scoped selectors based on the stories name when having a look at the second parameter of `render`: `render: (_, opts) => console.log(opts.storyId)`. With this you don´t have to define unique ids by yourself all the time.

packages/docs/stories/charts/chart.stories.ts
schilchSICKAG, line 1: I think we are missing a screenshot story. Or did you leave the Screenshots out on purpose?

packages/docs/stories/tokens/Colors.mdx
```diff
@@ -1,87 +1,33 @@
...
-  TokensForThemes,
-} from '../../src/shared-components/TokensForThemes.tsx';
 
 <Meta title="Tokens/Colors" tags={['Design Tokens', 'Colors']} />
 
-# Colors
-
-<MultiLineDescription value={docsTokens?.tokens?.color?.description?.value} />
-
-<TokensForThemes
```
schilchSICKAG, line 28: Clever reuse of your new ColorPalette. Like this :)

packages/mcp/test/watermarks/baseline.latest-release.json
```diff
@@ -46,7 +46,7 @@
     },
     {
       "id": "token-info:css:sick2025-light",
-      "tokens": 8474
+      "tokens": 10786
```
schilchSICKAG: Why this increase?

packages/tokens/scripts/outputs/js.js
```diff
@@ -20,6 +20,7 @@ export const createJS = (header, inputFile, outputFile) => {
       return [
         pascalCase(varName, {
           mergeAmbiguousCharacters: true,
+          split:  (value) => joinConsecutiveNumbers(split(value)),
```
schilchSICKAG: Trailing whitespace between `:__`.

packages/tokens/scripts/outputs/scss.js
```diff
@@ -20,6 +20,7 @@ export const createSCSS = (header, inputFile, outputFile) => {
       return [
         pascalCase(varName, {
           mergeAmbiguousCharacters: true,
+          split:  (value) => joinConsecutiveNumbers(split(value)),
```
schilchSICKAG: Doubled whitespace `:__`

packages/tokens/scripts/build-charts.js
```diff
@@ -0,0 +1,34 @@
...
+import { copyDefaultThemeAliases, runBuildPipeline } from './build-generic.js';
+
+const DEFAULT_THEME = getDefaultTheme();
+
+/**
+ * Returns full theme metadata for a chart theme.
+ * @param {string} theme - e.g. "sick2025-light"
+ * @param {string} mode - e.g. "light"
+ */
+const getChartThemeInformation = (theme, mode) => {
```
schilchSICKAG: If you rename this to `getThemeInformation`, you can use:

```javascript
  filter: ({filePath}) => filePath.includes('figma-charts'),
  getThemeInformation,
```

packages/tokens/test/test-css-variables.js
schilchSICKAG, line 1: Could you add a eslint disable for console into this file?

packages/tokens/scripts/build.js
schilchSICKAG, line 1: I think I would have phrased it the other way around: have one `build.js` as an entry target that runs all sub builders, something like:

```bash
scripts/
  src/
    builders/
      charts.js
      components.js
      index.js
    build.js
```

and then have the builders export generic async functions:

```javascript
// charts.js builder
export const buildCharts = async () => {
  return runBuildPipeline({
  buildPath: './dist/charts/',
  copyToDefault: (themesDir) => copyDefaultThemeAliases(themesDir, DEFAULT_THEME),
  // For charting tokens we need to include the component tokens in the build for resolving references,
  // but we don't want to output them in the charting CSS files since they're not used directly.
  filter: ({filePath}) => filePath.includes('figma-charts'),
  getThemeInformation: getChartThemeInformation,
  prefix: 'syn-',
  sourceDir: CHARTS_OUTPUT_DIR,
});
}
```

Then you could adjust build.js to:

```javascript
import * as builders from './builders/index.js';

const build = async () => {
const calls = builders.map(b => b());
const results = await Promise.allSettled(calls);
// Do something with results here if there are any
```

This would have the advantage that you still have just one entry (also in package.json) and leave this as an extendible pattern.

I am fine with the way it currently works, but its a bit different than our other scripts in other packages.

packages/components/src/components/chart/chart.ts
schilchSICKAG, line 1: I receive strange Typescript errors for this file (only in the Editor):

Image

packages/components/src/components/chart/chart.styles.ts
```diff
@@ -0,0 +1,19 @@
+import { css } from 'lit';
+
+export default css`
+  :host {
+    --syn-chart-aspect-ratio: 16 / 9;
```
schilchSICKAG: Should this be public? If so, we have to add it the chart component doc. Also, `--syn` should only be used for official variables, so maybe we should change this to `--aspect-ratio` if we want to keep it? Other css properties don´t have a syn-prefix

packages/components/src/components/chart/chart.styles.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    aspect-ratio: var(--syn-chart-aspect-ratio);
+    display: block;
+
+    /* TODO: do we want / need a default height? If yes, how much? */
+    min-height: 250px;
```
schilchSICKAG: If we need this, we should make it configurable, too?

packages/components/src/components/chart/types.ts
schilchSICKAG, line 1: I am thinking about if we should rename this to `types.ts` instead. There are no utilities in there, just the annotated narrow types for the future, right?

packages/components/src/components/chart/chart.component.ts
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * @csspart base - The component's base wrapper.
+ */
+export default class SynChart extends SynergyElement {
+  static styles: CSSResultGroup = [
+    componentStyles,
+    styles,
+  ];
+
+  @query('.chart', true)
```
schilchSICKAG: I think `true` is the default. We left it out everywhere else, maybe adjust? It does not break anything, but if it does not hurt performance I would remove this.

packages/components/src/components/chart/chart.component.ts
```diff
@@ -0,0 +1,183 @@
...
+} from 'echarts/components.js';
+import { classMap } from 'lit/directives/class-map.js';
+import SynergyElement from '../../internal/synergy-element.js';
+import componentStyles from '../../styles/component.styles.js';
+import styles from './chart.styles.js';
+import { PALETTE_TOKENS } from './chart.palettes.js';
+import type { ECOption } from './utilities.js';
+
+// TODO: Check, should we let the user define the *use* so the bundle size is optimized for their specific use case?
+use([
```
schilchSICKAG, line 21: We could load stuff dynamically. However, I notice that eCharts uses a global configuration, which means it maybe does not make any sense to dynamically import stuff at all - each instance will get all plugins anyways.

We could add presets for this, something like

```bash
components/
  charts/
    presets/
      line.ts // imports line and everything that is used)
      index.ts // Exports all presets
    chart.component.ts // checks which preset to use and calls use
```

However, this may be too much for a first version.

packages/components/src/components/chart/chart.component.ts
```diff
@@ -0,0 +1,183 @@
...
+   * ```js
+   * chart.option = {
+   *   xAxis: { type: 'category', data: ['Mon', 'Tue', 'Wed'] },
+   *   yAxis: { type: 'value' },
+   *   series: [{ type: 'line', data: [150, 230, 224] }],
+   * };
+   * ```
+   */
+  @property({ attribute: false })
+  option: ECOption = {};
```
schilchSICKAG: I really dislike the naming here. I know it is derived from the external library, but its ugly for multiple reasons:

a) Wording is singular, however you configure multiple option**s**
b) Key naming seem to be all over the place. Again this is foreign design, I get it, but typing something like `document.querySelector('syn-chart').option.options = {...}` feels weird.
c) As this is more of a settings object than anything else, I would opt with a name change, e.g. `config: ECOption` or `setup: ECOption`. Sounds much closer to what this actually does and leaves the duplicated wording option.options out.

packages/components/src/components/chart/chart.component.ts
```diff
@@ -0,0 +1,183 @@
...
+   * - `sequential-01` … `sequential-07` — 10-step single-hue ramps:
+   *   `01`=primary, `02`=accent, `03`=muted, `04`=purple, `05`=teal, `06`=magenta, `07`=neutral
+   * - `sequential-status-critical`, `sequential-status-error`, `sequential-status-info`,
+   *   `sequential-status-success`, `sequential-status-warning` — 10-step status ramps
+   *
+   * The palette sets the ECharts `color` array. If `option.color` is explicitly provided,
+   * it takes precedence over the palette.
+   */
+  @property({ reflect: true })
+  palette: 'categorical' | 'sequential-01' | 'sequential-02' | 'sequential-03' | 'sequential-04' | 'sequential-05'
```
schilchSICKAG: Can´t we infer the type from `chart.palettes.ts`, e.g. `palette: SynChartPalette`?. I darkly remember there could be problems when building (e.g. storybook may not be able to build a select out of the keys. Is this the reason you went with this?

packages/components/src/components/chart/chart.component.ts
```diff
@@ -0,0 +1,183 @@
...
+  | 'sequential-06' | 'sequential-07' | 'sequential-status-critical' | 'sequential-status-error' | 'sequential-status-info'
+  | 'sequential-status-success' | 'sequential-status-warning' = 'categorical';
+
+  /** Resolves palette CSS custom properties to computed color values and applies them to the chart. */
+  private applyPalette(): void {
```
schilchSICKAG, line 94: I am not sure if this may break SSR because its called in updated and firstUpdated. Could you enable SSR for angular and check if it works without breaking?

packages/components/src/components/chart/chart.component.ts
```diff
@@ -0,0 +1,183 @@
...
+    return this.chartInstance;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  render() {
+    return html`
+      <div
+        part="base"
+        class=${classMap({
+      chart: true,
```
schilchSICKAG: Formatting issue

schilchSICKAG (changes requested):
I think for a first iteration this is pretty stable. Good job, also making sure to remove more of the now unneeded dependencies was pretty nice to see.

packages/vue/package.json
kirchsuSICKAG: I think I would rather not do it as the charts are also part of the components and they are already getting exported via the **/components/*** export. 
Currently the import for the users look like this

"import '@synergy-design-system/components/components/chart/chart.js';"

If we do it like this:

```
   "./charts": {
      "types": "./dist/components/chart.d.ts",
      "import": "./dist/components/chart.js"
    }
```
then we have: 
"import '@synergy-design-system/components/chart/chart.js';"

But I don't see an advantage, except we say, the chart should not be part of the components

packages/_private/demo-utilities/tsconfig.json
kirchsuSICKAG: As long as we need to stick to echarts version 6.0.0 we are getting typescript errors e.g. for not using file endings in imports etc. Because of that the building of the demo-utilities package fails. 
For now I could either add this *"skibLibCheck: true"* or change the *moduleResolution* to bundler.

packages/components/scripts/jobs/synergy/adjustComponentsForDefaultSettings.js
kirchsuSICKAG: I added it because of this error: 
Bildschirmfoto 2026-05-19 um 12 27 25

Together with the new rule which I added to the eslint-config-syn package eslint does no longer complain, because it ignores underscored non-used variables:

```
      '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_', }],
 ```

File: packages/docs/stories/charts/chart.stories.ts (hunk 12 above)
Comment (kirchsuSICKAG) [reply packages/docs/stories/charts/chart.stories.ts:94]: That's nice. But unfortunately this does not work for the screenshot story 😕
There all of the stories have the same id and therefore only the first *syn-chart* element get's the *option* property set all the time:

Bildschirmfoto 2026-05-19 um 14 19 11

File: packages/docs/stories/charts/chart.stories.ts
Comment (kirchsuSICKAG) [reply packages/docs/stories/charts/chart.stories.ts:1]: I just forgot to add it back 😅

File: packages/components/src/components/chart/chart.component.ts (hunk 22 above)
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.component.ts]: I would be fine if we call it *config* instead of *option* on top level. As long as we link to the echarts documentation and everything remains the same under the hood as in echarts.

But there is no double option at all. Where does it come from? You mean this: https://echarts.apache.org/en/option.html#options?
This is something which is only used for the *timeline* feature if I see it correctly (the bottom time line thing in the image)
Bildschirmfoto 2026-05-19 um 14 33 22

So I don't think this is something which is used very often.

Normally it would look more like following:

```js
      document.querySelector('syn-chart').option = {
        series: [{ data: [150, 230, 224, 218, 135, 147, 260], type: 'line' }],
        tooltip: { trigger: 'item' },
        xAxis: { data: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'], type: 'category' },
        yAxis: { type: 'value' },
      };
 ```

File: packages/components/src/components/chart/chart.component.ts (hunk 23 above)
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.component.ts]: Yeah, first i did it like this: 

````
 @property({ reflect: true })
  palette: SynChartPalette = 'categorical';
````

But then we only get a textfield in the docs page instead of a select 😕

Bildschirmfoto 2026-05-19 um 14 41 06

Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.component.ts]: What I maybe could do is try do solve it in the story docs builder, but I think this would be a special handling for the chart (palette property)

File: packages/components/src/components/chart/chart.component.ts (hunk 25 above)
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.component.ts]: Fixed

File: packages/_private/eslint-config-syn/ts.js (hunk 3 above)
Comment (kirchsuSICKAG) [reply packages/_private/eslint-config-syn/ts.js]: Removed it. The lint error was solved differently

File: packages/_private/vue-demo/src/AllComponentParts/DemoChart.vue (hunk 4 above)
Comment (kirchsuSICKAG) [reply packages/_private/vue-demo/src/AllComponentParts/DemoChart.vue]: Fixed

File: packages/components/scripts/jobs/synergy/adjustComponentsForDefaultSettings.js (hunk 5 above)
Comment (kirchsuSICKAG) [reply packages/components/scripts/jobs/synergy/adjustComponentsForDefaultSettings.js]: Fixed this by removing the rule and just removing the error arg

File: packages/components/README.md (hunk 10 above)
Comment (kirchsuSICKAG) [reply packages/components/README.md]: Fixed

File: packages/components/README.md (hunk 11 above)
Comment (kirchsuSICKAG) [reply packages/components/README.md:220]: Done

File: packages/tokens/scripts/outputs/js.js (hunk 15 above)
Comment (kirchsuSICKAG) [reply packages/tokens/scripts/outputs/js.js]: Fixed

File: packages/tokens/scripts/outputs/scss.js (hunk 16 above)
Comment (kirchsuSICKAG) [reply packages/tokens/scripts/outputs/scss.js]: Fixed

File: packages/tokens/test/test-css-variables.js
Comment (kirchsuSICKAG) [reply packages/tokens/test/test-css-variables.js:1]: Added

File: packages/components/src/components/chart/chart.ts
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.ts:1]: Fixed via ts-ignore, as it is a false positive workspace typescript error

Image

File: packages/components/src/components/chart/types.ts
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/types.ts:1]: Done

File: packages/mcp/test/watermarks/baseline.latest-release.json (hunk 14 above)
Comment (kirchsuSICKAG) [reply packages/mcp/test/watermarks/baseline.latest-release.json]: Good catch. This was the new baseline before I separated the components tokens and chart tokens in the mcp tool token-info. I updated the baseline again

File: packages/components/src/components/chart/chart.component.ts (hunk 23 above)
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.component.ts]: I got it working in storybook. Fixed in the newest version

File: packages/vue/package.json (hunk 1 above)
Comment (kirchsuSICKAG) [reply packages/vue/package.json:15]: Oh now I see. I thought you meant it for the components package, but it is vue. 
originally I created the chart.ts file, because otherwise the d.ts files for the SynVueChart component was not generated (the plugin needs entry files for this).
But you are right, we could do that

File: packages/components/scripts/jobs/angular/createComponents.js (hunk 7 above)
Comment (kirchsuSICKAG) [reply packages/components/scripts/jobs/angular/createComponents.js]: Done

File: packages/components/scripts/jobs/react/createWrappers.js (hunk 8 above)
Comment (kirchsuSICKAG) [reply packages/components/scripts/jobs/react/createWrappers.js]: Done

File: packages/components/scripts/jobs/vue/createWrappers.js (hunk 9 above)
Comment (kirchsuSICKAG) [reply packages/components/scripts/jobs/vue/createWrappers.js]: Done

File: packages/components/src/components/chart/chart.styles.ts (hunk 18 above)
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.styles.ts]: Good question. Not sure if this should be public or not.
Theoretically could this also be overwritten via parts or on the component itself. 
If this would be public, the people would get a hint how to change it.

I think for now I am more team not-public and then we can also remove the css variable completely. 
But maybe this is also something which design needs to say, what is the best aspect-ratio for charts?
@reichst1SICKAG?

File: packages/components/src/components/chart/chart.styles.ts (hunk 19 above)
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.styles.ts]: @reichst1SICKAG Is there something like a good default min-height that charts should not get smaller than?

File: packages/components/src/components/chart/chart.component.ts (hunk 24 above)
Comment (kirchsuSICKAG) [reply packages/components/src/components/chart/chart.component.ts:94]: I tested it out with an angular-ssr project. Everything seems fine there and nothing was breaking

File: packages/tokens/scripts/build-charts.js (hunk 17 above)
Comment (kirchsuSICKAG) [reply packages/tokens/scripts/build-charts.js]: Done

File: packages/vue/package.json (hunk 1 above)
Comment (kirchsuSICKAG) [reply packages/vue/package.json:15]: I added the import. 

Instead of doing this (which still can be done):

```js
  import SynVueChart from '@synergy-design-system/vue/components/SynVueChart';
```

This can also be done:

```js
    import { SynVueChart } from '@synergy-design-system/vue/chart';
```

I opted to go for **chart** instead of **charts**. Not sure why, just from my feeling, what do you think is better @schilchSICKAG?

packages/tokens/scripts/build.js
kirchsuSICKAG: Done

packages/_private/demo-utilities/tsconfig.json
kirchsuSICKAG: Is removed, as echarts is now updated

packages/vue/package.json
schilchSICKAG: Agreed.

packages/components/src/components/chart/chart.component.ts
schilchSICKAG: Yes thats what I mean. I think the naming of echarts configuration is all over the place when it comes to variable and option naming. I would also go with config, but maybe we can discuss this with stakeholders. I know it can get more complicated when we rename things, but `option` is such a generic name and it also does not what I would think an `option` would (or should) do.

packages/components/src/components/chart/chart.component.ts
schilchSICKAG: Good solution, even if we have to do it by hand in Storybook, we will only do this once.

packages/docs/stories/tokens/ChartColors.mdx
```diff
@@ -0,0 +1,46 @@
...
+import {
+  generateStoryDescription
+} from '../../src/helpers/component.ts';
+import {
+  TokenPalette,
+} from '../../src/shared-components/TokenPalette.tsx';
+
+<Meta title="Tokens/Chart Colors" tags={['Design Tokens', 'Colors', 'Charts']} />
+
+# Color Palettes
```
reichst1SICKAG, line 15: can we show the source token/variable instead of the raw value?
So instead of #005aff we use primary-700?

packages/docs/stories/charts/chart.stories.ts
```diff
@@ -0,0 +1,217 @@
...
+  parameters: {
+    chromatic: {
+      modes: Chromatic_Modes_All,
+    },
+    design: generateFigmaPluginObject('41094-279501'),
+    docs: {
+      description: {
+        component:
+          [
+            '>⚠️ **Experimental Status**:',
```
reichst1SICKAG: could you use the same badge like we did in the components?

storybook-badges

packages/components/CHARTING_README.md
```diff
@@ -0,0 +1,230 @@
...
+- `LineSeriesOption` (line charts)
+- `CanvasRenderer`, `TitleComponentOption`, `TooltipComponentOption`, `LegendComponentOption`, `GridComponentOption`
+
+> If you use TypeScript and assign options for unsupported chart types (e.g. `BarSeriesOption`), you will get a **type error**. This is intentional — it reflects the actual runtime capabilities of the component.
+
+---
+
+## Bundle Size
+
+`syn-chart` uses [ECharts tree-shaking](https://echarts.apache.org/en/tutorial.html#Use%20ECharts%20with%20bundler%20and%20NPM) internally. Only the modules that are actually needed (currently `LineChart`, `CanvasRenderer`, `GridComponent`, `LegendComponent`, `TitleComponent`, `TooltipComponent`) are imported and registered via ECharts' `use([...])`. You do **not** need to register anything yourself.
```
reichst1SICKAG: make this "You do not need to register anything yourself." a bit more priminent and place this in a new line/paragraph?

packages/components/src/components/chart/chart.styles.ts
reichst1SICKAG: a donut can get pretty small and still be readable. I tested with 180px in Figma and looked good.

packages/components/src/components/chart/chart.styles.ts
```diff
@@ -0,0 +1,15 @@
+import { css } from 'lit';
+
+export default css`
+  :host {
```
reichst1SICKAG, line 4: did you use an animation duration token or how is the duration of the initial chart animation defined?

kirchsuSICKAG: @reichst1SICKAG  It is defined by echarts. There are different animation easings you can choose from and I think they have different defaults per chart type (https://echarts.apache.org/examples/en/editor.html?c=line-easing).
The current default animation duration is 1000ms.

Not in this branch, but in the next branch I add the synergy theming of echarts via a config. Then I can overwrite the animations there, also with our tokens. What would be the correct animation duration time?  then I will add it in the next branch

packages/docs/stories/charts/chart.stories.ts
kirchsuSICKAG: Unfortunatly not. It only allows string content here.
Maybe we can do something like this for each docs page in another ticket. Therefore we would have to add a storybook plugin or enhance our docs stories with the component state and component version.

packages/components/src/components/chart/chart.styles.ts
reichst1SICKAG: Understood. We have those durations currently. But do we need to overwrite them? We still have no overall animation concept as we are using shoelace and now echarts. It could fit together

duration.extra-fast 50ms
duration.fast 150ms
duration.normal 250ms
duration.slow 500ms
duration.extra-slow 1000ms

packages/components/src/components/chart/chart.styles.ts
kirchsuSICKAG: I exchanged it with the 180px

packages/docs/stories/tokens/ChartColors.mdx
kirchsuSICKAG: done. Now both raw value and css alias are shown

packages/components/src/components/chart/chart.component.ts
kirchsuSICKAG: renamed the property to *config*

packages/components/src/components/chart/chart.styles.ts
kirchsuSICKAG: Animations will be tackled after animation concept is defined

reichst1SICKAG (approved):

PR #1276, opened 2026-05-04
docs: 📚 Describe component context guidelines

### LINKED ISSUE #977: feat: ✨ MCP: Add support for a explain component prompt

# Story

As a developer using the Synergy Design System MCP server, I would like to have the ability to directly query the MCP via slash commands so that I have control over what is called without relying on formulating a prompt in my agent.

I want to have the possibility to explain the usage of a synergy component and get useful information back.

## Interface:

```bash
/explain syn-button in angular
/explain syn-input in vue
```

# Acceptance criteria

- [x] The prompt is able to use information from the metadata package
- [x] Design System Rules from #1187 are in place so we can use them.
- [x] Adjust the `component-info` tool to include it as an optional layer `rules`.
- [x] Documentation for the MCP is updated

### LINKED ISSUE #1187: feat: ✨ Describe component context guidelines
# Description

Similar to the brand page, we need to describe the context of each component and their variants.
Templates are helpful, but a description is also necessary. The template should be built on-top of these descriptions.

# Rationale

The user needs guidance that describes the situations in which a component and their variants should be used in which context.

E.g. a outline button should be used for secondary options. A filled button should be used for primary action, etc.

Good examples are:
 - https://mui.com/material-ui/react-button/
 - https://www.kern-ux.de

# Acceptance criteria

- [ ] Check old SDS for reuse: https://brand.sick.com/document/51#/-/elements
- [ ] Check for Solid Design System reuse https://solid-design-system.fe.union-investment.de/docs/?path=/docs/components-sd-checkbox-overview--docs&globals=viewport.value:mobile1;solidThemeGenerator:!true
- [ ] which to use checkbox and when a switch
- [ ] when to use which button variant
- [ ] when to use which font-size /-weight
- [ ] which navigation type is optimal when: tabs, side-navigation, ...
- [ ] button types and -order in dialogs
- [ ] which alert variant should be used when
- [ ] when to use progress-bar vs -ring
- [ ] tonality
- [ ] ...

### LINKED ISSUE #1283: feat: ✨ Introduce Optional Intent Policy Layer to Component Metadata

# Story

As a **consumer and integrator of the Synergy Design System** (including documentation, tooling, and AI-assisted workflows), I would like to have an **optional intent / usage policy layer** available in the metadata package, so that **tools and skills can reliably reason about *why* a component is used in a specific situation, without polluting the canonical component metadata**.

## Use case and problem

Synergy already provides:
- canonical component metadata (attributes, properties, slots, defaults),
- UX and accessibility rules that describe *good usage in principle*,
- and this metadata is used as a source of truth by tooling (e.g. MCP, docs generation).

However, there is currently no structured way to describe **contextual usage intent**, such as:
- “this button submits a form”,
- “this is the primary action”,
- “this input collects numeric data”,
- “this dialog represents a destructive action”.

These concepts are essential for:
- generating correct component examples,
- guiding AI-assisted code generation,
- validating usage consistently,
- and avoiding ambiguous or hallucinated decisions in tooling.

Encoding these intents directly into component metadata would break its contract (metadata must always be true and context‑independent).
Not encoding them at all forces downstream consumers (tools, MCP apps, local skills) to infer intent heuristically.

## Proposed approach (high level)

Introduce an **optional, opt-in “intent policy layer” inside the metadata package**, clearly separated from core component metadata.

This layer would:
- describe *possible intents and their relationship to components* (capabilities),
- define *safe preset configurations* for a component–intent combination,
- expose this information only when explicitly queried,
- and depend on core metadata, but never the other way around.

Importantly:
- core component metadata remains unchanged and intent-agnostic,
- intent information is not injected into base component objects,
- consumers must explicitly opt in to intent-aware behavior.

## Why this is needed

This enables:
- deterministic example generation (per intent),
- foundation for future validation or linting tools,
- consistent behavior across MCP, local developer skills, and documentation,
- reduced ambiguity for AI-assisted tooling,
- and keeps the core metadata stable, clean, and future-proof.

## First draft

This is a first draft of creating an intent System. It has to be sharpened and checked with design before implementation:

```javascript
// policy/intents.js
// List of available intents in the whole system
export const Intents = {
  ACTION_PRIMARY: "action.primary",
  ACTION_SUBMIT: "action.submit",
  ACTION_RESET: "action.reset",
  ACTION_NAVIGATION: "action.navigation",
};

// policy/capabilities.js
// This defines which components can be used in which capability
export const Capabilities = {
  "syn-button": ["action"],
};

// policy/presets/syn-button.js
// This shows how a syn-button may be mapped to intents
import { ACTION_PRIMARY, ACTION_SUBMIT, ACTION_RESET, ACTION_NAVIGATION } from './intents.js';

export const SynButton = {
  [ACTION_PRIMARY]: {
    preset: "syn-button.action.primary",
    props: {
      variant: "filled",
      size: "medium",
      type: "button",
    },
  },

  [ACTION_SUBMIT]: {
    preset: "syn-button.action.submit",
    props: {
      type: "submit",
      variant: "filled",
    },
    forbiddenProps: ["href"],
  },

  [ACTION_RESET]: {
    preset: "syn-button.action.reset",
    props: {
      type: "reset",
      variant: "text",
    },
  },

  [ACTION_NAVIGATION]: {
    preset: "syn-button.action.navigation",
    props: {
      href: "#",
      variant: "text",
    },
  },
};
```

# Acceptance criteria

- [ ] Intent information is modeled as a **separate, optional layer** within the metadata package
- [ ] Core component metadata remains **unchanged and intent-agnostic**
- [ ] Intents describe *contextual usage goals*, not always-true properties
- [ ] API allows querying possible intents / presets for a component explicitly
- [ ] No consumer is required to use intents unless explicitly opting in
- [ ] Design allows reuse by tooling (e.g. MCP, example generation, future validation)

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds some new features and data additions to a range of Synergy packages to make them usable in our Metadata, MCP and Docs packages:

### Components: Invalid since markers

Our components officially ship with `@since` markers (as can be seen [here](https://github.com/synergy-design-system/synergy-design-system/blob/main/packages/components/src/components/button/button.component.ts#L24)). However, those markers are the original ones from shoelace. This means they are completely off for most of the components, leading to wrong information.

For example, asking our MCP: "How long is syn-button available in Synergy?" will lead to an answer of `2.0`. However, the button was there from release `1.0` onwards always.

This is now fixed by adjusting all components with their corresponding correct markers as seen in the Changelog.

> This also means this will actually trigger a release of components and wrappers!

### Metadata: Component rules

Since the initial release of the metadata package, there is a [rules folder](https://github.com/synergy-design-system/synergy-design-system/tree/main/packages/metadata/config/rules) where we gather usage information for our customers and AI services. This was now adjusted to:

1. Include all missing components
2. Refactor the structure so it is easier to extend and reason about
3. Adjusted the Story Scraper for the new documentation style (see below)

Content is created by a mix of AI and me and can be seen as a first draft. External APIs are also extended so the data can be used externally. I also created a Markdown version that is used in the MCP directly.

### Docs: Better structure via custom overview pages

The rules created in the metadata package are also recreated as Storybook stories in this branch. Each component has its own Overview page associated with the given component. It handles formatting and linking automatically.

### MCP: Add support to provide the created rules

The MCP has multiple new capabilities:

1. `tool:component-info`: Component-Info can now be queried for the new `rules` layer. This will output the rules as Markdown
2. `prompt:explain-component-rules`: A prompt that can be used to create a new prompt message that explains the component rules. Can be used via `/mcp.synergy.explain-component-rules` (prefix may be different, depending on your instance name!).
3. `tool:intent*`: Tools that can be used to query intents of Synergy and also validate markup against them. This part is experimentally build into the MCP server and hidden behind a flag. Use the config (from readme) to see how to enable it. The current demo configuration is already configured to run it.

Prompts are now also part of the waterline checks.

### 🎫 Issues

Closes #1187 
Closes #977
Closes #1283 

## 👩‍💻 Reviewer Notes

- Have a really good look at the proposed rules and their structure. 

## 📑 Test Plan

Build everything in the root, open Storybook and have a look at the new overview pages. Those are the real visible items in this branch.

## ✅ DoD


packages/docs/src/helpers/rules-helper.tsx
```diff
@@ -0,0 +1,368 @@
...
+          <Markdown>{`# ${content.component}`}</Markdown>
+
+          <nav>
+            {coreMetadata?.status && (
+              <syn-tooltip>
+                <div slot="content">
+                  Current development status.
+                  <ul>
+                    <li style={{ color: 'var(--syn-color-neutral-0)' }}><strong>Stable:</strong> fully released and ready for production use.</li>
+                    <li style={{ color: 'var(--syn-color-neutral-0)' }}><strong>Warning:</strong> use with caution, may have breaking changes or be in early access.</li>
```
kirchsuSICKAG: I think we should differentiate between Breaking Changes and early access and not combine both in warning.
Another thought: do we really need the warning status for breaking changes?
For me I would have seen the states: "stable", "experimental" and "deprecated" 

```suggestion
                    <li style={{ color: 'var(--syn-color-neutral-0)' }}><strong>Deprecated:</strong> use with caution as this component will be removed in the future.</li>
                    <li style={{ color: 'var(--syn-color-neutral-0)' }}><strong>Experimental:</strong> use with caution as it is in early access and potentially gets changes in the future.</li>
```

packages/docs/src/helpers/rules-helper.tsx
```diff
@@ -0,0 +1,368 @@
...
+
+            {figmaComponentId && (
+              <syn-tooltip content="View Figma component">
+                <syn-icon-button
+                  href={`https://www.figma.com/design/bZFqk9urD3NlghGUKrkKCR/Synergy-Digital-Design-System?node-id=${figmaComponentId}`}
+                  src="/figma-logo.svg"
+                  label="Figma Component"
+                  size="large"
+                />
+              </syn-tooltip>
```
kirchsuSICKAG: I think it is enough to have only the figma link to the documentation instead of both, figma documentation and figma instance

packages/metadata/config/rules/component:syn-alert.js
```diff
@@ -0,0 +1,92 @@
...
+        'Danger variant: Use for critical errors, destructive actions, or situations requiring immediate response.',
+        'Neutral variant: Use for updates or system messages that are neither positive nor negative.',
+      ],
+      id: 'variants',
+      name: 'Variants & When to Use Them',
+    },
+    {
+      content: [
+        'Always include an appropriate icon that reinforces the message tone and helps with quick visual scanning.',
+        'Use system icons for consistency (info, check_circle, warning, status-error, settings).',
```
kirchsuSICKAG: In best case our users do not know anything about our system icons, therefore I would not state system icons here. Also we do not use system icons in our stories

packages/metadata/config/rules/component:syn-alert.js
```diff
@@ -0,0 +1,92 @@
...
+/** @type {import('../../src/config/types.js').ComponentRules} */
+export default {
+  accessibility: [
+    'Don\'t use alerts for trivial updates that don\'t require user awareness.',
+    'Ensure the alert content is concise and clear, focusing on the most critical information.',
+    'Use simple language that users can quickly understand without additional context.',
+    'Include action items or next steps when appropriate (e.g., "Your session ended. Please login again").',
+    'Avoid redundant information — don\'t repeat the title in the body text.',
+  ],
+  component: 'syn-alert',
```
kirchsuSICKAG, line 12: I like the common use cases of [solid](https://solid-design-system.fe.union-investment.de/docs/?path=/docs/components-sd-notification-overview--docs#usage-guidelines) to tell the user, when to use the toast and when the inline variant. I would add something similar to our docs

packages/metadata/config/rules/component:syn-badge.js
```diff
@@ -21,32 +21,47 @@ export default {
...
+        'Use default variants for standard notifications',
+        'Blue = neutral integration, green = more attention, red = maximum emphasis',
+        'Do not associate color variants with success or error states',
+      ],
+      id: 'color',
+      name: 'Color',
+    },
+    {
+      content: [
+        'Use only numbers as content',
```
kirchsuSICKAG: We do also allow text, so this is not correct for us

packages/metadata/config/rules/component:syn-badge.js
```diff
@@ -21,32 +21,47 @@ export default {
...
+      ],
+      id: 'background',
+      name: 'Background',
+    },
+    {
+      content: [
+        'Consistently apply one color choice for a cohesive look',
+        'Use default variants for standard notifications',
+        'Blue = neutral integration, green = more attention, red = maximum emphasis',
+        'Do not associate color variants with success or error states',
```
kirchsuSICKAG, line 34: Not sure if we can do a statement like this, when our variants are called 'success', 'danger', 'warning'

packages/metadata/config/rules/component:syn-breadcrumb.js
```diff
@@ -12,27 +12,43 @@ export default {
       'syn-breadcrumb-item',
     ],
     templates: [
-      'Application Shell',
+      'AppShell',
```
kirchsuSICKAG: Where does the link to the *Appshell* template come from? Currently there is no breadcrumb used in the appshell templares

packages/metadata/config/rules/component:syn-button-group.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+    {
+      content: [
+        'Small: Use for compact spaces, toolbars in tight layouts, or mobile interfaces.',
+        'Medium (default): Use for standard UI contexts and most common use cases.',
+        'Large: Use for prominent button groups that need extra visibility or accessibility.',
+        'Match button group size to the visual hierarchy and context of your page.',
+      ],
+      id: 'sizing',
+      name: 'Sizing',
```
kirchsuSICKAG: As a general input (not just for the syn-button-group, but for all components), I am not sure if we really want the point of "Sizing". For me it does not really make sense. But maybe I am alone with this feeling

packages/metadata/config/rules/component:syn-button-group.js
```diff
@@ -0,0 +1,91 @@
...
+    {
+      content: [
+        'Group related buttons that serve a common purpose or function together.',
+        'Use for actions that are logically connected (e.g., text formatting: bold, italic, underline).',
+        'Group buttons that control related settings or options within the same UI context.',
+        'Avoid grouping buttons that are unrelated or have conflicting purposes.',
+      ],
+      id: 'grouping',
+      name: 'When to Group Buttons',
+    },
```
kirchsuSICKAG: Wouldn't it make sense to move this point from *Usage Guidelines* to *Common use cases*?

packages/metadata/config/rules/component:syn-button-group.js
```diff
@@ -0,0 +1,91 @@
...
+      content: [
+        'Use for toolbars with icon buttons and tooltip hints.',
+        'Use for segmented controls with radio buttons in a group.',
+        'Use for text formatting controls (bold, italic, underline, etc.).',
+        'Use for pagination or navigation controls with multiple related actions.',
+        'Use for filter or view controls with related options.',
+      ],
+      id: 'patterns',
+      name: 'Common Patterns',
+    },
```
kirchsuSICKAG: Wouldn't it also make sense to move this to *Common use cases*?

packages/metadata/config/rules/component:syn-button-group.js
```diff
@@ -0,0 +1,91 @@
...
+    {
+      content: [
+        'Maintain visual consistency: all buttons in a group should have similar styling.',
+        'Don\'t mix different button variants within a single button group.',
```
kirchsuSICKAG: Redundant information. under heading *Variants* there is also the information *Keep variant consistent within a single button group for visual cohesion.*

packages/metadata/config/rules/component:syn-checkbox.js
kirchsuSICKAG: Should be removed, as we do not have a *syn-checkbox-group*

kirchsuSICKAG: Text should be updated, as we do not have a syn-checkbox-group

packages/metadata/config/rules/component:syn-combobox.js
kirchsuSICKAG: Should be removed, as this belongs to the two different label types of solid, which we do not have

kirchsuSICKAG: Should be removed, as we do not have the *visually-disabled* attribute

packages/metadata/config/rules/component:syn-dialog.js
kirchsuSICKAG: Wrong component name *syn-alert* instead of *syn-notification*

```suggestion
    'For non-critical content or information that doesn\'t require immediate user input, please use syn-alert to avoid workflow disruption.',
```

packages/metadata/config/rules/component:syn-drawer.js
```diff
@@ -9,29 +9,45 @@ export default {
...
+      content: [
+        'Ensure it does not cover critical content or actions on the main screen.',
+        'If using a tertiary action in the header, adjust the left padding to 0px for proper alignment.',
+      ],
+      id: 'layout',
+      name: 'Layout',
+    },
+    {
+      content: [
+        'Use the "header" slot to add navigation and/or action elements if desired.',
```
kirchsuSICKAG: We do not have a *header* slot in the syn-drawer, but additionally *label* and *header-actions*

packages/metadata/config/rules/component:syn-drawer.js
```diff
@@ -9,29 +9,45 @@ export default {
...
-      'Use the "header" slot to add navigation and/or action elements if desired.',
-      'Use the "default" slot to add main content. The "default" slot is always scrollable. For simplicity, syn-scrollable is integrated into the component as a wrapper for the slot in Figma, but will need to be used in code if desired.',
-      'Use the "footer" slot to add action elements if desired. The "footer" slot is always fixed.',
+      'AppShell',
     ],
   },
+  usageGuidelines: [
+    {
+      content: [
+        'Ensure the content is supplementary and does not include critical information that users must see immediately such as notifications or alerts; instead, use syn-notification or syn-dialog for those purposes.',
```
kirchsuSICKAG: Wrong component name, *syn-alert* instead of *syn-notification*


```suggestion
        'Ensure the content is supplementary and does not include critical information that users must see immediately such as notifications or alerts; instead, use syn-alert or syn-dialog for those purposes.',
```

packages/metadata/config/rules/component:syn-drawer.js
```diff
@@ -9,29 +9,45 @@ export default {
...
+        'Ensure it does not cover critical content or actions on the main screen.',
+        'If using a tertiary action in the header, adjust the left padding to 0px for proper alignment.',
+      ],
+      id: 'layout',
+      name: 'Layout',
+    },
+    {
+      content: [
+        'Use the "header" slot to add navigation and/or action elements if desired.',
+        'Use the "default" slot to add main content. The "default" slot is always scrollable. For simplicity, syn-scrollable is integrated into the component as a wrapper for the slot in Figma, but will need to be used in code if desired.',
```
kirchsuSICKAG: We do not have a *syn-scrollable*

packages/metadata/config/rules/component:syn-header.js
```diff
@@ -14,26 +14,38 @@ export default {
...
-      'Refrain from customizing elements in ways that deviate from the brand identity.',
-      'Include the company logo prominently; do not use a header without the corresponding logo.',
+      'AppShell',
     ],
   },
+  usageGuidelines: [
+    {
+      content: [
+        'Place navigation items to the left, the logo at the top left, and meta navigation on the right in wider viewports to keep consistency across products.',
+        'In smaller viewports, place an expandable menu on the right to host navigation items while keeping the logo consistently on the left across all viewport sizes.',
```
kirchsuSICKAG: The expandable menu is in our case on the left side, not on the right

kirchsuSICKAG (changes requested):

packages/metadata/config/rules/component:syn-input.js
kirchsuSICKAG: Should be removed, as this belongs to the two different label types of solid, which we do not have

packages/metadata/config/rules/component:syn-input.js
```diff
@@ -17,28 +17,48 @@ export default {
...
+      id: 'validation_and_formatting',
+      name: 'Validation and Formatting',
+    },
+    {
+      content: [
+        'Do not mix the two variants (default / floating label) within the same product, flow or form. Read the use cases above to know when to use each type.',
+      ],
+      id: 'variant_consistency',
+      name: 'Variant Consistency',
+    },
```
kirchsuSICKAG: Should be removed as we do not have the two different label variants

packages/metadata/config/rules/component:syn-input.js
kirchsuSICKAG: We do not have the *visually-disabled* attribute

packages/metadata/config/rules/component:syn-nav-item.js
```diff
@@ -14,25 +14,41 @@ export default {
...
+    {
+      content: [
+        'Be consistent in the use of descriptions: Include them for all items at the same level and keep them similar in length.',
+      ],
+      id: 'text_styles_and_descriptions',
+      name: 'Text Styles and Descriptions',
+    },
+  ],
   useCases: [
     'Populate an syn-header navigation bar at the top of a page, helping users easily access different sections.',
```
kirchsuSICKAG: We should state the syn-prio-nav here instead of the syn-header

```suggestion
    'Populate a syn-prio-nav navigation bar at the top of a page, helping users easily access different sections.',
```

packages/metadata/config/rules/component:syn-option.js
```diff
@@ -18,16 +18,24 @@ export default {
...
-    ],
-    content: [
-      'Make sure all items in the list are mutually exclusive and unambiguous to help users understand what they are selecting.',
-      'Provide a default choice where applicable to guide users.',
-    ],
-  },
+  usageGuidelines: [
+    {
+      content: [
+        'Use the "checkbox" variant for multi-select and the "default" variant for single-select choices.',
```
kirchsuSICKAG: We do not have two different variants

packages/metadata/config/rules/component:syn-select.js
kirchsuSICKAG: Should be removed, as this belongs to the two different label types of solid, which we do not have

packages/metadata/config/rules/component:syn-select.js
```diff
@@ -16,24 +16,36 @@ export default {
...
+        'Avoid using the same word or phrase at the beginning of options.',
+        'Provide a meaningful placeholder option to guide users.',
+      ],
+      id: 'content_and_labels',
+      name: 'Content and Labels',
+    },
+    {
+      content: [
+        'Use for lists with more than five options; for fewer options, consider using syn-radio for single selection or syn-checkbox for multiple selection.',
+        'Apply the "multiple" attribute when multiple selections are allowed, and pair it with "checkbox" for corresponding nested options.',
```
kirchsuSICKAG: We do not have attribute *checkbox* on options, so the second part of this sentence should be removed


```suggestion
        'Apply the "multiple" attribute when multiple selections are allowed.',
```

packages/metadata/config/rules/component:syn-tab-group.js
```diff
@@ -13,31 +13,51 @@ export default {
...
+      id: 'behavior',
+      name: 'Behavior',
+    },
+    {
+      content: [
+        'Choose labels that are concise and use no more than two words.',
+        'Ensure the first tab is the most relevant for the user.',
+        'Place related tabs next to each other for logical grouping.',
+        'Avoid overloading each tabs content with too much information; keep it manageable and focused.',
+        'Avoid using tabs for content that needs to be read in a specific order; use syn-step-group instead.',
```
kirchsuSICKAG: Should be removed as we do not have a syn-step-group

packages/metadata/config/rules/component:syn-tab-group.js
```diff
@@ -13,31 +13,51 @@ export default {
...
+        'Bold labels if desired, keeping this decision consistent across the group.',
+        'Customize to suit user requirements by changing label alignment or deleting the separator between group and content.',
+      ],
+      id: 'styling',
+      name: 'Styling',
+    },
+    {
+      content: [
+        'Use syn-tab-panel optionally when applying the "default" tab group variant.',
+        'When applying the "container" tab group variant, syn-tab-panel is mandatory.',
```
kirchsuSICKAG: This does not really make sense for me. Isn't the syn-tab-panel always mandatory?

packages/metadata/config/rules/component:syn-header.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-drawer.js
schilchSICKAG: Changed to header-actions.

packages/metadata/config/rules/component:syn-drawer.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-drawer.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-tag-group.js
```diff
@@ -0,0 +1,53 @@
+// @ts-check
+
+/** @type {import('../../src/config/types.js').ComponentRules} */
+export default {
+  accessibility: [
+    'Ensure each tag has clear readable text and a meaningful remove action when removable.',
+    'Expose group labeling context when tags represent selected filters or categories.',
+    'Keep keyboard interaction predictable for tag focus and removal.',
```
kirchsuSICKAG: Currently we do not support keyboard interaction on tags (this is related to shoelace not allowing to focus on it, see this pr which was not accepted: https://github.com/shoelace-style/shoelace/pull/1741 )
Therefore i would remove this line

packages/metadata/config/rules/component:syn-dialog.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-combobox.js
schilchSICKAG: Right, removed it.

packages/metadata/config/rules/component:syn-textarea.js
kirchsuSICKAG: Should be removed, as this belongs to the two different label types of solid, which we do not have

packages/metadata/config/rules/component:syn-combobox.js
schilchSICKAG: Shortened it even more because the last two entries dont make sense to me at all :)

packages/metadata/config/rules/component:syn-checkbox.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-button-group.js
schilchSICKAG: Thats right, adjusted it

packages/metadata/config/rules/component:syn-button-group.js
schilchSICKAG: Maybe we should have a global rule when to use which sizes and dont do it component specific. Think you are right, removed the reference.

packages/metadata/config/rules/component:syn-button-group.js
schilchSICKAG: Right, was changed

packages/metadata/config/rules/component:syn-tooltip.js
```diff
@@ -13,29 +13,49 @@ export default {
...
+      content: [
+        'Display a headline by bolding the text if it makes the content easier to understand.',
+        'Emphasize key information by bolding parts of the text.',
+      ],
+      id: 'styling',
+      name: 'Styling',
+    },
+    {
+      content: [
+        'Use any interactive element as a tooltip trigger by placing it in the provided slot; the default trigger (info-i) can be replaced based on context or preference.',
```
kirchsuSICKAG: Solis has a default trigger element which they are showing (the "i" icon), this is why they stated it here. As we do not have a default trigger, we should adapt the text here


```suggestion
        'Use any interactive element as a tooltip trigger by placing it in the provided slot.',
```

packages/metadata/config/rules/component:syn-button-group.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-breadcrumb.js
schilchSICKAG: Auto generated, removed

kirchsuSICKAG (changes requested):

packages/metadata/config/rules/component:syn-badge.js
schilchSICKAG: Intent is to day: I need a danger variant, not just a red badge. Its the semantics that matter, not the color as danger could also be pink for example. But I understand where this is coming from, lets discuss in the team.

packages/metadata/config/rules/component:syn-badge.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-alert.js
schilchSICKAG: Removed the word system

packages/metadata/config/rules/component:syn-alert.js
schilchSICKAG: I tried to structure them into usageGuidelines so each variant has its own scope. We currently cannot use free text but only lists, thats why I opted for this approach.

packages/metadata/config/rules/component:syn-tooltip.js
schilchSICKAG: Adjusted

packages/metadata/config/rules/component:syn-textarea.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-tag-group.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-tab-group.js
schilchSICKAG: Removed

packages/metadata/config/rules/component:syn-tab-group.js
schilchSICKAG: I removed the whole section.

packages/metadata/config/rules/component:syn-select.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-select.js
schilchSICKAG: Removed

packages/metadata/config/rules/component:syn-option.js
schilchSICKAG: Removed

packages/metadata/config/rules/component:syn-nav-item.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-input.js
schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-input.js
schilchSICKAG: Fixed

schilchSICKAG: Fixed

packages/metadata/config/rules/component:syn-checkbox.js
schilchSICKAG: Fixed

packages/docs/src/helpers/rules-helper.tsx
schilchSICKAG: Removed

packages/metadata/config/rules/component:syn-button-group.js
kirchsuSICKAG: I like the idea of global rules 👍

kirchsuSICKAG (approved):

packages/metadata/config/rules/component:syn-accordion.js
```diff
@@ -4,6 +4,7 @@
 export default {
   accessibility: [
     'Use only for non-critical information. Hiding content can become a potential barrier, making content more challenging to discover.',
+    'Use “close-others” attribute to keep only one item from the group open at a time, reducing the amount of information displayed at once and therefore reducing the cognitive load on the user.',
     'For purely decorative images, ALT-tags should be left empty so that screen readers can bypass them and concentrate on conveying meaningful content.',
```
reichst1SICKAG: a11y for image alt tags is not related to accordion and could be removed here

packages/metadata/config/rules/component:syn-alert.js
```diff
@@ -0,0 +1,116 @@
...
+        'Avoid stacking more than 3-4 alerts simultaneously to prevent cognitive overload.',
+        'Reserve alerts for messages that meaningfully impact the user\'s current task or workflow.',
+      ],
+      id: 'avoidance',
+      name: 'What to Avoid',
+    },
+  ],
+  useCases: [
+    'Inline notifications can provide contextual feedback to users. They are directly embedded within the content of a page or form.',
+    'Toast notifications provide brief, unobtrusive messages to users through toast notifications that overlay the interface and do not necessarily need attention.',
```
reichst1SICKAG: "do not necessarily need attention." don't fit with the usage guidlines where feedback, important updates and Highlight critical information is mentioned

packages/metadata/config/rules/component:syn-alert.js
```diff
@@ -0,0 +1,116 @@
...
+        'Warning variant: Use when an action has unexpected consequences or requires user attention.',
+        'Danger variant: Use for critical errors, destructive actions, or situations requiring immediate response.',
+        'Neutral variant: Use for updates or system messages that are neither positive nor negative.',
+      ],
+      id: 'variants',
+      name: 'Variants & When to Use Them',
+    },
+    {
+      content: [
+        'Always include an appropriate icon that reinforces the message tone and helps with quick visual scanning.',
```
reichst1SICKAG: Icons are optional in Synergy, so "Always" ist to hard. last list item states 'icons are optional'

packages/metadata/config/rules/component:syn-alert.js
```diff
@@ -0,0 +1,116 @@
...
+      ],
+      id: 'dismissal',
+      name: 'Dismissal & Duration',
+    },
+    {
+      content: [
+        'Use small size for brief, non-intrusive messages or in constrained spaces.',
+        'Use medium size (default) for most standard alerts and notifications.',
+        'Use large size for important messages that need to command attention.',
+        'Ensure size choices are consistent with message importance and context.',
```
reichst1SICKAG: We are using sizes differently, not based on the importance but rather related to a general component or application size.

packages/metadata/config/rules/component:syn-badge.js
```diff
@@ -21,32 +21,46 @@ export default {
...
+        'Place where it preserves relation to the assigned element',
+        'Do not obscure any informative element or text',
+        'Do not alter the number formatting',
+      ],
+      id: 'placement',
+      name: 'Placement',
+    },
+  ],
   useCases: [
     'Indicate the number of items selected, such as when using a filter',
```
reichst1SICKAG: isn't this filter stuff done with a syn-tag?

packages/metadata/config/rules/component:syn-badge.js
```diff
@@ -21,32 +21,46 @@ export default {
...
-    ],
-  },
+  usageGuidelines: [
+    {
+      content: [
+        'Use light backgrounds: white, neutral-100, primary-100',
+        'Use primary background when inverted',
+      ],
+      id: 'background',
+      name: 'Background',
```
reichst1SICKAG: was remove for accordion. Missed here? Don't makes sense.

packages/metadata/config/rules/component:syn-breadcrumb-item.js
```diff
@@ -15,14 +15,22 @@ export default {
...
-    ],
-  },
+  usageGuidelines: [
+    {
+      content: [
+        'Use light background options like white, neutral-100 or primary-100.',
+      ],
+      id: 'background',
+      name: 'Background',
+    },
```
reichst1SICKAG: remove, see comment at syn-badge

packages/metadata/config/rules/component:syn-breadcrumb.js
```diff
@@ -12,27 +12,42 @@ export default {
...
-    ],
-  },
+  usageGuidelines: [
+    {
+      content: [
+        'Use light background options like white, neutral-100 or primary-100.',
+      ],
+      id: 'background',
+      name: 'Background',
+    },
```
reichst1SICKAG: remove, see above

packages/metadata/config/rules/component:syn-button.js
```diff
@@ -15,35 +15,59 @@ export default {
...
+      id: 'action_priority',
+      name: 'Action Priority',
+    },
+    {
+      content: [
+        'Use light background options like white, neutral-100, primary-100, or use a primary background when inverted.',
+      ],
+      id: 'background',
+      name: 'Background',
+    },
```
reichst1SICKAG: remove

packages/metadata/config/rules/component:syn-card.js
```diff
@@ -0,0 +1,51 @@
...
+        'Use action labels that clearly describe the result.',
+        'Limit actions to the most relevant options for the card context.',
+      ],
+      id: 'actions',
+      name: 'Actions',
+    },
+  ],
+  useCases: [
+    'Display grouped information such as KPIs, summaries, or previews.',
+    'Provide a reusable container for dashboard modules.',
```
reichst1SICKAG: Provide a reusable container for dashboard elements.

packages/metadata/config/rules/component:syn-checkbox.js
```diff
@@ -5,32 +5,38 @@ export default {
...
+        'Ensure each selection operates independently unless used for bulk actions.',
+        'List selections in a logical order, such as alphabetical or numerical.',
+        'Refrain from using a single checkbox when the action should take effect immediately - use syn-switch instead.',
+      ],
+      id: 'interaction_and_usage',
+      name: 'Interaction and Usage',
+    },
+  ],
   useCases: [
     'Used for selections that don\'t immediately activate.',
```
reichst1SICKAG: I don't understand the intention of this maybe this sencence is not complete?

packages/metadata/config/rules/component:syn-combobox.js
```diff
@@ -17,31 +16,40 @@ export default {
...
-  },
+  usageGuidelines: [
+    {
+      content: [
+        'Use light backgrounds such as white, neutral-100, or primary-100 for clarity and consistency across different layouts.',
+        'Keep focus styles clearly visible when the user navigates through suggestions with a keyboard or screen reader.',
+      ],
+      id: 'background',
+      name: 'Background',
+    },
```
reichst1SICKAG: remove

packages/metadata/config/rules/component:syn-combobox.js
```diff
@@ -17,31 +16,40 @@ export default {
...
-      'Filter available options in real time as the user types; highlight or bold matching text to indicate relevance.',
-      'Show a message (e.g., "No matches found") when no options align with the user\'s input.',
-      'Consider limiting the maximum number of displayed suggestions to avoid overwhelming users. We recommend displaying 6-8 (with scrolling for additional results).',
-    ],
-  },
+  usageGuidelines: [
+    {
+      content: [
+        'Use light backgrounds such as white, neutral-100, or primary-100 for clarity and consistency across different layouts.',
+        'Keep focus styles clearly visible when the user navigates through suggestions with a keyboard or screen reader.',
```
reichst1SICKAG: remove, we do this internslly inside the component

packages/metadata/config/rules/component:syn-combobox.js
```diff
@@ -17,31 +16,40 @@ export default {
...
+        'Provide a clear, descriptive placeholder (e.g., "Search or select an option...") to help users understand they can type and choose from suggestions.',
+        'Avoid repeating the same initial word in multiple suggestions to reduce scanning difficulty.',
+      ],
+      id: 'content_and_labels',
+      name: 'Content and Labels',
+    },
+    {
+      content: [
+        'Filter available options in real time as the user types; highlight or bold matching text to indicate relevance.',
+        'Show a message (e.g., "No matches found") when no options align with the user\'s input.',
```
reichst1SICKAG: isn't this already implemented?

packages/metadata/config/rules/component:syn-details.js
```diff
@@ -0,0 +1,53 @@
...
+      name: 'Content',
+    },
+    {
+      content: [
+        'Use light background options like white, neutral-100 or primary-100.',
+      ],
+      id: 'background',
+      name: 'Background',
+    },
+  ],
```
reichst1SICKAG: remove

packages/metadata/config/rules/component:syn-divider.js
```diff
@@ -7,18 +7,30 @@ export default {
...
-    ],
-  },
+  usageGuidelines: [
+    {
+      content: [
+        'Use light background options like white, neutral-100 or primary-100.',
+      ],
+      id: 'background',
+      name: 'Background',
+    },
```
reichst1SICKAG: remove

reichst1SICKAG (commented):
Reviewed components A-D. More to come next week

packages/metadata/config/rules/component:syn-divider.js
schilchSICKAG: Done

packages/metadata/config/rules/component:syn-details.js
schilchSICKAG: Done

packages/metadata/config/rules/component:syn-combobox.js
schilchSICKAG: Right :)

packages/metadata/config/rules/component:syn-accordion.js
schilchSICKAG: I also think its more of a global rule. Should really think about providing global a11y rules somehow.

packages/metadata/config/rules/component:syn-alert.js
schilchSICKAG: What is meant here is that toasts don´t have the same "weight" as inline notifications. They are used for small messages that dont need interaction, e.g. an information that a language switch was successful. In this case, the message may be shown as a toast that is unobstrusive to the user and does not need any more attention. Maybe its just bad phrasing on my part.

packages/metadata/config/rules/component:syn-alert.js
schilchSICKAG: Right, removed this.

packages/metadata/config/rules/component:syn-alert.js
```diff
@@ -0,0 +1,116 @@
...
+      ],
+      id: 'sizing',
+      name: 'Sizing',
+    },
+    {
```
schilchSICKAG: Removed the size section entirely

packages/metadata/config/rules/component:syn-badge.js
schilchSICKAG: Right, removed.

packages/metadata/config/rules/component:syn-badge.js
schilchSICKAG: Removed

packages/metadata/config/rules/component:syn-breadcrumb-item.js
schilchSICKAG: Done

packages/metadata/config/rules/component:syn-breadcrumb.js
schilchSICKAG: Done

packages/metadata/config/rules/component:syn-button.js
schilchSICKAG: Done

packages/metadata/config/rules/component:syn-card.js
schilchSICKAG: Done

packages/metadata/config/rules/component:syn-checkbox.js
schilchSICKAG: Its the opposite of syn-switch: A switch **directly** triggers an action (e.g. Use Dark Theme? -> when true, directly triggers the dark theme). Checkboxes don´t change anything on a page until a form is submitted.

packages/metadata/config/rules/component:syn-combobox.js
schilchSICKAG: Done

packages/metadata/README.md
```diff
@@ -81,6 +81,51 @@ if (component.data) {
...
+
+### CLI: Install Local Skills
+
+This package also provides a CLI command to generate self-contained Synergy skill bundles for local developer workflows.
+
+```bash
+npx @synergy-design-system/metadata install-skills --path .github/skills
+```
+
+The command supports both `--path ./dir` and `--path=./dir`, creates missing directories, and writes portable bundles under:
```
kirchsuSICKAG, line 118: I tried out the cli command in a test package, but it did not work. The install-skills.js script is no longer available for the build and packaged metadata package: 

Image

packages/static-docs-page/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Routes
+
+| Route            | Description                                                  |
+| ---------------- | ------------------------------------------------------------ |
+| `/`              | Home page with a brief introduction to the site              |
+| `/component/:id` | Detail page for every component in the Synergy design system |
+| `/template/:id`  | Detail page for every template                               |
+| `/style/:id`     | Detail page for every CSS utility style                      |
+| `/changelog`     | Overview of all package changelogs                           |
```
kirchsuSICKAG: There is no such overview page
```suggestion
```

packages/static-docs-page/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+Mirrors the template page structure:
+
+- **Header** — style name and timestamp.
+- **Description** — summary from the style metadata.
+- **Code samples** — fenced code blocks extracted from the style's markdown example layer.
+
+### Changelog pages (`/changelog` and `/changelog/:id`)
+
+- `/changelog` — overview table listing every package with a short description and a link to its full changelog.
```
kirchsuSICKAG: There is no such Overview page

```suggestion
```

packages/static-docs-page/README.md
```diff
@@ -0,0 +1,193 @@
...
+  pages/
+    index.astro            home page
+    component/
+      [id].astro           component detail pages
+    template/
+      [id].astro           template detail pages
+    style/
+      [id].astro           style detail pages
+    changelog/
+      index.astro          changelog overview page
```
kirchsuSICKAG: There is no such index.astro for the changelog

kirchsuSICKAG (changes requested):

packages/metadata/README.md
schilchSICKAG: Seems I have forgotten to use public interfaces only. Will be adjusted in the next version

packages/static-docs-page/README.md
schilchSICKAG: Fixed

packages/static-docs-page/README.md
schilchSICKAG: Fixed

packages/static-docs-page/README.md
schilchSICKAG: Fixed

packages/metadata/data/layers/rules/component/component:syn-header.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Layout
+
+- Place navigation items in the navigation slot, the logo at the left, and meta navigation on the right in wider viewports to keep consistency across products.
+- In smaller viewports, place an expandable menu on the left to host navigation items while keeping the logo consistently on the left across all viewport sizes.
+- Keep the navigation fixed at the top of the website when scrolling back to the top.
+
+### Navigation Structure and Content
+
+- Include the logo, main navigation, and, if applicable, sub-brand logos and meta navigation.
```
reichst1SICKAG: remove 'sub-brand logos' as we don't have or allow them as far as I know.

packages/metadata/data/layers/rules/component/component:syn-icon-button.md
```diff
@@ -0,0 +1,43 @@
...
+
+- Use icon buttons for quick actions and compact controls.
+- Prefer regular buttons when text improves clarity.
+- Use disabled state sparingly and explain why an action is unavailable when possible.
+
+### Size and Color
+
+- Use size variants consistently within a control group.
+- Use semantic colors only when state or priority must be emphasized.
+- Do not overuse high contrast icon buttons in low priority contexts.
```
reichst1SICKAG: we don't have a high contrast icon button

packages/metadata/data/layers/rules/component/component:syn-icon.md
```diff
@@ -0,0 +1,42 @@
+# syn-icon
+
```
reichst1SICKAG, line 2: description part in Storybook has "No Description" content
https://64f819d70a69cb8728e06daf-yeszniccjq.chromatic.com/?path=/docs/components-syn-icon-overview--docs

packages/metadata/data/layers/rules/component/component:syn-menu-item.md
```diff
@@ -0,0 +1,13 @@
+# syn-menu-item
+
```
reichst1SICKAG, line 2: Usge Guidelines missing. Remove Headline or add a "custom"?

packages/metadata/data/layers/rules/component/component:syn-nav-item.md
```diff
@@ -0,0 +1,45 @@
...
+- Be consistent in the use of descriptions: Include them for all items at the same level and keep them similar in length.
+
+## Accessibility
+
+- Provide a logical tab order that follows the visual sequence of navigation items and their different levels.
+- Use aria-current="page" or aria-current="location" on the active navigation item to communicate the user's current location.
+
+## Related Components
+
+- syn-header
```
reichst1SICKAG: the nav-item is not related to the header, imho. only side and prio makes sense for me.

packages/metadata/data/layers/rules/component/component:syn-nav-item.md
```diff
@@ -0,0 +1,45 @@
...
+
+- Populate an `<syn-prio-nav>` within a `<syn-header>` navigation bar at the top of a page, helping users easily access different sections.
+- Implement navigation items in a `<syn-side-nav>` for a more detailed and hierarchical navigation structure.
+
+## Usage Guidelines
+
+### Background Options
+
+- Use with background options of white, neutral-100, and primary-100.
+
```
reichst1SICKAG, line 17: remove

packages/metadata/data/layers/rules/component/component:syn-prio-nav.md
```diff
@@ -0,0 +1,47 @@
...
+### Priority Handling
+
+- Place the most important destinations first so they stay visible longer.
+- Move lower priority items into overflow menus when space is limited.
+- Review priority order regularly as product usage evolves.
+
+### Labels and Structure
+
+- Keep labels short and distinct to prevent truncation and ambiguity.
+- Avoid deep nesting in top level priority navigation.
```
reichst1SICKAG, line 27: nesting ist not possible with our prio nav

packages/metadata/data/layers/rules/component/component:syn-radio-button.md
```diff
@@ -0,0 +1,49 @@
...
+- Select one option from a group where one is already preselected.
+- Switch between groups of settings.
+- Filter views by a parent category.
+
+## Usage Guidelines
+
+### Background
+
+- Use with light background options such as white, neutral-100, or primary-100.
+
```
reichst1SICKAG, line 18: remove

packages/metadata/data/layers/rules/component/component:syn-radio-button.md
```diff
@@ -0,0 +1,49 @@
+# syn-radio-button
```
reichst1SICKAG, line 1: completely missing in Storybook

packages/metadata/data/layers/rules/component/component:syn-radio-button.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Usage Guidelines
+
+### Background
+
+- Use with light background options such as white, neutral-100, or primary-100.
+
+### Behavior
+
+- Don't use as input method in forms, for they are not form elements. Use syn-radio instead.
```
reichst1SICKAG, line 21: they are form elements, or not?

packages/metadata/data/layers/rules/component/component:syn-radio-button.md
```diff
@@ -0,0 +1,49 @@
...
+- Slot inside of an syn-radio-group.
+- Use only in groups, as they are designed to allow the user to activate one of several options. To work with a single option, use syn-checkbox instead.
+- Pre-select always a "default" value; there is no invalid state.
```
reichst1SICKAG, line 24: not true? we have an invalid state, also pre-select is not stated anywhwere in our components.

packages/metadata/data/layers/rules/component/component:syn-radio-group.md
```diff
@@ -0,0 +1,43 @@
...
+- Provide clear and concise choices in settings or configuration panels.
+- Enable users to make a selection in surveys or questionnaires.
+- Offer options in filter panels where only one filter can be applied at a time.
+
+## Usage Guidelines
+
+### Background
+
+- Use with light background options such as white, neutral-100, or primary-100.
+
```
reichst1SICKAG, line 19: remove

packages/metadata/data/layers/rules/component/component:syn-radio.md
```diff
@@ -0,0 +1,13 @@
+# syn-radio
```
reichst1SICKAG, line 1: Usage guidelines has no content. but visible headline
https://64f819d70a69cb8728e06daf-yeszniccjq.chromatic.com/?path=/docs/components-syn-radio-overview--docs

packages/metadata/data/layers/rules/component/component:syn-range-tick.md
```diff
@@ -0,0 +1,13 @@
+# syn-range-tick
```
reichst1SICKAG, line 1: Usage guidelines has no content. but visible headline
https://64f819d70a69cb8728e06daf-yeszniccjq.chromatic.com/?path=/docs/components-syn-range-tick-overview--docs

packages/metadata/data/layers/rules/component/component:syn-range.md
```diff
@@ -0,0 +1,50 @@
...
+- Select time intervals for scheduling or reminder tasks.
+- Set minimum and maximum inputs (e.g., age, quantity, distance).
+- Visualize progress by assigning milestones within a larger goal or process.
+
+## Usage Guidelines
+
+### Background
+
+- Use light background options like white, neutral-100 or primary-100.
+
```
reichst1SICKAG: remove

packages/metadata/data/layers/rules/component/component:syn-select.md
```diff
@@ -0,0 +1,46 @@
...
+- Enable users to select one or more options from a list of predefined choices in forms.
+- Implement in filter panels to allow users to refine search results or data views.
+- Provide options in settings pages where users need to choose preferences from a list.
+
+## Usage Guidelines
+
+### Background
+
+- Use light background options such as white, neutral-100, or primary-100.
+
```
reichst1SICKAG: remove

packages/metadata/data/layers/rules/component/component:syn-spinner.md
```diff
@@ -0,0 +1,42 @@
...
+- Used within buttons or interactive elements to indicate that an action is being processed.
+- Apply to sections of a page that are loading independently, such as widgets or panels.
+- Display when a form is being submitted to show that the submission is in progress.
+
+## Usage Guidelines
+
+### Background
+
+- Use light background options like white, neutral-100, primary-100, or use a primary background when inverted.
+
```
reichst1SICKAG: remove

packages/metadata/data/layers/rules/component/component:syn-tab-group.md
```diff
@@ -0,0 +1,44 @@
...
+- Group settings or configuration options into categories for easier navigation.
+- Display different aspects of a product, such as description, reviews, and specifications.
+- Arrange user profile information into sections like personal details, activity, and settings.
+
+## Usage Guidelines
+
+### Background
+
+- Use background options of white, neutral-100, or primary-100.
+
```
reichst1SICKAG: remove

packages/metadata/data/layers/rules/component/component:syn-tab-group.md
```diff
@@ -0,0 +1,44 @@
...
+- Choose labels that are concise and use no more than two words.
+- Ensure the first tab is the most relevant for the user.
+- Place related tabs next to each other for logical grouping.
+- Avoid overloading each tabs content with too much information; keep it manageable and focused.
+
+### Styling
+
+- Bold labels if desired, keeping this decision consistent across the group.
+- Customize to suit user requirements by changing label alignment or deleting the separator between group and content.
+
```
reichst1SICKAG: remove, is defined in the component

packages/metadata/data/layers/rules/component/component:syn-tab-panel.md
```diff
@@ -0,0 +1,10 @@
+# syn-tab-panel
```
reichst1SICKAG, line 1: Usage guidelines is empty
https://64f819d70a69cb8728e06daf-yeszniccjq.chromatic.com/?path=/docs/components-syn-tab-panel-overview--docs

packages/metadata/data/layers/rules/component/component:syn-tab.md
```diff
@@ -0,0 +1,10 @@
+# syn-tab
```
reichst1SICKAG, line 1: Usage guidelines is empty
https://64f819d70a69cb8728e06daf-yeszniccjq.chromatic.com/?path=/docs/components-syn-tab-overview--docs

packages/metadata/data/layers/rules/component/component:syn-tag.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Behavior
+
+- Use the "removable" attribute to include a small close "x" button next to the label.
+- Avoid using a standalone, non-removable tag, as its selected/unselected state could be unclear. Apply the "removable" attribute for those cases.
+
+### Interactivity
+
+- Tags are interactive elements which trigger an action.
+
```
reichst1SICKAG: no, only interactive if closable. Otherwise no interactivity (currently) possible

packages/metadata/data/layers/rules/component/component:syn-textarea.md
```diff
@@ -0,0 +1,35 @@
...
+- Enable users to write comprehensive reviews of products or services.
+- Gather extensive information about issues or requests.
+- Capture detailed responses in surveys or questionnaires.
+
+## Usage Guidelines
+
+### Background
+
+- Use with light background options such as white, neutral-100, or primary-100.
+
```
reichst1SICKAG: remove

packages/metadata/data/layers/rules/component/component:syn-tag.md
```diff
@@ -0,0 +1,36 @@
+# syn-tag
```
reichst1SICKAG, line 1: Accessibility missing

packages/metadata/data/layers/rules/component/component:syn-textarea.md
```diff
@@ -0,0 +1,35 @@
+# syn-textarea
```
reichst1SICKAG, line 1: Accessibility missing

packages/metadata/data/layers/rules/component/component:syn-tooltip.md
```diff
@@ -0,0 +1,51 @@
...
+- Expand abbreviations or acronyms that might be unfamiliar to users.
+- Offer detailed information about specific data points in charts or graphs.
+- Provide more context for error messages or warnings and anticipate any questions that users may have.
+
+## Usage Guidelines
+
+### Background
+
+- Use with light background options of white, neutral-100 and primary-100.
+
```
reichst1SICKAG: remove

reichst1SICKAG (commented):
Done all in /rules/component.
Looks very good and is a really meaningfull addition.

packages/metadata/config/rules/component:syn-checkbox.js
reichst1SICKAG: ah understood now. Then maybe replace the 'activate' with 'trigger an action' or something like this?

packages/metadata/config/rules/component:syn-alert.js
reichst1SICKAG: ok, mabey the KI can rephrase this better?

packages/metadata/config/rules/component:syn-checkbox.js
schilchSICKAG: Done

packages/metadata/config/rules/component:syn-alert.js
schilchSICKAG: Done :)

packages/metadata/data/layers/rules/component/component:syn-tooltip.md
schilchSICKAG: Done

packages/metadata/data/layers/rules/component/component:syn-textarea.md
schilchSICKAG: Added, based on syn-input

packages/metadata/data/layers/rules/component/component:syn-textarea.md
schilchSICKAG: Removed

packages/metadata/data/layers/rules/component/component:syn-tag.md
schilchSICKAG: Added based on syn-tag-group

packages/metadata/data/layers/rules/component/component:syn-tag.md
schilchSICKAG: Removed

packages/metadata/data/layers/rules/component/component:syn-tab.md
schilchSICKAG: Fixed

packages/metadata/data/layers/rules/component/component:syn-tab-panel.md
schilchSICKAG: Fixed

packages/metadata/data/layers/rules/component/component:syn-tab-group.md
schilchSICKAG: Removed

packages/metadata/data/layers/rules/component/component:syn-tab-group.md
schilchSICKAG: Done

packages/metadata/data/layers/rules/component/component:syn-spinner.md
schilchSICKAG: done

packages/metadata/data/layers/rules/component/component:syn-select.md
schilchSICKAG: done

packages/metadata/data/layers/rules/component/component:syn-range-tick.md
schilchSICKAG: Fixed

packages/metadata/data/layers/rules/component/component:syn-header.md
schilchSICKAG: Done

packages/metadata/data/layers/rules/component/component:syn-icon-button.md
schilchSICKAG: Removed last two statements.

packages/metadata/data/layers/rules/component/component:syn-icon.md
schilchSICKAG: Fixed.

packages/metadata/data/layers/rules/component/component:syn-menu-item.md
schilchSICKAG: Fixed

packages/metadata/data/layers/rules/component/component:syn-nav-item.md
schilchSICKAG: Removed

reichst1SICKAG (approved):

packages/metadata/src/intentPolicy/domains/navigation/capabilities.ts
```diff
@@ -0,0 +1,56 @@
+import type { IntentCapability } from '../../types.js';
+
+export const navigationCapabilities: IntentCapability[] = [
+  {
+    categories: ['navigation', 'action'],
```
kirchsuSICKAG, line 5: Shouldn't the syn-link then also be part of the navigation capabilities file, if it is part of the category *'navigation'*?
I would either think, that here only the 'navigation' category is part of the array and the same structure but with *action* is part of the */domains/action/capabilities.ts*, or that here are both categories are in the array, but the same object for *syn-link* is also part of the action/capabilities.ts file

packages/metadata/src/public/domains/intent-policy.ts
```diff
@@ -0,0 +1,1259 @@
...
+    }
+  }
+};
+
+/**
+ * Validate whether a component usage aligns with intent-policy guidance.
+ *
+ * @experimental This API is still under active design.
+ */
+export const validateComponent = async (
```
kirchsuSICKAG, line 955: This is something, which could be an improvement in the future, but a thing that could be improved is, that currently the validation also comes with false negative values, because in the validation the default values of properties are not taken into account.

As example I have following code for the syn-button:

```html
<syn-button> My primary button </syn-button>
```

I validate this button with the intent **action.primary**.
Actually this code is completely correct, as the default of the "type" property is "button" and the default of the "variant" property is "filled".

But the validation tool says following:

Image

Image


For the future the default values of an property should also be taken into account.

kirchsuSICKAG (changes requested):

packages/metadata/src/public/domains/intent-policy.ts
schilchSICKAG: Ah, thats a good catch! I also plan to make types stricter, e.g. component should draw from HTMLDefinedElements. This could be done on a per component base, e.g. instead of `IntentUsagePattern` we would have `IntentUsagePattern<P>`. The rule for `actionPrimaryPattern` could then be `IntentUsagePattern<SynButton>` and I might be able to infer the default props from there.

packages/metadata/src/intentPolicy/domains/navigation/capabilities.ts
schilchSICKAG: Hmm, thats a good point. The original structure was much flatter (e.g. intentPolicy/capabilities/action.ts,navigation.ts), but reused the names of patterns everywhere (so there was a navigation.ts in capabilities, in actions, in categories etc). I think a domain structure like the one we now have needs to be in multiple domains. I am not sure how to progress on this further, each possibility has some caveats:

Copying the link: We will have duplicates
Leaving it in its major intent: We will have to guess all the time what is the primary.
Having a flat capability structure (one folder for them all) -> Easier to share the capabilities, but needs to be wired into them all.

What do you think?

kirchsuSICKAG: I think I would rather do one big capabilities file with all components and their categories as array and just filter the wanted category from this array. But I am not sure what would be the caveats or if this a big thing in the end for the implementation

schilchSICKAG: Yeah something like that:

```bash
capabilities:
  button.ts <- all the things button
  alert.ts <- all the things alert
  index.ts
```

and then:

```bash
domains/
  action/
    capabilities.ts
```

The capabilities could then either filter all intents (easy way, lazy) or import just those that you want to work. We could then remove the `categories` from `IntentCapabilities` and do the mapping in there.

Both would be possible. Question is if we want to do something like that now or if it has time for later refinements when the system grows.

kirchsuSICKAG: Do we really need the separate file? I just tested it out: removed all the single files and do one single capabilities file and it seems there is only one place, where the single capabilities are used and merged into one big array.
And it seems that only the whole summarized **targetCapabilities** array is used and nowhere a single capability like e.g. **actionCapabilities**. But maybe I overlook something

Bildschirmfoto 2026-05-21 um 10 47 14

schilchSICKAG: Reason is I dont want one big capabilities file. I had this originally - it gets large really fast

kirchsuSICKAG (approved):

PR #1279, opened 2026-05-05
fix: 🐛 syn-select: Selected option text not updating automatically when option label text is updated

### LINKED ISSUE #1265: fix: 🐛 syn-select: Selected option text not updating automatically when option label text is updated

# Version

The bug occurred with the following version of the component library: All

# Component

The issue arises when using the following component(s): syn-select, syn-combobox (not checked, but likely as they share a lot of code)

# Environment

The issue occurred in the following framework: Angular (propably others too!)

# Description

See [this stackblith](https://stackblitz.com/edit/angular-21-base-w2sxyzxc?file=src%2Fapp%2Fapp.component.html) for a easy reproducable error. Basically, when updating an options label (e.g. if the model changes that renders `<syn-option>`, but the value stays the same, the selected item in `<syn-select>` will **not** display the updated content from the selected item.

# Acceptance criteria

- [ ] Selected entries label updates automatically when the option label is updated

schilchSICKAG:
# Pull Request

## 📖 Description

This PR ensures `<syn-select>` and `<syn-combobox>` always show the current selected option text. If a selected option name changes, the field now updates right away instead of showing outdated text.

### 🎫 Issues

Closes #1265 

## 👩‍💻 Reviewer Notes

The issue was that `slotchange` does not fire if the nodes get reused. Because of this, the value was never updated. I added a `MutationObserver` for syn-select and updated the already existing one for `syn-combobox`, as well as adding regression tests.

## 📑 Test Plan

Check the local demos (AllComponents -> Combobox / Select). I added a Demo for bug. When clicking the button underneath the select, it will update the content of the second syn-option with a new value (once). You will notice that it now updates the display label when the option text changes. This is also true when using html in the `<syn-option>` slot.

## ✅ DoD


packages/components/src/components/combobox/combobox.component.ts
```diff
@@ -389,15 +389,36 @@ export default class SynCombobox extends SynergyElement implements SynergyFormCo
         return entry.oldValue !== currentValue && !!currentValue;
       });
 
+      const hasRelevantSelectedLabelChange = this.restricted && entries.some(entry => {
```
kirchsuSICKAG: We need to check for both, either `this.restricted` or `this.multiple`. Currently it does not update for this.multiple (see gif)


```suggestion
      const hasRelevantSelectedLabelChange = (this.restricted || this.multiple) && entries.some(entry => {
```

Image

Example code for testing (in angular):

```ts
import { Component, signal } from '@angular/core';

@Component({
  selector: 'home',
  standalone: false,
  template: `
  <syn-input
  label="Change name for selected item:"
  (synChangeEvent)="setValueName(selectedValue().id, $any($event.target).value)"
  [value]="selectedValue().name"
/>
<syn-combobox
  label="Select an item:"
  [value]="selectedValue().id"
  (synChangeEvent)="selectValue($any($event.target).value)"
  multiple
>
  @for (v of values; track v.id) {
  <syn-option [value]="v.id">{{ v.name }}</syn-option>
  }
</syn-combobox>
  `
})
export class Home {
protected values = [
    { id: 'value1', name: 'Value 1' },
    { id: 'value2', name: 'Value 2' },
    { id: 'value3', name: 'Value 3' },
  ];
  protected selectedValue = signal<{ id: string; name: string }>({
    id: 'value1',
    name: 'Value 1',
  });

  selectValue(id: string) {
    this.selectedValue.set(
      this.values.find((v) => v.id === id) ?? this.values[0]
    );
  }

  setValueName(id: string, name: string): void {
    const value = this.values.find((v) => v.id === id);
    if (value) {
      value.name = name;
      if (this.selectedValue().id === id) {
        this.selectedValue.set({ ...value });
      }
    }
  }
}


```

kirchsuSICKAG (changes requested):

packages/components/src/components/combobox/combobox.component.ts
schilchSICKAG: Good catch. Added it and also wrote two new regression tests to make sure it works with multiple, too. I also noticed a strange behavior when switching from `<syn-select value="option-2" />` to `<syn-select value="option-2" multiple />`: If multiple is set but a value is preselected, the `value` property is not updated and still returns the original `value` as string. I will create a separate ticket for this.

kirchsuSICKAG (approved):

PR #1281, opened 2026-05-05
fix: 🐛 syn-range missing syn-change event after programmatic value update

### LINKED ISSUE #1272: fix: 🐛 syn-range missing syn-change event after programmatic value update

# Version

The bug occurred with the following version of the component library: 3.12.1

# Component

The issue arises when using the following component(s): syn-range

# Environment

# Description

There is an issue with the `syn-range` slider when combining externally controlled `value` updates, user input and the `syn-change` event.

Use this codepen to check out the behavior: https://codepen.io/xfestek/pen/myObGoV?editors=1111

##  Current Behavior
1. User moves the range manually (e.g. from 50 → 20)  
   → `syn-change` is fired as expected

2. The slider is reset programmatically (e.g. via a reset button setting `value = 50`)  
   → no `syn-change` event is fired (expected behavior)

3. User moves the slider again to a value that was previously selected (e.g. back to 20)  
   →  **no `syn-change` event is fired**

### Expected behavior

The `syn-change` event should fire whenever the user actively changes the slider value — regardless of whether that value was previously selected before a programmatic reset.

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue where `syn-change` events where not fired for subsequent user interactions after a value was set in code.

### 🎫 Issues

Closes #1272 

## 👩‍💻 Reviewer Notes

Have a look at the provided e2e test demos. You may also want to try it per hand, e.g. by attaching the syn-change listener in storybook via dev console and watching it fire.

Initial reason for the issue was that we did not update the old value when programatically using the setter. With the two line fix applied this is now also taken into account and seems to work fine.

## ✅ DoD


packages/_private/e2e-demo-test/src/AllComponents/SynRange.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      x: targetX,
+      y: targetY,
+    },
+  });
+};
+
+test.describe('<SynRange />', () => {
+  createTestCases(({ name, port }) => {
+    test.describe(`Regression#1272: ${name}`, () => {
+      test('should emit syn-change for programmatic updates and when dragging back to the baseline value', async ({ page }) => {
```
kirchsuSICKAG, line 42: With the current test, the "old" code would not have been failed.
Also the order of the steps is not how to bug could be reproduced.

## Original steps
1. move slider to value (e.g. 20)
2. reset value programmatically ( e.g. 80)
3. move slider back to the same first value (20)

## Current test steps
1. Reset value programmatically (80)
2. drag to another value (14)
3. drag to programmatic value (80)

Here would be a test, which does fail for the old code but succeeds for the new code. But additionally the code for the **dragRangeToRatio** function needs to be updated, to use the thumb to drag instead of the whole range object. Otherwise the **dragTo** function makes the syn-change event fire twice for the second drag and therefore the "old" code would also succeed.

```ts
const dragRangeToRatio = async (range: Locator, ratio: number) => {
  const box = await range.boundingBox();
  if (!box) {
    throw new Error('Could not determine range dimensions for drag action.');
  }
  const thumb = range.locator('.thumb');

  const clampedRatio = Math.max(0, Math.min(1, ratio));
  const targetX = Math.max(1, Math.min(box.width - 1, box.width * clampedRatio));
  const targetY = box.height / 2;

  await thumb.dragTo(range, {
    targetPosition: {
      x: targetX,
      y: targetY,
    },
  });
};

test.describe('<SynRange />', () => {
  createTestCases(({ name, port }) => {
    test.describe(`Regression#1272: ${name}`, () => {
      test('should emit syn-change for programmatic updates and when dragging back to the baseline value', async ({ page }) => {
        const AllComponents = new AllComponentsPage(page, port);

        // 1. Load and fully open the range section in All Components.
        await AllComponents.loadInitialPage();
        await page.locator('#tab-Range').click();
        await expect(page.locator('#tab-content-Range')).toBeVisible();

        const range = AllComponents.getLocator('range1272ProgramaticValueChange');
        const button = AllComponents.getLocator('range1272ChangeValueButton');

        // 2. Create a listener that counts syn-change emissions.
        await range.evaluate((el) => {
          const elementWithCounter = el as SynRangeWithCounter;
          elementWithCounter.__synChangeCount = 0;
          elementWithCounter.addEventListener('syn-change', () => {
            elementWithCounter.__synChangeCount = (elementWithCounter.__synChangeCount ?? 0) + 1;
          });
        });

        // 3. Drag thumb to a new value.
        await dragRangeToRatio(range, 0.15);
        await expect(range).toHaveJSProperty('value', '14');
        const synChangeCountAfterFirstDrag = await getSynChangeCount(range);
        expect(synChangeCountAfterFirstDrag, 'Drag should fire syn-change').toBe(1);

        // 4. Click the button to programmatically change the value.
        await button.click();
        const synChangeCountAfterProgrammatic = await getSynChangeCount(range);

        // 5. Ensure syn-change does not fire for programmatic value changes.
        expect(synChangeCountAfterProgrammatic, 'Programmatic value change should not fire syn-change').toBe(synChangeCountAfterFirstDrag);

        // 6. Drag back to the first dragged value and ensure syn-change fires again.
        await dragRangeToRatio(range, 0.15);
        const valueAfterSecondDrag = await range.evaluate((el: SynRange) => Number(el.value));
        const synChangeCountAfterSecondDrag = await getSynChangeCount(range);
        expect(valueAfterSecondDrag).toBe(14);
        expect(synChangeCountAfterSecondDrag).toBe(2);
        expect(synChangeCountAfterSecondDrag).toBeGreaterThan(synChangeCountAfterFirstDrag);
      }); // Test syn-change emission
    }); // Regression#1272
  }); // End of test cases
});
```

kirchsuSICKAG (changes requested):

packages/_private/e2e-demo-test/src/AllComponents/SynRange.spec.ts
schilchSICKAG: You are right, just checked and works way better. I updated to use this code and also included latest main. Feel free to check again or approve.

kirchsuSICKAG (approved):

PR #1291, opened 2026-05-29
feat: ✨Add chart axes mvp

### LINKED ISSUE #1203: feat: ✨ chart axes (mvp)
# Component

# Story

As a user of the Synergy Design System, I would like to use axes that a are styled according to the Figma design guide.

# Acceptance criteria

- [ ] All features/styles below are defaults (to match design guideline) and can be overwritten (and are documented):
  - [ ] Category axis
    - [ ] labels
    - [ ] icons
    - [ ] title
  - [ ] Values axis
    - [ ] labels
    - [ ] title

### LINKED ISSUE #1261: research: 🧐 find best way to create configurations for charting configs
# Story

As a user of the Synergy Design System, I would like to have an easy and intuitive way to create charts.

# Acceptance criteria

- [ ] factories
- [ ] default configs
- [ ] single / mixed charts
- [ ] good defaults
- [ ] maximum flexibility
- [ ] good DX

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds a global synergy theme that is used from echarts.
The first default styling for chart axis mvp was added. 
Additionally it introduces a middleware-style config composition system and a set of predefined config modifiers for controlling axis appearance — all styled according to the Figma design guide.

### 🎫 Issues
Closes #1203 
Closes #1261 

## 👩‍💻 Reviewer Notes
Have a look at the middleware style config composition system. Has it a good DX?

## Open Question
Design needs to define, what should happen, if the axes labels do not have the same width for the y-axis. Should they be right-aligned or centered or,...

## 📑 Test Plan
Have a look at storybook and play around with the predefined configs.

## ✅ DoD


.changeset/1203-tokens.md
```diff
@@ -0,0 +1,12 @@
+---
+"@synergy-design-system/tokens": minor
+---
+
+feat: ✨ Add new component and chart tokens
+
+**Component tokens**
+- `--syn-typography-color-text-quiet`
+- `--syn-typography-color-text-quiet-inverted`
```
schilchSICKAG, line 12: Do we display those anywhere in Sorybook or are they just internal representation?

kirchsuSICKAG: Currently not. It is the same with *--syn-typography-color-text* and *--syn-typography-color-text-inverted*. They are there and used internally but not documented

kirchsuSICKAG: But it would be the question if we should add all tokens (also the component tokens like *syn-alert-error-color-background*, *--syn-alert-error-color-border*, ... ) at some point

packages/docs/src/docs-code-optimizer/index.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Optimizes docs preview source by stripping ignore sections and empty string assignments.
+ * - Removes blocks between HTML markers: <!-- preview-ignore:start --> ... <!-- preview-ignore:end -->
+ * - Removes blocks between JS markers: // preview-ignore:start ... // preview-ignore:end
+ * - Removes empty assignments like = ""
+ */
+export default function docsCodeOptimizer(code: string) {
+  const optimizedCode = code
+    .replace(/<!-- preview-ignore:start -->[\s\S]*?<!-- preview-ignore:end -->/g, '')
+    .replace(/\/\/ preview-ignore:start[\s\S]*?\/\/ preview-ignore:end/g, '')
+    .replace(/\s*=\s*""/g, '');
```
schilchSICKAG: I wonder if this might break some stories. A story like the following could potentially cause issues here:

<div id="something"></div>
<script type="module">
let content = "";

content = Math.random();
document.querySelector('#something').innerHTML = content;
</script>

This would result in the following story:

<div id="something"></div>
<script type="module">
let content;

content = Math.random();
document.querySelector('#something').innerHTML = content;
</script>

I am not sure if its worth the removal. I like the ability to remove blocks of code from MCP, but this one triggers me :) What do you think?

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -0,0 +1,344 @@
...
+          <Title />
+          <Subtitle />
+          <Description />
+          <Stories title="" />
+        </>
+      ),
+    },
+  },
+  tags: ['Charting', 'Data Visualization'],
+  title: 'Charts/Axes',
```
schilchSICKAG: I wonder if it is possible to have syn-chart as the first item? Depending on how we want to structure in the future, it might be easier to have something like Charts/Features/Axes or something like that? We will have much more of those substories, right?

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -0,0 +1,344 @@
...
+export const AxesLabelsWithIcons: Story = {
+  parameters: {
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'axes-labels-with-icons'),
+      },
+    },
+  },
+  render: () => html`
+      <div class="controls">
```
schilchSICKAG: Wrong indention

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -0,0 +1,344 @@
+import React from 'react';
```
schilchSICKAG, line 1: On initial load, I was not able to see the axes at all. Don´t know if this is problem on my machine. After clearing the cache it worked, but before I wondered what those stories do at all :).

Just tested it with another browser and empty Cache and I don´t see any labels or Icons in there, too. I need to hard reset via Cache clearing to receive the icons.

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -0,0 +1,344 @@
...
+        chart.config = enhanceConfig(baseConfig)
+          .usePreset('axes.hide-labels')
+          .usePreset('axes.split-lines')
+          .build();
+      </script>
+    `
+  ,
+};
+
+export const AxesLabelsWithIcons: Story = {
```
schilchSICKAG, line 250: This story is also only loading the items when they are already cached, which means you don´t get any icons to show up when first rendering the story. After a second render, they will show up (likely cached?)

packages/components/src/components/chart/configs/axes/types.ts
```diff
@@ -0,0 +1,36 @@
...
+export type AxisKey = 'xAxis' | 'yAxis';
+
+type UnpackedArray<T> = T extends Array<infer U> ? U : T;
+
+export type AxisOption<T extends AxisKey> = UnpackedArray<NonNullable<ECConfig[T]>>;
+
+export type AxisIndexSelection = number | number[];
+
+export type AxisPatchOptions = {
+  axisIndex?: AxisIndexSelection;
```
schilchSICKAG: Would it make sense to add comments what those properties actually do? For example I think I get what xAxisIndex should do, but it does not tell why I should set it.

packages/components/src/components/chart/configs/axes/utilities.ts
```diff
@@ -0,0 +1,399 @@
...
+ * - Category axes: uses the explicit `yAxis.data` array.
+ * - Value axes: approximates with the min/max of all numeric series data points,
+ *   which tends to represent the widest labels ECharts will render.
+ * - Multiple y-axes: combines label candidates from all configured y-axis entries.
+ *
+ * @param config - The current chart config.
+ * @returns Label texts that can be used for width estimation of rich y-axis labels.
+ */
+export function extractYAxisLabelTexts(config: ECConfig): string[] {
+  const getDataFromAxis = (axis: unknown): string[] => {
```
schilchSICKAG: This function gets redefined whenever `extractYAxisLabelTexts` is called. Maybe we should move it out of scope as it does not use any local variables at all?

packages/components/src/components/chart/configs/config.ts
```diff
@@ -0,0 +1,139 @@
...
+// Low-level deep-merge primitives
+// ---------------------------------------------------------------------------
+
+type Mergeable = Record<string, unknown>;
+type ConfigLayer = Partial<ECConfig> | undefined;
+
+const isMergeableObject = (value: unknown): value is Mergeable => typeof value === 'object' && value !== null && !Array.isArray(value);
+
+export const mergeObjects = (target: Mergeable, source: Mergeable): Mergeable => {
+  const merged = { ...target };
```
schilchSICKAG: When using `structuredClone`, we could remove the array check because each item will be cloned. Might save some performance if it works :)

packages/components/src/components/chart/chart.component.ts
```diff
@@ -173,8 +200,8 @@ export default class SynChart extends SynergyElement {
       <div
         part="base"
         class=${classMap({
```
schilchSICKAG: Broken indention. Also, if we don´t need multiple classes on this, I think it would be best to remove classMap utility alltogether.

packages/components/CHARTING_README.md
```diff
@@ -173,19 +173,89 @@ chart.getInstance()?.setOption(
...
+    }),
+  )
+  .build();
+```
+
+`enhanceConfig(base)` also supports named presets via `.usePreset(name, options)`.
+
+```js
+chart.config = enhanceConfig(baseConfig)
+  .usePreset("axes.split-lines")
```
schilchSICKAG: Would it be a good idea to have a table that shows all presets, together with a short description of what it does?

schilchSICKAG: Should have read further :)

packages/tokens/scripts/formats/css-vars.js
```diff
@@ -172,7 +183,12 @@ export const cssVariableFormatter = {
           convertOriginalToCssVarRecursive(dict[key]);
         } else {
           const name = dict[key]?.name;
-          if (name && BRAND2025_IGNORE_PATTERNS.includes(name)) {
+
+          // TODO: currently we have a double token name (syn-tooltip-background-color) in components and charts. We need to remove it from charts manually until design renames it.
```
schilchSICKAG: I think this makes it a hard blocker until Design renames, right?

schilchSICKAG (changes requested):
Looks pretty good, except for the strange reload thing in Storybook.

packages/docs/src/docs-code-optimizer/index.ts
kirchsuSICKAG: Theoretically we could remove the whole code-optimizer as I removed the import for your proposal with the "window." for the axes.stories.ts.

How you like best, I could either remove all or just the `.replace(/\s*=\s*""/g, '');`

packages/docs/stories/charts/axes.stories.tsx
kirchsuSICKAG: Should be fixed now with the new preprocessor handling

packages/docs/stories/charts/axes.stories.tsx
kirchsuSICKAG: Yes, there will be more of them. 
I like the idea of having syn-chart as first item and doing the other stuff under another subfolder.
I will have a look if i can sort it somehow

packages/components/src/components/chart/configs/config.ts
kirchsuSICKAG: I am thinking about using a library like lodash.merge or deepmerge, instead of creating an own deep merge function. What do you think @schilchSICKAG ?

kirchsuSICKAG: Also we could not use structuredClone here, as we also have callbacks in the options, and then it will fail:

DataCloneError: Failed to execute 'structuredClone' on 'Window': (value, i) => `{icon_${i}|}\n{label|${value}}` could not be cloned.

packages/tokens/scripts/formats/css-vars.js
kirchsuSICKAG: It is not a blocker for this PR, as this token is not yet needed.

packages/components/src/components/chart/configs/axes/types.ts
kirchsuSICKAG: Done

packages/docs/src/docs-code-optimizer/index.ts
kirchsuSICKAG: For now i just removed the replacement of an empty string

packages/docs/stories/charts/axes.stories.tsx
kirchsuSICKAG: I found a way for sorting the different chart sections. Now *syn-chart* is before *features*

packages/docs/stories/charts/axes.stories.tsx
kirchsuSICKAG: done

packages/docs/stories/charts/axes.stories.tsx
kirchsuSICKAG: Should be fixed now with the new preprocessor handling

packages/components/src/components/chart/configs/axes/utilities.ts
kirchsuSICKAG: done. Moved the function outside as separate function

packages/components/src/components/chart/chart.component.ts
kirchsuSICKAG: done

.changeset/1203-tokens.md
schilchSICKAG: Adding them all would also mean to document them, so I don´t think that we have the time to do this (yet).

packages/docs/src/docs-code-optimizer/index.ts
schilchSICKAG: Thats fine for me, so at least we can use it later on.

packages/components/src/components/chart/configs/config.ts
schilchSICKAG: Pulling in lodash for this is a bit overkill I think, especially because lodash.merge might have its own utility, but uses other lodash utilities underneath. Would rather write this by ourselfs as a one off. Maybe we already have something like that in the components package, might very well be.

packages/docs/stories/charts/axes.stories.tsx
schilchSICKAG: Works fine now

packages/components/src/components/chart/chart.component.ts
```diff
@@ -103,7 +104,13 @@ export default class SynChart extends SynergyElement {
...
-      this.chartInstance.setOption({ color: colors }, { notMerge: false });
+      const oldOption = this.chartInstance.getOption();
+      if (!oldOption) return;
+
+      oldOption.color = colors;
+      // We can not only replace 'color' via { replaceMerge: ['color'] }. Echarts does not allow to do this.
+      // Therefore we need to completely replace it, with the old option but new color.
+      this.chartInstance.setOption(oldOption, { notMerge: true });
     }
   }
```
schilchSICKAG, line 145: Maybe we can do this a bit lighter as all of your checks need chartInstance:

```typescript
  protected updated(changedProperties: PropertyValues<this>): void {
    if (changedProperties.has('config') && this.chartInstance) {
      this.chartInstance.setOption(this.config, { notMerge: true });
    }
    if ((changedProperties.has('palette') || changedProperties.has('config')) && this.chartInstance) {
      this.applyPalette();
    }
  }
```

```typescript
  protected updated(changedProperties: PropertyValues<this>): void {
    if (!this.chartInstance) {
      return;
    }

    if (changedProperties.has('config')) {
      this.chartInstance.setOption(this.config, { notMerge: true });
    }
    if ((changedProperties.has('palette') || changedProperties.has('config'))) {
      this.applyPalette();
    }
  }
```

packages/components/src/components/chart/chart.component.ts
```diff
@@ -116,10 +123,21 @@ export default class SynChart extends SynergyElement {
...
+    /**
+     * Depending if x-axis or y-axis, the axis name has different positions and alignments. This preprocessor ensures that the correct styles are applied to the axis names based on the axis type.
+     * This is needed because ECharts does not provide a way to set specific styles for x and y axis, only for axis types.
+     */
+    registerPreprocessor(applyAxisDefaultsPreprocessor);
+  }
+
   // Initialize echarts instance and resize observer
   protected firstUpdated(_changedProperties: PropertyValues): void {
     if (this.chartContainer !== null && this.chartContainer !== undefined) {
```
schilchSICKAG, line 168: I personally am a fan of reversing such statements as indention gets clearer:

```typescript
// Original
if (this.chartContainer !== null && this.chartContainer !== undefined) {

// Reversed:
if (this.chartContainer === null || this.chartContainer === undefined) {
  return;
}

// Original code with less indents

File: packages/components/src/components/chart/chart.component.ts [hunk 15]
```diff
@@ -116,10 +123,21 @@ export default class SynChart extends SynergyElement {
...
-      this.chartInstance = init(this.chartContainer);
+      this.chartInstance = init(this.chartContainer, 'default');
```
schilchSICKAG, line 169: I wonder if it may be beneficial for our users to have events to listen to for stuff like initializing, updating the theme or other changes in the library that may need attention. Something like:

```typescript
this.chartInstance = init(this.chartContainer, 'default');
this.emit('syn-chart-initialized', {
  chart: chartInstance,
});
```

There are certainly more examples that could benefit from this, for example something like a `syn-chart-theme-updated` event that is called when the theme updates so users can register on it. What do you think?

packages/components/src/components/chart/configs/axes/utilities.ts
```diff
@@ -0,0 +1,422 @@
...
+  }
+
+  if (!shouldPatchIndex(0)) return axisConfig;
+  return mergeSingleAxis((axisConfig ?? {}) as AxisOption<T>);
+};
+
+/**
+ * Default styles for axis labels
+ */
+const getDefaultAxisLabelStyle = (): TextCommonOption => ({
```
schilchSICKAG: Those functions are called often during render (e.g. three times per render in axes labels with icons). This means we are reading all the keys (color, fontFamily etc) 3 times from the computed styles. I think we should monitor this closely and think about caching in the long term so it does not always read from the live dom without caching. However, this may be premature optimization, its just something we have to measure over time.

packages/components/src/components/chart/configs/config.ts
kirchsuSICKAG: We have a function deepMerge in the components package, but only for our html linter in *code-quality/html-validate*. Also it handles arrays differently than I do it right now in my merge function. Therefor I would just leave it as it is.

packages/components/src/components/chart/configs/config.ts
```diff
@@ -0,0 +1,139 @@
...
+   */
+  with(modifier: ConfigModifier): this {
+    this.#config = mergeConfigs(this.#config, modifier(this.#config));
+    return this;
+  }
+
+  /**
+   * Applies a named preset by resolving it to a ConfigModifier.
+   */
+  usePreset<K extends SynChartPresetName>(...args: PresetTuple<K>): this {
```
klaarseSICKAG: alternative: addPreset

packages/components/src/components/chart/configs/config.ts
```diff
@@ -0,0 +1,139 @@
...
+ *   .build();
+ * ```
+ *
+ * Modifiers can themselves be composed using `compose`:
+ * ```ts
+ * const myModifier = compose(withAxesSplitLines(), withHiddenYAxisLabels());
+ * chart.config = enhanceConfig(baseConfig).with(myModifier).build();
+ * ```
+ */
+export const enhanceConfig = (base: ECConfig): ChartConfigBuilder => new ChartConfigBuilder(base);
```
klaarseSICKAG: alternative: extendConfig, enrichConfig

packages/components/src/components/chart/configs/axes/utilities.ts
```diff
@@ -0,0 +1,426 @@
...
+  alignTicks: true,
+  axisLabel: {
+    color: getRealStyleValue('--syn-typography-color-text-quiet'),
+    fontFamily: getRealStyleValue('--syn-font-sans'),
+    fontSize: getRealStyleValue('--syn-font-size-x-small'),
+    fontWeight: getRealStyleValue('--syn-font-weight-normal'),
+  },
+  axisLine: {
+    lineStyle: {
+      color: getRealStyleValue('--syn-grid-lines-color'),
```
klaarseSICKAG: should have a different style than splitLine

klaarseSICKAG (dismissed):

packages/components/src/components/chart/configs/axes/types.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * `axisIndex` can be provided as a single index or as a list of indices, to define which axis or axes should be updated.
+ * If no index is set, the calling function can decide which axis or axes should be patched by default.
+ */
+export type AxisPatchOptions = {
+  /**
+   * Index or indices of the axis or axes to be patched. If not provided, the calling function can decide which axis or axes should be patched by default.
+   */
+  axisIndex?: AxisIndexSelection;
+};
```
kirchsuSICKAG, line 28: what do you mean with amount? 

Currently you can either give in:
- a single number `axisIndex: 1`, which would mean the preset is used for axis with index 1
- an array `axisIndex: [0, 2]`, which would mean the preset is used for axes with index 0 and 2

packages/components/src/components/chart/configs/config.ts
kirchsuSICKAG: because of the naming I did another round with the ai. Here are all current proposals we have for the naming. What do you (@klaarseSICKAG, @schilchSICKAG ) all think is best (we could also mix the proposals):

Current state:
```js
enhanceConfig(baseConfig)
  .with(customModifier)
  .usePreset('axes.split-lines')
  .build()
```

Proposal Sebastian:
```js
extendConfig(baseConfig)
  .with(customModifier)
  .addPreset('axes.split-lines')
  .build()
```

Proposal AI 1:

```js
configure(baseConfig)
  .with(customModifier)
  .apply('axes.split-lines')
  .build()
```

Proposal AI 2:
```js
createConfig(baseConfig)
  .with(customModifier)
  .applyPreset('axes.split-lines')
  .build()
```

kirchsuSICKAG: I think I like proposal AI 1 the most. It is short but can be read and understood easily in the flow from my point of view.

schilchSICKAG: > I think I like proposal AI 1 the most. It is short but can be read and understood easily in the flow from my point of view.

I like that one too.

reichst1SICKAG (approved):
After the clarifying comments in Chromatic it all looks good for me.

packages/components/src/components/chart/chart.component.ts
```diff
@@ -70,10 +80,19 @@ export default class SynChart extends SynergyElement {
...
+   *     yAxis: { type: 'value' },
+   *     series: [{ type: 'line', data: [150, 230, 224] }],
+   *   });
+   *   handle.axesShowSplitLines();
+   * };
    * ```
    */
   @property({ attribute: false })
-  config: ECConfig = {};
+  config: ChartConfigType = {};
```
schilchSICKAG, line 106: I like this approach much ❤️. Only thing I notice is that we now have a config and resolvedConfig. There is potential there that they might go out of bounds. What about adding getters and setters for the config property instead and moving the logic of `resolveConfigInput` into them? With this, there would be no need for a resolvedConfig at all. What do you think?

packages/components/src/components/chart/chart.component.ts
```diff
@@ -1,5 +1,5 @@
 import {
```
schilchSICKAG, line 1: We need to update metadata/config/overrides/component:syn-chart with the new values when this branch is merged to make sure it points to the correct links. Also, add figmaComponentId if possible.

schilchSICKAG: We need to wire up intents in metadata/src/intentPolicy for later use. However, I am fine leaving it out for now as we don´t have the full api with sub-components currently. Maybe part of https://github.com/synergy-design-system/synergy-design-system/issues/1312?

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -0,0 +1,357 @@
...
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'axes-split-lines-hidden'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-chart id="chart-lines-hidden"></syn-chart>
+    <script type="module">
+      const charts = document.querySelectorAll('#chart-lines-hidden');
```
schilchSICKAG, line 60: If you exchange this with `document.querySelector('#chart-lines-hidden');`, you can omit the `forEach` below.

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -0,0 +1,358 @@
...
+  parameters: {
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'axes-split-lines-visible'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-chart id="chart-lines-visible"></syn-chart>
+    <script type="module">
```
schilchSICKAG, line 86: See comment on first item: ID selectors will always return the first item as far as I know, so its save to use querySelector instead querySelectorAll.

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -0,0 +1,358 @@
...
+          type: 'category',
+        },
+        yAxis: {
+          name: 'Values',
+          type: 'value',
+        },
+      };
+
+      charts.forEach(chart => {
+        chart.config = handle => {
```
schilchSICKAG: At least for one of the demos we might also show chaining:

```javascript
handle
  .baseConfig(baseConfig)
  .axesShowSplitLines();
```

packages/components/src/components/chart/configs/axes/utilities.ts
kirchsuSICKAG: done

packages/components/src/components/chart/configs/config.ts
kirchsuSICKAG: Was changed after stakeholder feedback

packages/components/src/components/chart/configs/config.ts
kirchsuSICKAG: Was change dafter stakeholder feedback

schilchSICKAG (approved):
This looks pretty nice and stable. We can still make small adjustments as we are still in beta phase, so lets go

PR #1294, opened 2026-06-01
fix: 🐛 syn-dialog font looks blurry when the prevent close animation is run

### LINKED ISSUE #1285: fix: 🐛 syn-dialog - Stacking multiple dialogs leads to blurred text

- [x] I can provide a simple, reproducible example of the issue (e.g. via code-sandbox or sample code, ...). => https://codepen.io/andelawork/pen/raWLpXB

# Version

The bug occurred with the following version of the component library: 3.13.4

# Component

The issue arises when using the following component(s): syn-dialog

# Environment

The issue occurred with the following operating system(s): Windows 11, Windows 10 probably
The issue occurred in the following browser(s): Edge, propably all Chrome based browsers

# Description

When using stacked instances of `<syn-dialog>`, the font of underlying dialogs gets blurred after closing the top dialog. This happens on Windows 11 (and possibly Windows 10, too).

See attached files.

Image
Image
Image

# Acceptance criteria

- [ ] Replicate the pen in Storybook
- [ ] Check in Windows 11 / Windows 10
- [ ] Provide fix

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes an issue introduced with [Synergy@3.11.1](https://synergy-design-system.github.io/?path=/docs/packages-components-changelog--docs#3111) that could lead to text looking blurry in `<syn-dialog>` in Chromium based browsers.

As we are not able to remove the `will-change` that is causing the issue and there is no way to have the same animation as before, I opted to adjust the default animation for closing to get rid of this error.

### 🎫 Issues

Closes #1285

## 👩‍💻 Reviewer Notes

See comment above. We have to change the animation to still indicate its forbidden to leave, but still show user feedback. I also checked other components that have a similar annimation (e.g. syn-drawer), but the error does not happen there. They are not having the same problem (because the crucial part, rendering on the gpu, is missing).

## 📑 Test Plan

1. Have a look at the [current story](https://synergy-design-system.github.io/?path=/story/components-syn-dialog--preventing-the-dialog-from-closing). Notice when clicking outside the dialog a small animation will appear and it will prevent the dialog from being removed.
2. Do the same in Chromatic (same story). You will notice that a new animation is running and the text is not blurred anymore.

## ✅ DoD


packages/metadata/data/layers/full/component/component:syn-dialog/components/dialog.component.ts
```diff
@@ -352,8 +352,14 @@ setDefaultAnimation('dialog.hide', {
...
 
 setDefaultAnimation('dialog.denyClose', {
-  keyframes: [{ scale: 1 }, { scale: 1.02 }, { scale: 1 }],
-  options: { duration: 250 },
+  keyframes: [
+    { translate: 'calc(var(--syn-spacing-small) * -1)' },
+    { translate: 'var(--syn-spacing-small)' },
+    { translate: 'calc(var(--syn-spacing-small) * -1)' },
+    { translate: 'var(--syn-spacing-small)' },
+    { translate: '0' },
```
kirchsuSICKAG: For me this looks a bit chaotic / wild. What do you think about this:

    { translate: '0' },
    { translate: 'calc(var(--syn-spacing-small) * -1)' },
    { translate: 'var(--syn-spacing-small)' },
    { translate: '0' },

kirchsuSICKAG (changes requested):

packages/metadata/data/layers/full/component/component:syn-dialog/components/dialog.component.ts
schilchSICKAG: Yeah, we have to check with design which kind of animation would match best. I just left it there so we can discuss.

reichst1SICKAG: @schilchSICKAG @kirchsuSICKAG I have tried something out on codepen: https://codepen.io/reichst1SICKAG/pen/YPqQbgg

schilchSICKAG: I will update our version with this animation. We can still tune the used variables when the real animation is in place on Chromatic.

schilchSICKAG: Updated, please have another look :)

kirchsuSICKAG: The animation itself is fine for me. I would rather go with Stefans tokens, as the used tokens here make the animation a bit too strong for me.

kirchsuSICKAG (dismissed):

packages/metadata/data/layers/full/component/component:syn-dialog/components/dialog.component.ts
schilchSICKAG: Just changed them. Please have a look again @reichst1SICKAG, @kirchsuSICKAG

reichst1SICKAG (approved):

PR #1302, opened 2026-06-10
fix: 🐛 Components should use correct headline styles (#1290)

### LINKED ISSUE #1290: fix: 🐛 headline styles are bold

# Version

The bug occurred with the following version of the component library: 2.1.0

# Component

The issue arises when using the following component(s): headline styles and componetns like syn-dialog and syn-drawer using them

# Environment

The issue occurred with the following operating system(s):
The issue occurred in the following browser(s):

# Description

New SICK Branding defines all headlines from large to 4x-large (h1-h5) with a regular font-weight. Only medium (h6) is semibold.

Variables and Styles in Figma are correct.

Image
Image

# Acceptance criteria

schilchSICKAG:
# Pull Request

## 📖 Description

This PR fixes multiple typographic issues for headers in the following components:

- `<syn-card>`
- `<syn-dialog>`
- `<syn-drawer>`
- `<syn-header>`

It also includes:

- adjustments for templates to use Synergy headline styles classes
- Adjustments on the global Storybook configuration so it uses our font reliably

### 🎫 Issues

Closes #1290

## 👩‍💻 Reviewer Notes

~~Nothing~~ Something fancy in there:

Code:

- I opted to remove the migration layer for `<syn-card>` and `<syn-drawer>` as both css styles needed changes anyhow. Other than this merge nothing big changed.

Storybook:

Does not override STyles in Stories anymore - Each story is now truly standalone™. This means I had to adjust some stories to make it work.

## 📑 Test Plan

Follow the regular procedure.

## ✅ DoD


reichst1SICKAG (approved):

packages/docs/package.json
```diff
@@ -61,14 +61,15 @@
...
+    "setup-fonts": "cp node_modules/@synergy-design-system/fonts/dist/sick-intl-inline.css public",
+    "build": "pnpm setup-fonts && storybook build --output-dir dist --stats-json",
     "build-storybook": "pnpm build",
     "lint:css": "stylelint \"src/**/*.{css}\" \"src/**/*.styles.ts\"",
     "lint:js": "eslint src",
     "lint": "pnpm run /^lint:.*/",
     "format.eslint": "pnpm lint --fix",
     "chromatic": "chromatic ${CHROMATIC_PROJECT_TOKEN} --build-script-name=build",
-    "start": "storybook dev -p 6006",
+    "start": "pnpm setup-fonts && storybook dev -p 6006",
```
kirchsuSICKAG: Instead of doing the font stuff here, wouldn't it be better to do it in the *".storybook/main.ts" file like the other static files like this: 

```js
    {
      from: '../../fonts/dist/sick-intl-inline.css',
      to: '/',
    },
```

Image

packages/docs/src/helpers/rules-helper.tsx
```diff
@@ -231,7 +231,7 @@ export const RulesHelper: React.FC<RulesHelperProps> = ({ children, forItem }) =
   return (
     <>
       <syn-card
-        className="rules-helper"
+        className="rules-helper sbdocs"
```
kirchsuSICKAG, line 235: Not sure if it is because of this but the overview page looks a bit strange. There are several h2 headlines, which look different.
In the image left is the new overview page and right is the right overview page.

Image

packages/docs/src/docs.css
```diff
@@ -1,113 +1,64 @@
...
+
+    > :is(hr) {
+      font: var(--syn-heading-medium);
+      margin: var(--syn-spacing-medium) 0;
+    }
+
+    > :is(p, ul, li) {
+      font: var(--syn-body-medium-regular) !important;
+      font-family: unset;
+      margin: 0 0 var(--syn-spacing-small) 0;
```
kirchsuSICKAG: Our Doc Pages with much text and headlines now look a bit cramped. E.g. *Principles and vision*, *Accessibility*, *Contributing*, *Limitations/Components*, ...

As example see image with left new and right old *Principles and vision* page: 

Image

kirchsuSICKAG (changes requested):

packages/docs/package.json
schilchSICKAG: You are right, so much easier. Fixed

packages/docs/src/helpers/rules-helper.tsx
schilchSICKAG: Its kind of tricky to get the styling right because there are multiple overrides in Storybook directly that I cannot adjust (basically I would have to `!important` everything). Think I got this one covered (and the sizing of `<h3>` now, too).

packages/docs/src/docs.css
schilchSICKAG: I think the problems stems from the fact that we did not have any real Synergy aligned styles on this page at all. Everything was done using inline styles. It does not help that we don´t have any hard rules when it comes to typography. That said, if you have a look at the A11y section it looks pretty good. The reason the Goals page is kind of strange is it is a heavy mix of headlines with one sentence paragraphs beneath them. Maybe we should redesign this page completely?

kirchsuSICKAG: Not sure if we really need to redesign it. For me it would be fine just to add back a bit of margin to the top of the headlines and subheadlines. Even in the accessibility section I think it is a bit too cramped. But maybe this is just my personal feeling.
Bildschirmfoto 2026-06-22 um 12 45 50

kirchsuSICKAG (approved):

packages/docs/src/docs.css
schilchSICKAG: I added more spacing now, seems to fit better. Not quite right I think, maybe we should just wait for the whole typograhic stuff we still need to implement (syn-prose would help a lot here!)

PR #1314, opened 2026-06-29
feat: ✨ Create new components syn-fieldset and syn-checkbox-group

### LINKED ISSUE #794: feat: ✨ Add basic form styling to styles package

- [x] I would be willing to provide a PR for this component.

# Story

As a developer using the Synergy Design System, I would like to have a baseline of form formatting styles so that all my forms are styled according to Synergies specs. This partly exists as template at https://synergy-design-system.github.io/?path=/docs/templates-forms--docs, but is not available as a component style.

# Acceptance criteria

- [ ] Styles for forms as they are currently defined in the template are available as utility classes in `@synergy-design-system/styles`.
- [ ] Current form stories are migrated to the new utility classes
- [ ] Framework demos are migrated to the new form styles

### LINKED ISSUE #1282: feat: ✨ checkbox group
# Story

As a user of the Synergy Design System, I would like to have a checkbox-group so that I have a consistent spacing between my checkboxes.

# Acceptance criteria

- [ ] All checkboxes have the same margin
- [ ] Checkboxes can be grouped horizontally and vertically
- [ ] Group label
- [ ] Helptext
- [ ] Size
- [ ] Disabled state
- [ ] Invalid state
- [ ] Readonly state
- [ ] form associated
- [ ] framework wrapper

schilchSICKAG:
# Pull Request

## 📖 Description

This PR adds the new component `<syn-fieldset>` with the following features:

- Supports `legend` and `description` via attributes and named slots.
- Adds configurable `layout` (`one-column`/`two-columns`) and `item-spacing` (`normal`/`dense`).
- Syncs disabled state to nested form controls.

### 🎫 Issues

Closes #794
Closes #1282

## 👩‍💻 Reviewer Notes

This is not yet 100% aligned with design - we will do this later in this week. Next to the component, the following adjustments where made:

- All framework demos use the new wrapper, custom code was removed
- Form template uses the new component, custom code was removed
- E2E tests where adjusted to only load the tab that is currently active. This was done to prevent errors in various parts of the applications (e.g. Charts errors to the console because it was rendered in a hidden frame). Speeds up rendering significantly as page load is not that heavy anymore. However, this required custom framework solutions (use of state in demo applications).
- Custom wiring into the Metadata Intent System. The system now has a new intent to group form fields that it will happily provide. It also includes a warning if users do not provide a label property or slot.

## Changes we have to talk about:

- The original order of entries in the contact form could not be preserved. We used custom `css grid` to display a fixed amount of options. This no longer works with a shared approach. I still think this is a minor trade off, but ymmv.
- The stories in Storybook where split between one- and two column layout for easier management.
- In dev, we need a couple of more properties for easier management (e.g. `item-spacing`).
- I did build the system with a currently private `--items-per-row` parameter that is hard coded to `2`. Users could override this if wanted, but I did not document it. I think this could be beneficial later on, e.g. to have the possibility to use three or even four column displays. Open to suggestions, if we want to hide it completely we could move it to the rendered `fieldset` instead.

## 📑 Test Plan

Test the new stories.

## ✅ DoD


packages/metadata/data/layers/full/component/component:syn-fieldset/components/fieldset.styles.ts
```diff
@@ -0,0 +1,85 @@
...
+    width: 100%;
+  }
+
+  .description {
+    font-size: var(--syn-font-size-medium);
+    margin: 0 0 var(--syn-font-size-large);
+  }
+
+  .fields {
+    align-items: end; /* Todo: Test if this is really what we want! */
```
kirchsuSICKAG: From my point of view I would rather go for "align-items: start". Because if people mix inputs / selects with textarea or radiogroup it looks a bit strange having it at the end. But maybe this is a personal taste. Maybe this should be clarified with design.

## align-items: start
Image

## align-items: end

Image

packages/metadata/data/layers/full/component/component:syn-fieldset/components/fieldset.component.ts
```diff
@@ -0,0 +1,220 @@
...
+  @property({ reflect: true, type: Boolean }) disabled = false;
+
+  /**
+    * The layout of the fieldset. This determines how the fields are displayed.
+    * Defaults to `one-column`.
+    *
+    * - `one-column`: All fields are displayed in a single column.
+    * - `two-columns`: Fields are displayed in two columns. Will automatically fall back to one-column if the fieldset is too narrow to display two columns.
+   */
+  @property({ reflect: true }) layout: 'one-column' | 'two-columns' = 'one-column';
```
kirchsuSICKAG, line 99: Not sure about the naming. Maybe better something like `stacked` and `column`. If we make `--items-per-row` available, the naming suits better.

packages/components/src/components/radio-group/radio-group.component.ts
```diff
@@ -88,6 +88,14 @@ export default class SynRadioGroup extends SynergyElement implements SynergyForm
   /** The radio group's size. This size will be applied to all child radios and radio buttons. */
   @property({ reflect: true }) size: 'small' | 'medium' | 'large' = 'medium';
 
+  /**
+   * Defines how options are laid out inside the radio group.
+   *
+   * - `stacked`: options are displayed in a single column (default)
+   * - `inline`: options are displayed in a wrapping row layout
+   */
+  @property({ attribute: 'option-layout', reflect: true }) optionLayout: 'stacked' | 'inline' = 'stacked';
```
kirchsuSICKAG: Not sure if we really want a new property for the radio-group, that only takes actions, when it is inside a syn-fieldset (or an element with "container-type") and otherwise does not have any effect for the component. Also it would change the appearance of the radio-button and does no longer use syn-button-group around it:

Image

Also where does this requirement for the inline layout come from? do we really need it?

packages/components/src/components/fieldset/fieldset.component.ts
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * @csspart base - The component's base wrapper.
+ * @csspart legend - The component's legend element.
+ * @csspart description - The component's description element.
+ * @csspart field-container - The container for the fieldset's fields.
+ *
+ * @cssproperty --item-gap-normal - The gap between the fields in the fieldset when using normal layout. Defaults to `--syn-spacing-large`
+ * @cssproperty --item-gap-dense - The gap between the fields in the fieldset when using dense layout. Defaults to `--syn-spacing-x-small`
+ */
+export default class SynFieldset extends SynergyElement {
```
kirchsuSICKAG, line 37: I am a bit unsure. Do we really want to create an own component for it? I see that there would be some advantages but otherwise isn't this something what we have the styles package for?

packages/components/src/components/fieldset/fieldset.component.ts
```diff
@@ -0,0 +1,220 @@
...
+        }
+      }
+    }
+  }
+
+  private getDirectChildRadioGroups() {
+    return Array.from(this.children).filter(
+      el => el.tagName.toLowerCase() === 'syn-radio-group',
+    ) as HTMLElement[];
+  }
```
kirchsuSICKAG: This would not work if the syn-radio-group would be wrapped with something (e.g. syn-validate or syn-tooltip, ..)

kirchsuSICKAG (changes requested):
I am not sure if we really want to create an own web component or just do a style

schilchSICKAG:
> I am not sure if we really want to create an own web component or just do a style

We prototyped it with a style and finally decided against it. There are just too many unknowns and not much we could do, so we chose to use a full fledged component.

packages/metadata/data/layers/full/component/component:syn-fieldset/components/fieldset.styles.ts
schilchSICKAG: Yeah, much better. Will change it to this.

packages/metadata/data/layers/full/component/component:syn-fieldset/components/fieldset.component.ts
schilchSICKAG: I already discussed this with Eileen briefly. We may want to create other layouts in the future (e.g. specially styled three column layouts etc). Thats why we chose this approach: We can guard from misuse and allow new stuff when it becomes available.

packages/components/src/components/radio-group/radio-group.component.ts
schilchSICKAG: Radio button was not ready when creating this. The branch needs to be updated. Basically, you cannot do what you would regulary do with other components as syn-radio/radio-button always need wrappers. Because of those wrappers, normal flexbox does not apply anymore (not even when using `display: contents` on the root). Because of this, we have to trigger the layout somehow. I would be fine with trying a css variable based approach, but won´t hold my breath if it does not work.

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: There are many customizations and not all of them can be done with plain css. We would have to provide multiple styles (e.g. one for fields intended for use in radios, one with regular fields etc. Plus we cannot enforce users to use label and descriptions.

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: Thats valid, will adjust

packages/components/src/components/fieldset/fieldset.component.ts
```diff
@@ -0,0 +1,220 @@
...
+   * The legend for the fieldset. This is displayed as the title of the fieldset.
+   * If not provided, the fieldset will not have a legend.
+   */
+  @property({ reflect: true, type: String }) legend = '';
+
+  /**
+   * Whether the fieldset is disabled.
+   * When true, all form controls inside the fieldset are disabled
+   */
+  @property({ reflect: true, type: Boolean }) disabled = false;
```
kirchsuSICKAG, line 84: Another thought on this, if having a web component, we should add a size property which propagates to all element, or what do you mean?
Also do we have a design for small, medium and large ( e.g. for the title, description,...) or are they always the same?

kirchsuSICKAG (changes requested):

packages/metadata/data/layers/full/component/component:syn-fieldset/components/fieldset.component.ts
kirchsuSICKAG: sounds valid

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: Thats a good idea. Currently, they are only using one size, but I can see that this would be beneficial. Will discuss with Eileen.

.changeset/794-syn-fieldset.md
```diff
@@ -0,0 +1,20 @@
+---
+"@synergy-design-system/angular": minor
+"@synergy-design-system/components": minor
+"@synergy-design-system/metadata": minor
+"@synergy-design-system/react": minor
+"@synergy-design-system/tokens": minor
+"@synergy-design-system/vue": minor
+---
+
+feat: ✨ syn-fieldset (#794)
```
kirchsuSICKAG: Add also syn-checkbox-group into the changelog title


```suggestion
feat: ✨ syn-fieldset (#794) and syn-checkbox-group (#1282)
```

.changeset/794-syn-fieldset.md
```diff
@@ -0,0 +1,20 @@
...
+
+Adds a new `<syn-fieldset>` component for grouping related form controls with semantic markup and accessibility support.
+
+- Supports `legend` and `description` via attributes and named slots.
+- Adds configurable `layout` (`one-column`/`two-columns`) and `item-spacing` (`normal`/`dense`).
```
kirchsuSICKAG: There is not longer an `item-spacing` property


```suggestion
- Adds configurable `layout` (`one-column`/`two-columns`)
```

packages/_private/angular-demo/src/AllComponentParts/Combobox.ts
```diff
@@ -196,13 +196,11 @@ import {
...
 export class Combobox implements OnInit {
-  levels: SelectItem[] = [];
+  levels: SelectItem[] = mockData('selectItems');
   delimiterItems = mockData('selectItemsWithSpace');
   numericItems = mockData('selectItemsMixedValue');
   cb632Value: string = '';
-  // @ts-ignore
-  asyncValue: string;
-
+  asyncValue: string = mockData('valueWithSpace');
```
kirchsuSICKAG: Why changing this behavior?
With this changes the async test cases for the combobox are no longer tested, as it becomes initially the same value.

packages/_private/angular-demo/src/AllComponentParts/Select.ts
```diff
@@ -155,11 +155,10 @@ import {
   `
 })
 export class Select implements OnInit {
-  levels: SelectItem[] = [];
+  levels: SelectItem[] = mockData('selectItems');
   numericItems = mockData('selectItemsMixedValue');
   delimiterItems = mockData('selectItemsWithSpace');
-  // @ts-ignore
-  asyncValue: string;
+  asyncValue: string = mockData('valueWithSpace');
```
kirchsuSICKAG: Why changing this behavior?
With this changes the async test cases for the select are no longer tested, as it becomes initially the same value.

packages/components/src/components/radio-group/radio-group.styles.ts
```diff
@@ -100,4 +100,13 @@ export default css`
...
     --syn-color-neutral-0: var(--syn-readonly-color-text);
   }
+
+  /**
+   * #794: Allow radio-groups to be displayed in a row when using syn-radio-button.
+   * Will not apply to syn-radio-button, which is always displayed in a column.
+   */
+  .form-control--is-horizontal:not(.form-control--has-button-group) .form-control-input {
+    flex-flow: row wrap;
+    gap: var(--syn-spacing-medium);
```
kirchsuSICKAG: In design the row-gap and column-gap is not the same (see image).

Image


```suggestion
    gap: var(--syn-spacing-x-small) var(--syn-spacing-medium);
```

packages/components/src/components/checkbox-group/checkbox-group.component.ts
```diff
@@ -0,0 +1,218 @@
...
+ * @slot label - The checkbox group's label. Required for proper accessibility. Alternatively, you can use the `label`
+ *  attribute.
+ * @slot help-text - Text that describes how to use the checkbox group. Alternatively, you can use the `help-text` attribute.
+ *
+ * @csspart form-control - The form control that wraps the label, input, and help text.
+ * @csspart form-control-label - The label's wrapper.
+ * @csspart form-control-input - The input's wrapper.
+ * @csspart form-control-help-text - The help text's wrapper.
+ * @csspart button-group - The button group that wraps radio buttons.
+ * @csspart button-group__base - The button group's `base` part.
```
kirchsuSICKAG: This parts are not available here. Probably a copy+paste

```suggestion
```

packages/components/src/components/checkbox-group/checkbox-group.component.ts
```diff
@@ -0,0 +1,218 @@
...
+          'form-control--has-help-text': hasHelpText,
+          'form-control--has-label': hasLabel,
+          'form-control--is-horizontal': this.layout === 'horizontal',
+          'form-control--large': this.size === 'large',
+          'form-control--medium': this.size === 'medium',
+          'form-control--small': this.size === 'small',
+        })}
+        aria-labelledby="label"
+        aria-describedby="help-text"
+        aria-errormessage="error-message"
```
kirchsuSICKAG: We do not have an *error-message* here in this component which could be referenced. Probably a copy+paste.


```suggestion
```

packages/components/src/components/checkbox-group/checkbox-group.component.ts
```diff
@@ -0,0 +1,218 @@
...
+  @property({ attribute: 'help-text' }) helpText = '';
+
+  /** The checkbox group's size. This size will be applied to all child checkboxes. */
+  @property({ reflect: true }) size: 'small' | 'medium' | 'large' = 'medium';
+
+  /**
+   * The layout of the checkbox group. This determines how the checkboxes are displayed.
+   *
+   * - `horizontal`: Radios are displayed in a row.
+   * - `vertical`: Radios are displayed in a column.
```
kirchsuSICKAG: Copy+paste


```suggestion
   * - `horizontal`: Checkboxes are displayed in a row.
   * - `vertical`: Checkboxes are displayed in a column.
```

packages/components/src/components/checkbox-group/checkbox-group.component.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    availableCheckboxes[index].focus();
+
+    event.preventDefault();
+  }
+
+  private handleLabelClick() {
+    this.focus();
+  }
```
kirchsuSICKAG, line 162: The focus via handleLabelClick only works (visible) when no checkbox is checked. If one is checked there is no focus ring displayed. See gif:

Image

packages/components/src/components/checkbox-group/checkbox-group.component.ts
```diff
@@ -0,0 +1,218 @@
...
+   * - `vertical`: Radios are displayed in a column.
+   */
+  @property({ reflect: true }) layout: 'horizontal' | 'vertical' = 'vertical';
+
+  /**
+   * By default, form controls are associated with the nearest containing `<form>` element. This attribute allows you
+   * to place the form control outside of a form and associate it with the form that has this `id`. The form must be in
+   * the same document or shadow root for this to work.
+   */
+  @property({ reflect: true }) form = '';
```
kirchsuSICKAG, line 67: If we provide this convenience property *form*, I would have expected, that we also provide the function *getForm*, so the user also does not have to get into the single checkbox to get the corresponding form. If this is somehow easy achievable

packages/components/src/components/checkbox-group/checkbox-group.styles.ts
```diff
@@ -0,0 +1,37 @@
...
+  .form-control--has-help-text.form-control--radio-group .form-control__help-text {
+    margin-top: var(--syn-spacing-x-small);
+  }
+
+  /**
+   * Allow checkbox-groups to be displayed in a row when using syn-radio-button.
+   */
+  .form-control--is-horizontal .form-control-input {
+    flex-flow: row wrap;
+    gap: var(--syn-spacing-medium);
```
kirchsuSICKAG: In design the row-gap and column-gap is not the same (see image).

Image


```suggestion
    gap: var(--syn-spacing-x-small) var(--syn-spacing-medium);
```

packages/components/src/components/checkbox-group/checkbox-group.styles.ts
```diff
@@ -0,0 +1,37 @@
...
+    flex-direction: column;
+    gap: var(--syn-spacing-x-small);
+  }
+
+  .form-control--has-help-text.form-control--radio-group .form-control__help-text {
+    margin-top: var(--syn-spacing-x-small);
+  }
+
+  /**
+   * Allow checkbox-groups to be displayed in a row when using syn-radio-button.
```
kirchsuSICKAG: Copy+paste


```suggestion
   * Allow checkbox-groups to be displayed in a row.
```

packages/metadata/data/layers/examples/component/component:syn-checkbox-group.md
```diff
@@ -0,0 +1,171 @@
...
+<syn-checkbox-group label="This is a label">
+  <syn-checkbox name="checkbox-1" value="checkbox-1">Option</syn-checkbox>
+  <syn-checkbox name="checkbox-2" value="checkbox-2">Option</syn-checkbox>
+  <syn-checkbox name="checkbox-3" value="checkbox-3">Option</syn-checkbox>
+</syn-checkbox-group>
+```
+
+---
+
+## Layout Vertical
```
reichst1SICKAG: Rename to "Vertical Layout" -> reads easier

packages/metadata/data/layers/examples/component/component:syn-checkbox-group.md
```diff
@@ -0,0 +1,171 @@
...
+
+  <syn-checkbox name="checkbox-2" value="checkbox-2">Option</syn-checkbox>
+
+  <syn-checkbox name="checkbox-3" value="checkbox-3">Option</syn-checkbox>
+</syn-checkbox-group>
+```
+
+---
+
+## Layout Horizontal
```
reichst1SICKAG: Rename to "Horizontal Layout" -> reads easier

packages/components/src/components/checkbox-group/checkbox-group.styles.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  .form-control-input {
+    display: flex;
+    flex-direction: column;
+    gap: var(--syn-spacing-x-small);
+  }
+
+  .form-control--has-help-text.form-control--radio-group .form-control__help-text {
+    margin-top: var(--syn-spacing-x-small);
```
kirchsuSICKAG, line 32: This styling would never be applied, as it looks for class *form-control--radio-group* instead of *form-control--checkbox-group*. 
But if I see it correctly we do not even need it, as the margin-top is set by *form-control.styles.js* file.

See image:

Image

packages/components/src/components/checkbox-group/checkbox-group.test.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable */
+import '../../../dist/synergy.js';
+import { expect, fixture } from '@open-wc/testing';
+import type SynCheckbox from './checkbox-group.js';
+
+describe('<syn-checkbox-group>', () => {
+
```
kirchsuSICKAG: There are currently no tests 😉

packages/components/src/components/fieldset/fieldset.styles.ts
```diff
@@ -0,0 +1,58 @@
...
+     * The gap between the fields in the fieldset.
+     */
+    --item-gap: var(--syn-spacing-large);
+
+    display: block;
+  }
+
+  .fieldset {
+    border: none;
+    margin: 0;
```
kirchsuSICKAG, line 24: We need to set an explicit color. For storybook it currently is correct, as we set the color there correctly, but if the wrapping element has another color set (e.g. red) the text will become red too. See image:

Image


```suggestion
    border: none;
    color: var(--syn-typography-color-text);
    margin: 0;
```

packages/components/src/components/fieldset/fieldset.styles.ts
```diff
@@ -0,0 +1,58 @@
...
+  .legend {
+    display: block;
+    font: var(--syn-heading-large);
+    margin-bottom: var(--syn-spacing-large);
+    padding: 0;
+    width: 100%;
+  }
+
+  .description {
+    font-size: var(--syn-font-size-medium);
```
kirchsuSICKAG: We also have a problem with the font-weight of the description. If setting another font-weight on the wrapping element, it will use this font-weight, as it is not explicitly set in the component. 
See image:

Image

Maybe we should do some basic font styling in the *fieldset* class, so we do not need to set everything on both classes *legend* and *description*

packages/components/src/components/fieldset/fieldset.styles.ts
```diff
@@ -0,0 +1,58 @@
...
+  .fieldset {
+    border: none;
+    margin: 0;
+    padding: 0;
+  }
+
+  .legend {
+    display: block;
+    font: var(--syn-heading-large);
+    margin-bottom: var(--syn-spacing-large);
```
kirchsuSICKAG, line 31: The margins between the different elements of the fieldset are different in design. 

If there is legend + description + field-container:
- spacing between legend + description is 12px
- spacing between description + field-container is 24px

If there is legend + field-container:
- spacing between legend + field-container is 24px

If there is description + field-container:
- spacing between description + field-container is 24px

packages/components/src/components/fieldset/fieldset.styles.ts
```diff
@@ -0,0 +1,58 @@
...
+    padding: 0;
+    width: 100%;
+  }
+
+  .description {
+    font-size: var(--syn-font-size-medium);
+    margin: 0 0 var(--syn-font-size-large);
```
kirchsuSICKAG: The margins between the different elements of the fieldset are different in design. 

If there is legend + description + field-container:
- spacing between legend + description is 12px
- spacing between description + field-container is 24px

If there is legend + field-container:
- spacing between legend + field-container is 24px

If there is description + field-container:
- spacing between description + field-container is 24px

packages/components/src/components/fieldset/fieldset.component.ts
```diff
@@ -0,0 +1,226 @@
...
```
kirchsuSICKAG, line 84: Currently we are not really consistent with our form elements disabled styling.
This becomes noticeable when you use the fieldset with all kinds of form elements and then set it to “disabled.”
This is the result:

Image

For every form element the *label* is getting a disabled styling, but for the checkbox-group and the radio-group this is not the case.
The problem here is, they do not have a *disabled* property. To be consistent we have two possibilities from my point of view:
1. we could remove the disabled styling for the *label* on every form element
2. we add a *disabled* property for both, syn-radio-group and syn-checkbox-group and give their *label*s also the same disabled style

packages/components/src/components/fieldset/fieldset.component.ts
```diff
@@ -0,0 +1,226 @@
...
+    }
+  }
+
+  protected updated(_changedProperties: PropertyValues<this>) {
+    super.updated(_changedProperties);
+    if (_changedProperties.has('disabled')) {
+      this.syncDisabledState();
+    }
+
+    this.scheduleGroupedControlLayoutSync();
```
kirchsuSICKAG: Do we really want this do trigger everytime any property changes?
We only need this for the *layout* and *disable-auto-group-layout* property

packages/components/src/components/fieldset/helpers.ts
```diff
@@ -0,0 +1,84 @@
...
+      const typedGroup = group as HTMLElement & { layout: string };
+      if (typedGroup.layout !== targetLayout) {
+        typedGroup.layout = targetLayout;
+      }
+      return;
+    }
+
+    if (group.getAttribute('layout') !== targetLayout) {
+      group.setAttribute('layout', targetLayout);
+    }
```
kirchsuSICKAG: Is there a specific reason, why to not just set it via property and remove the setAttribute paret?

packages/docs/stories/components/checkbox-group.stories.ts
```diff
@@ -0,0 +1,181 @@
...
+        </syn-checkbox-group>
+      `)}
+    </div>
+  `,
+};
+
+
+/* eslint-disable sort-keys */
+export const Screenshot: Story = generateScreenshotStory({
+  Default,
```
kirchsuSICKAG, line 279: Currently only the default story is screenshotted

packages/docs/stories/components/checkbox-group.stories.ts
```diff
@@ -0,0 +1,181 @@
...
+    <div style="display: flex; flex-direction: column; gap: var(--syn-spacing-large);">
+      ${['small', 'medium', 'large'].map(size => html`
+        <syn-checkbox-group label="This is a label" help-text="Choose the most appropriate option." size="${size}">
+          ${createCheckboxes(2)}
+        </syn-checkbox-group>
+      `)}
+    </div>
+  `,
+};
+
```
kirchsuSICKAG, line 276: *Disabled* and *readonly* story is missing, but they are available in design

packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -810,6 +810,78 @@
...
+          "type": "text"
+        }
+      },
+      "help-text": {
+        "title": {
+          "value": "Help Text",
+          "type": "text"
+        },
+        "description": {
+          "value": "Add descriptive help text to a switch with the help-text attribute. For help texts that contain HTML, use the help-text slot instead.\n\nThe help-text attribute should not be used to display error messages. To handle validation and error messaging, use syn-validate for proper error management.",
```
kirchsuSICKAG: copy+paste

```suggestion
          "value": "Add descriptive help text to a checkbox-group with the help-text attribute. For help texts that contain HTML, use the help-text slot instead.\n\nThe help-text attribute should not be used to display error messages. To handle validation and error messaging, use syn-validate for proper error management.",
```

packages/docs/stories/templates/form.stories.tsx
```diff
@@ -150,59 +94,63 @@ export const ContactForm = {
...
+          id="topics"
+          layout="two-columns"
+          legend="${getTranslation('contactForm.topicLabel')}"
+          style="--item-gap: var(--syn-spacing-x-small)"
+        >
+          ${(getTranslation('contactForm.topics') as string[]).map((topic, index) => html`
+            <syn-checkbox name="topic[${index}]" value="${topic}">
+              ${topic}
+            </syn-checkbox>
+          `)}
```
kirchsuSICKAG: We should remove the *--item-gap: var(--syn-spacing-x-small)* css overwrite and use the syn-checkbox-group now, as we have it

kirchsuSICKAG (changes requested):

packages/_private/angular-demo/src/demos-template/demos-template.component.ts
```diff
@@ -20,22 +20,30 @@ export class DemosTemplateComponent {
...
+    if (!this.demoNames.has(name)) {
+      return;
+    }
+
     (e.target as HTMLElement).parentElement?.scrollTo(0, 0);
 
     const dialog = document.querySelector('syn-dialog');
     if (dialog) {
       dialog.open = name === 'Dialog';
     }
```
kirchsuSICKAG: Unfortunately, this does no longer work. With the new handling of loading the specific components, the syn-dialog has no longer time to be there, until the dialog tab is clicked.
Therefore the dialog does no longer open for any framework demo application.

We have two possibilities:
1. Delay this functionality
2. Do the opening in the dialog component itself, when we know the component is really there

kirchsuSICKAG (changes requested):

packages/docs/stories/templates/form.stories.tsx
schilchSICKAG: We will take this back to design. Basically, we currently just alter the flex-direction, so its not possible to show it in a grid. We could implement this, but would also have to do the same in syn-radio-group.

packages/tokens/src/figma-tokens/_docs.json
schilchSICKAG: Fixed

packages/docs/stories/components/checkbox-group.stories.ts
schilchSICKAG: Both where added

packages/docs/stories/components/checkbox-group.stories.ts
schilchSICKAG: Fixed in latest

packages/components/src/components/fieldset/helpers.ts
schilchSICKAG: Not anymore. Originally, radio-group had no layout switch, so i needed both variants for type safety. Gone now.

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: Corrected.

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: Right, we should create a new ticket for this and forward this to design.

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: Talked about this with Eileen, we will just use one size for now and add a size parameter later on when needed. She does currently not see the need for this.

packages/metadata/data/layers/full/component/component:syn-fieldset/components/fieldset.styles.ts
schilchSICKAG: Done

packages/components/src/components/radio-group/radio-group.component.ts
schilchSICKAG: Its now `variant="vertical"` and only available on the `<syn-radio-group>`. The code now makes sure that it is not forwarded to the radio-buttons button-group wrapper.

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: Closing as discussed.

packages/components/src/components/fieldset/fieldset.component.ts
schilchSICKAG: This is now exchanged for direct `document.querySelectorAll` call in `applyGroupedControlLayout`

packages/metadata/data/layers/examples/component/component:syn-checkbox-group.md
schilchSICKAG: Done

packages/metadata/data/layers/examples/component/component:syn-checkbox-group.md
schilchSICKAG: Done. Must be done in Design, too

.changeset/794-syn-fieldset.md
schilchSICKAG: Fixed. I used `h4 (####)` so when the changelog is created it has the correct markdown structure.

.changeset/794-syn-fieldset.md
schilchSICKAG: Removed the prop table for good.

packages/components/src/components/radio-group/radio-group.styles.ts
schilchSICKAG: Fixed. Please also note that until #1323 is fixed, the gaps are NOT the same as in Figma.

packages/_private/angular-demo/src/demos-template/demos-template.component.ts
schilchSICKAG: And fortunately it is also no longer needed to work :). As the page is now loading completely, we are able to just use `<syn-dialog open>` without wrapping the whole page. I removed the listener and also made sure to have a correct closing behavior for the button. One less bug :)

packages/_private/angular-demo/src/AllComponentParts/Select.ts
schilchSICKAG: Will have another look into this.

packages/_private/angular-demo/src/AllComponentParts/Combobox.ts
schilchSICKAG: The ngOnInit never worked as intended - we tested a side-effect. The problem is, that ngOnInit was called before the whole stuff was already on the page, but the rendering was only done when the tab became visible. Change detection did not pick it up. I fixed it using AI, but did not have a look into the fix itself. My fault.

Just adjusted it again so it now will call the change detector directly to make sure the data is applied. Seems to work fine and async again!

packages/_private/angular-demo/src/AllComponentParts/Select.ts
schilchSICKAG: The ngOnInit never worked as intended - we tested a side-effect. The problem is, that ngOnInit was called before the whole stuff was already on the page, but the rendering was only done when the tab became visible. Change detection did not pick it up. I fixed it using AI, but did not have a look into the fix itself. My fault.

Just adjusted it again so it now will call the change detector directly to make sure the data is applied. Seems to work fine and async again!

packages/_private/angular-demo/src/AllComponentParts/Combobox.ts
kirchsuSICKAG: Could have also used the lifecycle hook ngAfterViewInit instead of ngOnInit. At this point all elements are visible in the dom.
But if this works I'm fine

packages/components/src/components/fieldset/fieldset.component.ts
kirchsuSICKAG: We created a new issue for this (https://github.com/synergy-design-system/synergy-design-system/issues/1324) and will fix it in a new PR

packages/components/src/components/fieldset/helpers.ts
```diff
@@ -0,0 +1,79 @@
...
+  const selector = [
+    'button',
+    'fieldset',
+    'input',
+    'select',
+    'textarea',
+
+    // Synergy elements
+    'syn-button',
+    'syn-checkbox',
```
kirchsuSICKAG, line 21: We should directly add the syn-checkbox-group here as well. As it also has the *size* property and in the future will get the *disabled* property, so it getts synced correctly from the fieldset

packages/components/src/components/checkbox-group/checkbox-group.component.ts
schilchSICKAG: Fixed

packages/components/src/components/checkbox-group/checkbox-group.component.ts
schilchSICKAG: Fixed

packages/components/src/components/checkbox-group/checkbox-group.component.ts
schilchSICKAG: Fixed

packages/components/src/components/checkbox-group/checkbox-group.component.ts
schilchSICKAG: Fixed. This issue is also appearing in radio-group. Fixed that there, too.

kirchsuSICKAG (changes requested):

packages/components/src/components/checkbox-group/checkbox-group.component.ts
schilchSICKAG: Hmm, not sure about this one. The problem is that the checkbox-group is not form-controlled, so it has no associated form. You could also have weird stuff like:

```html
<syn-checkbox-group>
  <syn-checkfox form="forma" />
  <syn-checkfox form="formb" />
</syn-checkbox-group>
```

Which would we return in this case? The problem is that a radio-group has a single value, but a checkbox-group is a plain wrapper for elements. Don´t know if this works.

packages/components/src/components/checkbox-group/checkbox-group.styles.ts
schilchSICKAG: Fixed

packages/components/src/components/checkbox-group/checkbox-group.styles.ts
schilchSICKAG: Fixed

packages/components/src/components/checkbox-group/checkbox-group.styles.ts
schilchSICKAG: I fixed it and chose to leave the selector in. Else I would have to remove it from the radio-group, too

packages/components/src/components/fieldset/fieldset.styles.ts
schilchSICKAG: Fixed

packages/components/src/components/fieldset/fieldset.styles.ts
schilchSICKAG: I now chose to use `font: var(--syn-body-medium-regular);`, which should make sure it is setup correctly.

packages/components/src/components/fieldset/fieldset.styles.ts
schilchSICKAG: Fixed

packages/components/src/components/fieldset/fieldset.styles.ts
schilchSICKAG: Adjusted to --syn-spacing-large

packages/components/src/components/fieldset/helpers.ts
schilchSICKAG: Fixed

packages/components/src/components/checkbox-group/checkbox-group.component.ts
kirchsuSICKAG: Yeah your right. Then let just ignore this. But maybe we could make an addition to the documentation of the form propertty. Something like:

"This component does not participate on the form control. It only passes through the *form* property to the underlying form element"

What do you think about this?

schilchSICKAG: Yes, will do that next

schilchSICKAG: Done

kirchsuSICKAG (approved):

packages/docs/stories/templates/form.stories.tsx
schilchSICKAG: Fixed, using syn-checkbox-group with a custom container query now

PR #1319, opened 2026-07-07
feat: ✨ Add chart legend

### LINKED ISSUE #1204: feat: ✨ chart legend
# Component

# Story

As a user of the Synergy Design System, I would like to have a legend with each chart so I can relate and focus colors and line-styles to provided data sets.

# Acceptance criteria

- [ ] List of data indicator-items consisting of label and colored symbols
- [ ] Symbol colors match the series colors and shape
- [ ] Each indicator can be used to toggle a data series (if wanted)
- [ ] Toggling can be enabled/disabled (default is enabled)
- [ ] Toggling via API should be reflected in the legend UI
- [ ] Placement can be defined
- [ ] Displaying the legend is optional

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds the synergy default styling for the ECharts legend as well as preset for positioning the legend in a convenient way.

### 🎫 Issues

Closes #1204 

### Decision for the preset
We currently have one preset for the legend, which handles the position of the legend together with some magic like adding grid configs to make the chart not overlap with the chart.
Currently the preset only allows to position ONE legend. Theoretically multiple legends could be added to the chart. Now we have multiple possibilities:

1. for now we only handle the position of one legend with **legendShow** in the future (or now) we add another function called **legendsShow** which can position multiple legends
```js
// Position one legend
legendShow({ position: 'left' });
// Position multiple legends
legendsShow({ position: ['left', 'right'] });
```

2. we enhance the *legendShow* method to give in arrays to it can handle multiple legends directly
```js
legendShow({ position: ['left', 'right'] });
```

### To be discussed with design
Currently the visibility icon is ALWAYS shown in dev (see image for what the visibility icon is).
Bildschirmfoto 2026-07-10 um 11 00 08

The current state in design is, that the visibility icon is an optional property. In the past dashboarding solution, the icon was mandatory. After talking with Sebastian we decided, that the icon should either be always mandatory or if not needed from accessibility perspective, it should be completely removed.

This needs to be discussed and decided with design.

## 📑 Test Plan
Test as usual.

## ✅ DoD


.changeset/1204-chart-legend-tokens.md
```diff
@@ -0,0 +1,6 @@
+---
+"@synergy-design-system/tokens": minor
+---
+
+feat: ✨ Adds the new chart token `--syn-chart-disabled-color` (#1204)
```
schilchSICKAG, line 5: Just a reminder for me: The check reports that the tokens package is missing. I think I should refactor our check so it includes anything, not the main file only.

packages/components/src/components/chart/configs/legend/presets.ts
```diff
@@ -0,0 +1,35 @@
...
+ * for multiple legends or grids. The position only applies to the first legend in the array.
+ *
+ * @param {LegendOptions} options Preset options.
+ * @param {'top' | 'bottom' | 'left' | 'right'} [options.position='top'] Position used to derive default legend placement and grid offsets.
+ * @param {ECConfig['legend']} [options.legendOptions] Custom legend configuration merged onto the position defaults.
+ * @param {ECConfig['grid']} [options.gridOptions] Custom grid configuration merged onto the computed legend-safe offsets.
+ * @see https://echarts.apache.org/en/option.html#legend
+ * @see https://echarts.apache.org/en/option.html#grid
+ */
+export const legendShow = (options: LegendOptions = {}): ConfigModifier => (config) => {
```
schilchSICKAG: We could perhaps get rid of some of the code by using a default legendOptions and merging it:

```typescript
// Current
export const legendShow = (options: LegendOptions = {}): ConfigModifier => (config) => {

// New
export const legendShow = (
  options: LegendOptions = {
    position: 'top',
    legendOptions = {}.
    gridOptions = {},
  }
): ConfigModifier => (config) => {
```

In this case, we could then remove:

```typescript
const position = options.position ?? 'top';
const legendOptions = options.legendOptions ?? {};
const gridOptions = options.gridOptions ?? {};
```

However, I think this is a very minor issue and it is save as is, nothing that is wrong about the used approach here.

packages/components/src/components/chart/configs/legend/utilities.ts
```diff
@@ -0,0 +1,170 @@
+import type { LegendComponentOption } from 'echarts/types/dist/echarts.js';
+import type { ECConfig } from '../../types.js';
+import { measureMaxTextWidth } from '../axes/utilities.js';
+import { getRealStyleValue, getRealValueWithoutUnit } from '../../themes/utilities.js';
+import type { LegendPosition } from './types.js';
+import { colorSvgDataUrl } from '../utilities.js';
+
+const VISIBILITY_ICON = `<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
schilchSICKAG: I wonder if we could draw those icons directly from `packages/components/src/components/icon/sick2025-system-icons.ts`? We will always need this file, as we ship it with anything in the library. It already is the eye svg as far as I can see. It also would auto-update.

packages/components/src/components/chart/configs/legend/utilities.ts
```diff
@@ -0,0 +1,170 @@
...
+</svg>
+`;
+
+const getVisibilityIconDataUrl = (isVisible: boolean): string => {
+  const svg = isVisible ? VISIBILITY_ICON : VISIBILITY_OFF_ICON;
+  const dataUrl = `data:image/svg+xml;base64,${btoa(svg)}`;
+  return colorSvgDataUrl(dataUrl, getRealStyleValue('--syn-typography-color-text-quiet'));
```
schilchSICKAG, line 13: I see we do lots of calls to `getRealStyleValue`. Maybe it would be beneficial to cache the results and add a monitor for when a value has changed or something like that? There are getting more and more direct dom writes, could be performance problems later on down the road.

packages/components/src/components/chart/configs/legend/utilities.ts
```diff
@@ -0,0 +1,170 @@
...
+});
+
+/**
+ * Builds the default legend placement settings for a given position.
+ *
+ * @param {LegendPosition} position Legend position.
+ * @returns Legend config fragment that places the legend for the requested position.
+ */
+export const getLegendConfigForPosition = (position: LegendPosition): NonNullable<ECConfig['legend']> => {
+  const legendByPosition: Record<LegendPosition, NonNullable<ECConfig['legend']>> = {
```
schilchSICKAG: If we add this as a type into types.ts, we could make this a one liner with direct return.

packages/components/src/components/chart/configs/legend/utilities.ts
```diff
@@ -0,0 +1,170 @@
...
+export const getGridForLegendPosition = (position: LegendPosition, legendStyle: NonNullable<ECConfig['legend']>, config: ECConfig): NonNullable<ECConfig['grid']> => {
+  const series = config?.series;
+  if (!series) {
+    return {};
+  }
+
+  // Early return for top and bottom positions, as we don't need to calculate the width for those positions
+  if (position === 'top' || position === 'bottom') {
+    return {
+      [position]: 80,
```
schilchSICKAG: Should we share such magic numbers via a constants.ts?

schilchSICKAG (commented):
Looks pretty good from my pov! Just had another thought about the use of dom based variables: What if we would just export a javascript object with the original values in the tokens package and provide a sync utility? Basically, we would export something like:

```javascript
// New file: dist/charts/js/theme-XXX.js
export const StaticValuesForCharts = {
  SynChartCategorical10: {
    dynamic: 'var(--syn-chart-categorical-10)',
    resolved: '#f00',
  }
};
```

We could then use this map to either work with the dynamic values or the resolved ones. When someone wants to update a token, we could then provide a `syncTokens(instance: SynChart)` (name pending...) command that updates the resolved values for each of the keys and triggers a redraw.

Just a first idea, we might also think about better ones :)

kirchsuSICKAG:
> Looks pretty good from my pov! Just had another thought about the use of dom based variables: What if we would just export a javascript object with the original values in the tokens package and provide a sync utility? Basically, we would export something like:
> 
> ```
> // New file: dist/charts/js/theme-XXX.js
> export const StaticValuesForCharts = {
>   SynChartCategorical10: {
>     dynamic: 'var(--syn-chart-categorical-10)',
>     resolved: '#f00',
>   }
> };
> ```
> 
> We could then use this map to either work with the dynamic values or the resolved ones. When someone wants to update a token, we could then provide a `syncTokens(instance: SynChart)` (name pending...) command that updates the resolved values for each of the keys and triggers a redraw.
> 
> Just a first idea, we might also think about better ones :)

I really like the idea! I think that would also help for the dark / light theme switching.
We would need this for all tokens (also components). In the next PR I will have a look at it. If this is not that big to do, I will add it there

packages/components/src/components/chart/configs/legend/utilities.ts
kirchsuSICKAG: But would it be correct (semantically), if we add the whole *legendByPosition* object to the types.ts? I would have expected it only include types and no preset objects, that is what the utilities.ts is for, aren't it?

packages/components/src/components/chart/configs/legend/presets.ts
kirchsuSICKAG: As we changed the complete handling of input arguments and what is possible (either string or object), we can no longer do such defaults in the input arguments themself

packages/components/src/components/chart/configs/legend/utilities.ts
kirchsuSICKAG: Thats a nice idea. Changed it to use the system icons

packages/components/src/components/chart/configs/legend/utilities.ts
kirchsuSICKAG: I added a simple caching mechanism. We can probably get rid of it as soon as we also export resolved values for our tokens.
I did not add a monitor mechanism, as we do not yet have implemented the theme switching mechanism and therefore the resolved values do not yet change.

packages/components/src/components/chart/configs/legend/utilities.ts
kirchsuSICKAG: Done

packages/components/src/components/chart/configs/axes/utilities.ts
```diff
@@ -232,7 +214,7 @@ const getYAxisLabelEffectiveWidth = (
     ? measureMaxTextWidth(texts, getFontShorthand(labelsStyle))
     : 0;
 
-  return measured > 0 ? measured : 30;
+  return measured > 0 ? measured : AXIS.LABEL_FALLBACK_WIDTH;
```
schilchSICKAG, line 217: I like that much better. Now we still have magic numbers, but at least they are scoped somewhere else together.

packages/components/src/components/chart/configs/legend/utilities.ts
schilchSICKAG: Thats fine for me, the branch is big enough for now I think :)

packages/components/src/components/chart/themes/utilities.ts
```diff
@@ -1,13 +1,20 @@
...
+ * Results are cached per token
  *
  * @param token CSS custom property name, e.g. --syn-color-primary.
  * @returns Trimmed computed value or an empty string when the token is not defined.
  */
 export const getRealStyleValue = (token: string): string => {
-  const computedStyles = getComputedStyle(document.body);
-  const value = computedStyles.getPropertyValue(token).trim();
-  return value || '';
+  if (styleTokenCache.has(token)) {
```
schilchSICKAG, line 11: I like that better. However, there can be something like:

- with sick2025-light: Chart calls getRealStyleValue('color1'); // Maps to red
- user switches to sick2025-dark: Chart gets new data and used getRealStyleValue('color2'); // Maps to pink, should be yellow

What do you think about prepolulating the cache, by basically:

```typescript
const styleTokenCache = new Map<string, string>();

export const warmupStyleTokenCache = (
  target: Element = document.body,
  options: { includeOnly?: string[]; prefix?: string } = {},
): number => {
  if (typeof window === 'undefined' || !target) {
    return 0;
  }

  const { includeOnly, prefix = '--' } = options;
  const computed = getComputedStyle(target);
  let count = 0;

  // Optional selective warmup (recommended for chart-only tokens)
  if (includeOnly && includeOnly.length > 0) {
    includeOnly.forEach((token) => {
      const value = computed.getPropertyValue(token).trim();
      styleTokenCache.set(token, value || '');
      count += 1;
    });
    return count;
  }

  // Full warmup of all custom properties available on the computed style
  for (let i = 0; i < computed.length; i += 1) {
    const name = computed.item(i);

    if (!name.startsWith(prefix)) {
      continue;
    }

    const value = computed.getPropertyValue(name).trim();
    styleTokenCache.set(name, value || '');
    count += 1;
  }

  return count;
};
```

We could also add prefixes and stuff to make sure not everything is taken into account. Then we could invoke the cache when the first chart is rendered. Cool thing about this is we could also refresh the cache later on when the body classes change from 25 to 18 or dark to light. We could also do this in a requestAnimationFrame so it does not block rendering I think, but I am not sure.

packages/docs/stories/charts/legend.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'legend-top'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-chart id="chart-show-legend-top"></syn-chart>
+    <script type="module">
+      const charts = document.querySelectorAll('#chart-show-legend-top');
```
schilchSICKAG, line 58: If you exchange this for document.querySelector, you wont get an array and dont need to use foreach down below. As when you have an id set, it should be the first element returned as ids need to be unique on a page.

packages/docs/stories/charts/legend.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+          { data: [320, 332, 301, 334, 390, 330, 320], type: 'line', name: 'Series C' },
+        ],
+        xAxis: {
+          data: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'],
+          type: 'category', name: 'Days',
+        },
+        yAxis: { type: 'value', name: 'Values' },
+      };
+
+      charts.forEach(chart => {
```
schilchSICKAG, line 108: Same here, use querySelector instead.

packages/docs/stories/charts/legend.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'legend-right'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-chart id="chart-show-legend-right"></syn-chart>
+    <script type="module">
+      const charts = document.querySelectorAll('#chart-show-legend-right');
```
schilchSICKAG, line 128: And here too

packages/docs/stories/charts/legend.stories.tsx
```diff
@@ -0,0 +1,194 @@
...
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'legend-bottom'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-chart id="chart-show-legend-bottom"></syn-chart>
+    <script type="module">
+      const charts = document.querySelectorAll('#chart-show-legend-bottom');
```
schilchSICKAG, line 163: For completness, here too :)

schilchSICKAG (changes requested):
Looks pretty solid! I am not sure about my comment with the token prewarming, but would think it adds some stability and performance. However, I would also be fine if we tackle this later on when theming gets made.

packages/components/src/components/chart/configs/legend/utilities.ts
schilchSICKAG: Fine for me, too :)

packages/docs/stories/charts/legend.stories.tsx
kirchsuSICKAG: This is a needed fix as otherwise it does not work if you go to a single story and than switch back to the full docs page with all stories. If we do not do this via array, the configs are not set correctly and therefore the charts are no longer shown in the docs page 😕

packages/docs/stories/charts/legend.stories.tsx
kirchsuSICKAG: see other comment

packages/docs/stories/charts/legend.stories.tsx
kirchsuSICKAG: see other comment

packages/docs/stories/charts/legend.stories.tsx
kirchsuSICKAG: see other comment

packages/docs/stories/charts/legend.stories.tsx
schilchSICKAG: Understood.

packages/components/src/components/chart/themes/utilities.ts
kirchsuSICKAG: Added this functionality

packages/metadata/data/layers/full/component/component:syn-chart/components/chart.component.ts
```diff
@@ -155,18 +156,40 @@ export default class SynChart extends SynergyElement {
 
   connectedCallback() {
     super.connectedCallback();
-    registerTheme('default', synergyLightTheme);
+    /**
+     * We need to fill up the token cache with the current theme values, so that the chart can be rendered correctly on first render.
+     * This is needed before setting the theme, so the cache is used there.
+     */
+    warmupStyleTokenCache();
```
schilchSICKAG: Beware, when using multiple instances on the page they all will warmup. Maybe we should add a static thingy in here instead, something like:

```typescript
class SynChart {
  static #sucessfullyWarmedUp = 0;

  connectedCallback() {
    if (!self.#successfullyWarmedUp) {
      self.#successfullyWarmedUp = warmupStyleTokenCache();
    }
```

With this, we also know how many variables are in there. `0` would be false in boolean checks, so it would also work. However, this leaves the edge case of having no variables assigned, so maybe a boolean fits better?

schilchSICKAG (changes requested):
One small performance addition (maybe the warmup should be only done once). Otherwise it works fine.

You could also scope the cache if you wanted, but I think this is too much complexity for too little gain (e.g. using a factory with one cache per component so they are independent)

packages/metadata/data/layers/full/component/component:syn-chart/components/chart.component.ts
kirchsuSICKAG: I added this, but for now I decided to do it as boolean

schilchSICKAG (approved):

PR #1329, opened 2026-07-14
chore: 🔧 Create a docker container with auto publish for the mcp server [skip-chromatic]

### LINKED ISSUE #1321: chore: 🔧 Create a docker container with auto publish for the mcp server
# Description

We need to add an automatic docker automatic publish system for our mcp server to make it available for cloud deployments. 

# Rationale

We want to have an always up 2 date version of our mcp server that can be used in k8s and single cloud deploys.

schilchSICKAG:
# Pull Request

## 📖 Description

This release adds more abilities to the Synergy MCP-Server:

- Adds support to create docker containers
- Allows middleware compression for mcp resources
- Adds the ability to log to stdout for easier usage in k8s environments
- Adds ability to silence log outputs

### 🎫 Issues

Closes #1321

## 👩‍💻 Reviewer Notes

I have added instructions, as well as local setup scripts that you may use to create containers locally. Pushing them to the registry is currently out of scope - this will be handeled internally.

The new parameters cli where added because we need to get the log data from the process stdout instead from the file system - however both is possible by using `docker run --rm -p 9119:9119 -v "PATH_TO_YOUR_CONFIG/synergy-mcp.example.json:/config/synergy-mcp.json:ro" -v "$(pwd)/logs:/opt/synergy/logs" synergy-design-system/mcp:local` which enables local filesystem based logging, as well as stdout (if configured).

I also took about 5 minutes time to create a colored version of the Synergy shell ascii art (basically making it blue). This is currently used in the metadata package (generate-skills) and the mcp server, so I chose to add it to the metadata package. Should do no harm at all, but if you are not fine with it, will remove it again :).

## ✅ DoD

### DIFF
--- a/packages/mcp/docker/start.sh
+++ b/packages/mcp/docker/start.sh
@@ -1,0 +1,3 @@
+#/bin/sh
+mkdir -p ./logs && chmod 777 logs
+docker run --rm -p 9119:9119 -v "$(pwd)/../synergy-mcp.example.json:/config/synergy-mcp.json:ro" -v "$(pwd)/../logs:/opt/synergy/logs" synergy-design-system/mcp:local
--- a/packages/mcp/docker/start.sh
+++ b/packages/mcp/docker/start.sh
@@ -1,3 +1,3 @@
-#/bin/sh
+#!/bin/sh
 mkdir -p ./logs && chmod 777 logs
 docker run --rm -p 9119:9119 -v "$(pwd)/../synergy-mcp.example.json:/config/synergy-mcp.json:ro" -v "$(pwd)/../logs:/opt/synergy/logs" synergy-design-system/mcp:local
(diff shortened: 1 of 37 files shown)


packages/mcp/docker/start.sh
```diff
@@ -0,0 +1,3 @@
+#/bin/sh
```
kirchsuSICKAG: The explanation mark is missing in this bash script


```suggestion
#!/bin/sh
```

kirchsuSICKAG (changes requested):

packages/mcp/docker/start.sh
schilchSICKAG: Fixed

kirchsuSICKAG (approved):

PR #1331, opened 2026-07-17
feat: ✨ new variants for syn-alert and syn-badge

### LINKED ISSUE #1199: feat: ✨ new variants for syn-alert
# Component

I suggest to add the feature to the following component: syn-alert

# Story

As a user of the Synergy Design System, I would like to have a critical variant since yellow is not visible to good

# Acceptance criteria

- [ ] Decide on the name for the new variant
- [ ] Think about providing other namur colors (e.g. instead of a new variant, you could do something like `<syn-alert use-namur (name pending!) variant="danger"></syn-alert>`. This would mean we get two times the variants and colors would always be namur save.

schilchSICKAG:
# Pull Request

## 📖 Description

This release adds the new `critical` and `error` variants to alerts and badges. It also deprecates the `danger` variant. Please use `error` in the future.

### 🎫 Issues

Closes #1199

## 👩‍💻 Reviewer Notes

- There is no possibility to flag a variant in an enum as deprecated.
- I did intentionally not add a console.log as our current logging system is not good for this.
- Because of that, I wrote a first Synergy V4 migration document, highlighting the changes instead.
- ☠️ Token fetching went badly wrong - the reason is that charting branch uses stale references (e.g. coupled to main). Because of that, you will have to make sure to omit fetching charts when working in local branches that have updates to colors and such. Because of that, I added a small utility that allows us to run fetching without charts using `build:nocharts`. With this addition, I was able to add the missing tokens.
- I am aware that I pulled in the new spacing token for `syn-radio-button-spacing-small`. There will be another branch that has the same token in it and should land before this branch is released. If it is released first, I can omit fetching the tokens for the other branch, so I think its fine :).

## 📑 Test Plan

Have a look at the changed stories basically.

## ✅ DoD


packages/tokens/scripts/figma/transformer-variables-generic.js
```diff
@@ -382,14 +384,51 @@ const resolveValue = (variablesData, variable, modeId) => {
...
+          'Unable to resolve COLOR alias to an RGBA object.',
+          `variable: ${name}`,
+          `variableId: ${variableId}`,
+          `modeId: ${modeId}`,
+          `aliasTargetId: ${aliasTargetId}`,
+          `resolvedValueType: ${typeof modeValue}`,
+          `resolvedValue: ${JSON.stringify(modeValue)}`,
+        ].join(' '));
+      }
+    }
```
kirchsuSICKAG: I would remove this as we have the exact same check in the following if / else check


```suggestion
```

packages/components/src/components/alert/alert.component.ts
```diff
@@ -44,40 +46,41 @@ import { enableDefaultSettings } from '../../utilities/defaultSettings/decorator
...
    * use the `show()` and `hide()` methods and this attribute will reflect the alert's open state.
    */
-  @property({ type: Boolean, reflect: true }) open = false;
+  @property({ reflect: true, type: Boolean }) open = false;
 
   /** Enables a close button that allows the user to dismiss the alert. */
-  @property({ type: Boolean, reflect: true }) closable = false;
+  @property({ reflect: true, type: Boolean }) closable = false;
 
   /** The alert's theme variant. */
```
kirchsuSICKAG: I am thinking about adding the deprecation information in the property description. What do you think about that?

Maybe something like this:

```suggestion
  /**
  * The alert's theme variant.
  *
  * > *The `danger` variant is deprecated and will be removed in a future major release.
  * Use `error` instead.*
  */
```

packages/components/src/components/badge/badge.component.ts
```diff
@@ -22,24 +22,26 @@ import { enableDefaultSettings } from '../../utilities/defaultSettings/decorator
  */
 @enableDefaultSettings('SynBadge')
 export default class SynBadge extends SynergyElement {
-  
   private readonly localize = new LocalizeController(this);
-  static styles: CSSResultGroup = [componentStyles, styles, customStyles];
+
+  static styles: CSSResultGroup = [componentStyles, styles];
 
   /** The badge's theme variant. */
```
kirchsuSICKAG: I am thinking about adding the deprecation information into the property description. What do you think about that?

Maybe something like this:
```suggestion
  /**
  * The badge's theme variant.
  *
  * > *The `danger` variant is deprecated and will be removed in a future major release.
  * Use `error` instead.*
  */
```

packages/components/src/components/alert/alert.styles.ts
kirchsuSICKAG, line 110: I would add the alert-...-color-text token here, as we also ship it with our tokens package and we might need it in the future. What do you mean?


```suggestion
    --variant-color-indicator: var(--syn-alert-success-color-indicator);
    --variant-color-text: var(--syn-alert-success-color-text);
```

kirchsuSICKAG, line 118: I would add the alert-...-color-text token here, as we also ship it with our tokens package and we might need it in the future. What do you mean?


```suggestion
    --variant-color-indicator: var(--syn-alert-neutral-color-indicator);
    --variant-color-text: var(--syn-alert-neutral-color-text);
```

kirchsuSICKAG, line 144: I would add the alert-...-color-text token here, as we also ship it with our tokens package and we might need it in the future. What do you mean?

```suggestion
    --variant-color-indicator: var(--syn-alert-error-color-indicator);
    --variant-color-text: var(--syn-alert-error-color-text);
```

kirchsuSICKAG: This is not the correct token for the close button. It is getting visible for the dark mode 2025, as *warning* and *critical* are getting a black text compared to the other variants with white text. Our old alerts (before these changes here) every text and icon was white in dark mode, but now it is mixed (white or black), depending on the variant.

## Current dev appearance
The close buttons of *warning* and *critical* are white, when everything else is black
Image

## Current design appearance

Image

packages/docs/stories/components/alert.stories.ts
```diff
@@ -98,16 +98,28 @@ export const Variants: Story = {
...
+
+      <syn-alert variant="critical" open>
         <syn-icon slot="icon" name="warning"></syn-icon>
         <strong>Your session has ended</strong><br />
         Please login again to continue.
       </syn-alert>
 
-      <syn-alert variant="danger" open>
+      <syn-alert variant="error" open>
         <syn-icon slot="icon" name="status-error" library="system"></syn-icon>
```
kirchsuSICKAG, line 107: Something just occurred to me: I don't think we should use system icons as code examples, since they're just for internal use. What do you think?


```suggestion
        <syn-icon slot="icon" name="report"></syn-icon>
```

packages/docs/stories/components/alert.stories.ts
```diff
@@ -98,16 +98,28 @@ export const Variants: Story = {
...
         <strong>Your account has been deleted</strong><br />
         We're very sorry to see you go!
       </syn-alert>
+
+      <syn-alert variant="danger" open>
+        <syn-icon slot="icon" name="status-error" library="system"></syn-icon>
```
kirchsuSICKAG, line 113: Something just occurred to me: I don't think we should use system icons as code examples, since they're just for internal use. What do you think?


```suggestion
        <syn-icon slot="icon" name="report"></syn-icon>
```

packages/docs/stories/components/alert.stories.ts
```diff
@@ -98,16 +98,28 @@ export const Variants: Story = {
       </syn-alert>
 
       <syn-alert variant="warning" open>
+        <syn-icon slot="icon" name="emergency_home"></syn-icon>
```
kirchsuSICKAG: Unfortunately there is no such icon like *emergency_home* for the 2018 icons.. 🙈
but from my side we can just ignore it

Image

kirchsuSICKAG (changes requested):

packages/docs/stories/components/alert.stories.ts
reichst1SICKAG: I would keep the system icons here. They are the default icons and should only be changed if really neccessary.

packages/components/src/components/alert/alert.component.ts
schilchSICKAG: Done.

packages/components/src/components/alert/alert.styles.ts
schilchSICKAG: Good catch, added for all items.

schilchSICKAG: Fixed.

packages/docs/stories/components/alert.stories.ts
schilchSICKAG: Fixed, using the system icons for those stories now

packages/docs/stories/components/alert.stories.ts
schilchSICKAG: We do from now on.

packages/docs/stories/components/alert.stories.ts
schilchSICKAG: Done for all alerts

packages/components/src/components/badge/badge.component.ts
schilchSICKAG: Done

packages/tokens/scripts/figma/transformer-variables-generic.js
schilchSICKAG: Leftover, removed.

packages/docs/stories/components/alert.stories.ts
```diff
@@ -346,14 +365,27 @@ const Toast = {
...
+        <strong>Your session has ended</strong><br />
+        Please login again to continue.
+      </syn-alert>
+
+      <syn-alert variant="critical" open closable>
         <syn-icon slot="icon" name="warning"></syn-icon>
-        Your session has ended
+        <strong>Your session has ended</strong><br />
+        Please login again to continue.
       </syn-alert>
```
kirchsuSICKAG, line 372: Even if this story is internal only and is just used for our chromatic screenshots, I would also exchange the icons with the system-icons.

kirchsuSICKAG (changes requested):

packages/docs/stories/components/alert.stories.ts
schilchSICKAG: Fixed.

kirchsuSICKAG (approved):

PR #1338, opened 2026-07-24
docs: 📚 Create template for syn-dialog

### LINKED ISSUE #550: docs: 📚  Create template for syn-dialog
# Description

Dialog templates should show how common dialog usecases are implemented and how edge cases are handled.

# Acceptance criteria

- [ ] confirm dialog (incl. keyboard navigation)
- [ ] accordion as content (dialog should prevented from closing if accordion is clicked on)
- [ ] small screen estates
- [ ] evaluate medici examples

# Ideas

- [ ] Forms in Dialogs (e.g. Application Setting Forms)

schilchSICKAG:
# Pull Request

## 📖 Description

This PR includes three new templates for dialogs that show:

- Usage and Design of Confirmation Dialogs
- Usage of Forms in Dialogs
- A responsive example, based on the Cookie Consent Banner, as initially planned in the Design Ticket.

### 🎫 Issues

Closes #550 

## 👩‍💻 Reviewer Notes

- Check wording for the dialogs and especially actions.
- We state that we should not use desctructive wording, but the Delete Dialog says "Delete" on it. I think when using non destructive wording, "Confirm deletion" would be a better fit, but I am no UX writer. However, we should find something that fits with our wording strategy.
- The content of the cookie stuff is taken strictly from sick.com as of time of writing (2026/07/24). This may change.
- Have a look at the small UX improvements I made and also test in mobile mode (for example, I am not sure where to place the "Legal Statement" in the Footer in very small resolutions).

## 📑 Test Plan

- Regular procedure. Nothing internally changed, so I would opt to just test the new Stories.

## ✅ DoD


packages/docs/stories/templates/dialog.stories.tsx
```diff
@@ -0,0 +1,1204 @@
...
+  };
+
+  const onRequestClose = (e: SynRequestCloseEvent) => {
+    if (e.detail.source === 'overlay') {
+      e.preventDefault();
+      return;
+    }
+
+    closeDialog();
+  };
```
kirchsuSICKAG, line 267: In terms of reusability and readability, this approach with extracting all the functions and functionalities is great. The problem, however, is that because of that it doesn't appear in the code preview. That means if someone copies the template from our storybook docs, they won't get the entire handling logic (like not closing the dialog, when the overlay is clicked, etc)

packages/docs/stories/templates/dialog.stories.tsx
```diff
@@ -0,0 +1,1204 @@
...
+      return acc;
+    }, {} as Record<string, typeof cookies>);
+
+    return html`
+      <syn-header label="Complex Cookie Banner"></syn-header>
+      <main class="cookie-banner-template">
+        <syn-button
+          @click=${async () => {
+            await document.querySelector<SynDialog>('.cookie-banner-dialog')?.show();
+          }}
```
kirchsuSICKAG, line 791: Also such functionalities are not shown in the code preview code

Bildschirmfoto 2026-07-29 um 15 11 00

packages/docs/stories/templates/dialog.stories.tsx
```diff
@@ -0,0 +1,1204 @@
...
+            <a
+              class="syn-link syn-link--medium"
+              href="https://www.sick.com/de/de/sick-datenschutzerklaerung/w/dataprotection"
+              target="_blank"
+              rel="noopener noreferrer"
+            >
+              Privacy Policy
+            </a>
+            <syn-button variant="text">Save settings</syn-button>
+            <syn-button variant="filled">Allow and continue</syn-button>
```
kirchsuSICKAG, line 957: Is it intentional, that these buttons do not have any effect? If yes, the action on the button outside of the dialog does not really need a trigger action (opening the dialog)

packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -4959,6 +4959,48 @@
...
+          "type": "text"
+        }
+      },
+      "responsive": {
+        "title": {
+          "value": "Responsive Dialogs",
+          "type": "text"
+        },
+        "description": {
+          "value": "Example demonstrating responsive dialog behavior. The dialog is designed to adapt to different screen sizes and orientations, ensuring a consistent and user-friendly experience across devices. On smaller screens, the dialog may take up the full width of the viewport, while on larger screens, it may be centered with a fixed width. This responsiveness helps maintain usability and accessibility for all users.",
```
kirchsuSICKAG, line 5000: I don't think the wording is quite right here. Shouldn't the word `"width"` be exchanged with `"height"`? As in the css the height stretches to the whole view height and not width?

```html
     @container cookie-banner-template (max-width: 500px) {
        .cookie-banner-dialog {
          &::part(panel) {
            max-height: 95dvh;
          }
    ....
```


```suggestion
          "value": "Example demonstrating responsive dialog behavior. The dialog is designed to adapt to different screen sizes and orientations, ensuring a consistent and user-friendly experience across devices. On smaller screens, the dialog may take up the full height of the viewport, while on larger screens, it may be centered with a fixed height. This responsiveness helps maintain usability and accessibility for all users.",
```

kirchsuSICKAG (changes requested):

packages/docs/stories/templates/dialog.stories.tsx
schilchSICKAG: Yes, I did not want to add functionality there that would also be part of the story data. I could also add a small alert message after saving and closing the dialog.

packages/docs/stories/templates/dialog.stories.tsx
schilchSICKAG: I did this intentionally to not include any code that has nothing to do with the dialog itself.

packages/docs/stories/templates/dialog.stories.tsx
schilchSICKAG: I initially did this on purpose so only the markup for the dialog is the main content. Reasoning was that people might not need the information about how data flows into the application. However, I think you are right that at least the closing is probably interesting. I would not do that for the complete state handling (e.g. updating fields, setting state etc) because it will water down what I think is the real usage example: Layouting a dialog window for different purposes.

packages/tokens/src/figma-tokens/_docs.json
schilchSICKAG: What I wanted to say is that it adopts when there is not enough width (as it is based on the maximal amount of available width). That said, maybe this is not needed at all and we just may remove the media query alltogether as it does not give real value.

packages/docs/stories/templates/dialog.stories.tsx
schilchSICKAG: Done

reichst1SICKAG:
@schilchSICKAG can you adjust the breakpoints to 480px and 640px for mobile and tablet

kirchsuSICKAG (approved):

PR #1341, opened 2026-07-27
feat: ✨ Vue: Create type only wrappers for Vue

### LINKED ISSUE #1340: feat: ✨ Vue: Create type only wrappers for Vue

# Story

As a developer using the Synergy Design System and Vue.js, I would like to just use the native Synergy web-components so that it is correctly type checked but does not create bigger bundles.

# Acceptance criteria

- [x] Type wrappers poc successfully created
- [x] Adjust demo application and successfully run e2e tests
- [x] Create documentation and known issue documents
- [x] Add types to automated build
- [x] Add types to exports

schilchSICKAG:
# Pull Request

## 📖 Description

This release introduces typescript only wrappers in the `@synergy-design-system/vue` package.
You now get type safety for Synergy Components without using the Vue wrappers components for less bundle size.

### 🎫 Issues

Closes #1340

## ✅ DoD


packages/vue/README.md
```diff
@@ -55,7 +61,54 @@ import App from "./App.vue";
...
+  <syn-button type="submit">Submit me</syn-button>
+  <syn-input name="my-input" required @syn-change="onChange" />
+</template>
+```
+
+This enables automatic type checks and autocompletion for Synergy intrinsic elements in Vue templates.
+
+#### 4.1. Migrating from Vue wrappers to native components
+
+1. Upgrade `@synergy-design-system/vue` to the latest version.
```
kirchsuSICKAG, line 102: I would add the point here, that the dependency should be shifted to dev dependency, as it is no longer needed as dependency except the two way data binding module is used

packages/_private/vue-demo/src/AllComponentParts/DemoAccordion.vue
kirchsuSICKAG, line 1: I am thinking about if we really want to remove all the vue wrapper from the demo application. If we do this, we do no longer test them with our e2e test.
On the other hand we are also no longer testing the react wrapper, since we have the react types.

I think it would make more sense to test for both (react and vue) the framework wrapper instead of the native components, as they are tested with our vanilla demo anyways. What do you think?

kirchsuSICKAG (changes requested):

packages/vue/README.md
schilchSICKAG: Its a bit more complicated in this case :(. Basically, you cannot have automated two way databinding with the type only wrappers. I created a synergy module for this that you have to instanciate. So basically, if you don´t need two way databinding you are good to go with a dev dependency, but if you need the two way flow, you have to add the following to your page:

```typescript
import { createApp } from 'vue';
import { SynModelPlugin } from '@synergy-design-system/vue';

// Your app lives here!
import App from './App.vue';

createApp(App)
  // Mandatory, or two way binding will fail!
  .use(SynModelPlugin)
  .mount('#root');

File: packages/_private/vue-demo/src/AllComponentParts/DemoAccordion.vue
Comment (schilchSICKAG) [reply packages/_private/vue-demo/src/AllComponentParts/DemoAccordion.vue:1]: I think we mainly test the correct rendering, so it should not matter that much. The complex examples (Forms) do both (DemoForm uses native, ValidationForm uses the wrappers). I think its fine that way - otherwise we would have to adjust the react demos, too.

File: packages/vue/README.md [hunk 2]
```diff
@@ -237,7 +290,86 @@ You may use it in one of the following ways:
...
+  const name = ref("");
+  const formValues = ref({
+    email: "",
+    agreed: false,
+    rating: 5,
+  });
+</script>
+
+<template>
+  <syn-input v-model="name" label="Name" required />
```
kirchsuSICKAG: The v-model for syn-input only works partly. If using syn-input type="number", the v-model only works, if typing something into the input. using keys or using the spin buttons, will not result in a change of the v-model value. See gif:

Image


Example code: 

```vue
<script setup lang="ts">
import { ref } from 'vue';

const name = ref("");
const name2 = ref("");
</script>

<template>
  <div style=" display:flex">
  <label>Html input type="number"</label>
  <input v-model="name2" type="number" />
  <span>{{ name2 }}</span>
 <syn-input
    type="number"
    label="Synergy input type=number"
    v-model="name"
  />
  <span>{{ name }}</span>
  </div>
</template>
```

kirchsuSICKAG (approved):
Looks really good and feels also good in usage

Issue #1220, opened 2026-03-05
feat: ✨ theme switching for syn-chart


klaarseSICKAG:
# Component

I suggest to add the feature to the following component: syn-chart

# Story

As a user of the Synergy Design System, I would like to switch themes for applications charts are used in.

# Consideration

Since charts are not based on css the solution is more complex than for css based components.

# Acceptance criteria

- [ ] All DoD items successfully completed
- [ ] light & dark theme
- [ ] switch chart theme if global theme settings are switched
- [ ] registerTheme() / setTheme()
- [ ] getComputedStyle()

kirchsuSICKAG:
Currently a minimalistic approach was added with a dedicated MutationObserver per syn-chart instance that observes the body element for theme class changes.

We should take a closer look at this implementation:

- whether we should keep one observer per chart instance or use a single global observer shared across all syn-chart instances,
- if a shared observer would provide measurable performance benefits compared to the additional complexity of maintaining a registry of active chart instances,
- whether there is a better and more performant theming mechanism that does not rely on DOM mutation observation,
- if parent elements should be checked for theme classes as this is how our other components currently resolve theming,
- how to avoid expensive DOM tree traversal or parent-chain checks whenever theme-related changes occur.

PR #1343, opened 2026-07-28
feat: ✨ Add line series to syn-chart

### LINKED ISSUE #1218: feat: ✨ line chart
# Component

# Story

As a user of the Synergy Design System, I would like to show lines in my chart so that I can visualize related data over time.

# Acceptance criteria

- [ ] Multiple lines with various colors, line-styles and line-thickness
- [ ] Positive and negative values
- [ ] Data coloring based on palettes
- [ ] Data coloring based on custom colors
- [ ] Focus lines by hovering
- [ ] Tooltip shows actual values at a specific time / category along a vertical intersection line
- [ ] Zoom
- [ ] Series can be hidden via legend

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds the synergy default styling for the ECharts line series, tooltip and datazoom. Additionally it adds new presets for creation of line series and enabling of tooltips.

### 🎫 Issues
Closes #1218 

## 📑 Test Plan
As usual

## ✅ DoD


packages/metadata/external-data/storybook/_docs.json
```diff
@@ -706,6 +766,126 @@
...
           "type": "text"
         }
+      },
+      "line-chart-default": {
+        "title": {
+          "value": "Line Chart",
+          "type": "text"
+        },
+        "description": {
+          "value": "Line charts offer a clear way to visualize trends, changes and directional movements over time or across categories. By connecting individual data points, they highlight patterns such as growth, decline or stability at a glance, making them ideal for dashboards, KPI overviews and lightweight data summaries. \n\nThe component provides flexible styling options that adapt to different visual needs, from straight linear segments to smooth curved lines. Combined with optional symbols, line charts can be tailored to emphasize exactly the level of detail or simplicity required in your layout.",
```
klaarseSICKAG, line 776: this is the exact text from line 255

packages/_private/angular-demo/src/AllComponentParts/LineChart.ts
```diff
@@ -0,0 +1,23 @@
...
+  standalone: true,
+  imports: [
+    SynChartComponent,
+  ],
+  template: `
+    <syn-chart [config]="lineChartConfig"></syn-chart>
+  `,
+})
+export class LineChart {
+  lineChartConfig: ChartConfigCallback = (handle) => {
```
schilchSICKAG: I think this is the same version across all frameworks, right? You could also add this to the shared utilties in this case (e.g. the complete ChartConfigCallback as a function) so you can share it in all demos.

packages/_private/react-demo/src/AllComponentParts/LineChart.tsx
```diff
@@ -0,0 +1,13 @@
+import '@synergy-design-system/components/components/chart/chart.js';
+import type { ChartConfigCallback } from '@synergy-design-system/components/components/chart/types.js';
+import { mockData } from '@synergy-design-system/demo-utilities';
+
+const lineChartConfig: ChartConfigCallback = (handle) => {
```
schilchSICKAG: See angular comment, just add to demo utilities, e.g. something like:

```typescript
// _private/demo-utilities/src/charts.ts
export const LineChartDefaultCallback: ChartConfigCallback....

// index.ts
export * as Charts from './charts.js';

// Usage in the components:
import { Charts } from '@synergy-design-system/demo-utilities';

export const LineChart = () => (
  <syn-chart config={Charts.LineChartDefaultCallback} />
);
```

Makes code easier to follow and updates easier later on. Also, when we establish this pattern it will be easier in the future too. What do you think?

packages/components/src/components/chart/themes/theme.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param {Themes} mode Theme mode used to resolve the design tokens.
+ */
+// Synergy ECharts Theme
+export const getSynergyTheme = (mode: Themes = 'light') => ({
+  // Remove the starting animation for chart series
+  animationDuration: 0,
+  categoryAxis: getDefaultAxisStyles(mode),
+  // Default color palette for charts, is categorical by default but can be overridden by setting the palette property on the chart component or by directly setting config.color
+  color: getCategoricalColors(mode),
+  darkMode: 'auto',
```
schilchSICKAG: What is this for?

packages/components/src/components/chart/themes/theme.ts
```diff
@@ -0,0 +1,51 @@
...
+  // Default line series styles
+  line: {
+    // Currently there is a bug in ECharts with symbol type 'none', where in the legend a filled circle is shown. This is a known bug in ECharts (https://github.com/apache/echarts/issues/20958)
+    symbol: 'none',
+    symbolSize: getRealValueWithoutUnit('SynSpacingXSmall', mode),
+  },
+  logAxis: getDefaultAxisStyles(mode),
+  // Global font style
+  textStyle: {
+    color: getRealStyleValue('SynTypographyColorText', mode),
```
schilchSICKAG: Just an idea, but if we adjust the named import we could get rid of a couple of bytes here. We are using `getRealStyleValue` quite often, so it should be noticable and as it is internal only I don´t think we need to fully write this:

```typescript
import {
  type Themes,
  getRealStyleValue as style,
  getRealValueWithoutUnit,
} from './utilities.js';
```

```typescript
// later on
color: style('SynTypographyColorText', mode),
```

Maybe this is premature optimization, but I think file size will grow large with this feature later on.

packages/components/src/components/chart/chart.component.ts
```diff
@@ -54,7 +56,7 @@ export default class SynChart extends SynergyElement {
 
   private resolvedConfig: ECConfig = {};
 
-  static #tokenWarmUpDone = false;
+  private themeObserver: MutationObserver;
```
schilchSICKAG, line 60: Premature optimization again, but I think it would be beneficial if we only had one observer for all instances. In this case, it would be more costly however:

- The observer function would need to know the elements that are available on the page ( read: each syn-chart)
- syn-chart would have to add itself and remove itself in the given callbacks

This would make it possible to reuse one observer for the whole page. However, this stands and falls with where the value for dark mode is coming from. If its only on the body, then such a solution will work fine. If its not, than we may have to use a different solution anyway (as we should not check the whole dom tree each time it changes to detect if one of the parents received a new dark/light switch).

Fine for now, but we should think about how to do this in a performant way.

packages/components/src/components/chart/chart.palettes.ts
```diff
@@ -22,89 +24,89 @@ export type ChartPalette =
...
-/** Maps each palette name to the ordered list of CSS custom property names to resolve. */
-export const PALETTE_TOKENS: Record<ChartPalette, string[]> = {
+/** Maps each palette name to the ordered list of token names to resolve. */
+export const PALETTE_TOKENS: Record<ChartPalette, ResolvedTokensName[]> = {
   categorical: [
-    '--syn-chart-categorical-01', '--syn-chart-categorical-02', '--syn-chart-categorical-03',
-    '--syn-chart-categorical-04', '--syn-chart-categorical-05', '--syn-chart-categorical-06',
-    '--syn-chart-categorical-07', '--syn-chart-categorical-08', '--syn-chart-categorical-09',
-    '--syn-chart-categorical-10', '--syn-chart-categorical-11', '--syn-chart-categorical-12',
+    'SynChartCategorical01', 'SynChartCategorical02', 'SynChartCategorical03',
```
schilchSICKAG, line 30: Nice and clean, not dom reading anymore. Like this.

packages/docs/stories/charts/line-chart.stories.tsx
```diff
@@ -0,0 +1,466 @@
...
+
+      const baseConfig = {
+        xAxis: {
+          data: ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'],
+          type: 'category', name: 'Days',
+        },
+        yAxis: { type: 'value', name: 'Values' },
+      };
+
+      const lineData = [[820, 932, 901, 934, 1290, 1330, 1320], [620, 732, 701, 734, 1090, 1130, 1120]];
```
schilchSICKAG: Do you use lineData here?

packages/docs/stories/charts/line-chart.stories.tsx
```diff
@@ -0,0 +1,466 @@
...
```
schilchSICKAG: Do you use this anywhere?

packages/docs/stories/charts/line-chart.stories.tsx
```diff
@@ -0,0 +1,466 @@
...
```
schilchSICKAG: Not used, remove

packages/docs/stories/charts/line-chart.stories.tsx
```diff
@@ -0,0 +1,466 @@
...
+    },
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'line-series-tooltip'),
+      },
+    },
+  },
+  play: async ({ canvasElement }) => {
+    // Only run this code when running in Chromatic, as it is used to show the tooltip for the snapshot.
+    if (isChromatic()) {
```
schilchSICKAG, line 364: As discussed, we should check if we can create a solution that does such checks automatically. Suggestion (pseudo code):

```
// ChartDecorator
import { html } from 'lit';
import type { Decorator } from '@storybook/web-components-vite';
import isChromatic from 'chromatic/isChromatic';

export const chartDecorator: Decorator = (story, context) => {
  if (!isChromatic()) {
    return story(context.args); // maybe story(context); is enough?
  }

  const originalPlayFn = context.play;

  // Dont know the type...
  const play = async (...args) => {
    const chart = canvasElement.querySelector('syn-chart');
    if (chart) {
      await new Promise<void>(resolve => {
        instance.on('finished', () => { resolve(); }
      });
    }

    // Dont know how this works
    if (originalPlayFn) {
      await originalPlayFn(...args);
    }
  }
}

File: packages/docs/stories/charts/zooming.stories.tsx [hunk 12]
```diff
@@ -0,0 +1,143 @@
...
+} from '../../src/helpers/component.js';
+import { generateFigmaPluginObject } from '../../src/helpers/figma.js';
+import { Chromatic_Modes_Sick_2025 } from '../../.storybook/modes.js';
+
+const meta: Meta = {
+  component: 'syn-chart',
+  parameters: {
+    chromatic: {
+      modes: Chromatic_Modes_Sick_2025,
+      pauseAnimationAtEnd: false,
```
schilchSICKAG: Why is this needed?

packages/docs/stories/components/checkbox-group.stories.ts
```diff
@@ -59,7 +59,7 @@ const meta: Meta = {
     chromatic: {
       modes: Chromatic_Modes_All,
     },
-    design: generateFigmaPluginObject('50279-43830'),
+    design: generateFigmaPluginObject('50279-39073'),
```
schilchSICKAG, line 62: Like discussed, I think I made a mistake here by using the wrong id in metadata. Also it came to me that we are now essentially duplicating the information as this is also available on the overview via a link from metadata. Hot take: Maybe we dont need `generateFigmaPluginObject` at all anymore because we do this ourselfs? Basically, its just a link sitting idle most of the time, hidden in a drawer. What do you think? Should we maybe just get rid of the duplicates by just removing the functionality alltogether?

packages/docs/stories/components/fieldset.stories.ts
```diff
@@ -56,7 +56,7 @@ const meta: Meta = {
     chromatic: {
       modes: Chromatic_Modes_All,
     },
-    design: generateFigmaPluginObject('41310-271865'),
+    design: generateFigmaPluginObject('50945-47173'),
```
schilchSICKAG, line 59: Still not sure if it is just the checkbox-group I messed up in metadata, but I was sure that all the others are linking to the correct item. Think we have to discuss what to link where. Also for the pagination etc.

packages/metadata/data/core/component/component__syn-checkbox-group.json
```diff
@@ -99,7 +99,7 @@
         {
           "description": {
             "type": "text",
-            "value": "Add descriptive help text to a switch with the help-text attribute. For help texts that contain HTML, use the help-text slot instead.\n\nThe help-text attribute should not be used to display error messages. To handle validation and error messaging, use syn-validate for proper error management."
+            "value": "Add descriptive help text to a checkbox-group with the help-text attribute. For help texts that contain HTML, use the help-text slot instead.\n\nThe help-text attribute should not be used to display error messages. To handle validation and error messaging, use syn-validate for proper error management."
```
schilchSICKAG, line 102: Thanks for fixing this :)

packages/metadata/src/internal/collectors/storybook/source/configs.ts
```diff
@@ -7,7 +7,7 @@ interface StorybookEntry {
 }
 
 const resolveComponentName = (prefix: string, baseId: string): string => {
-  if (prefix === 'charts' && baseId.startsWith('charts-features-')) {
+  if (prefix === 'charts' && (baseId.startsWith('charts-features-') || baseId.startsWith('charts-series-types-'))) {
```
schilchSICKAG: I think there will be more of those prefixes later on, right? If so, we could also go the clean route and do something like this:

```javascript
const allowedChartPrefixes = [
  'charts-features-',
  'charts-series-types-',
];

if (prefix === 'charts' && allowedChartPrefixes.some(baseId)) {
```

With this, you could just adjust the `allowedChartPrefixes` later on (and there will be new ones, right?). Ignore if there will be no more prefixes.

schilchSICKAG (changes requested):
Looks solid for a first go :)

packages/docs/stories/charts/zooming.stories.tsx
kirchsuSICKAG: This was a try to disable the animations of echarts and make the screenshots stable. Unfortunately it was useless.
I removed it again, as we now have a global play function for the charts, that (hopefully) make the screenshots stable

packages/metadata/src/internal/collectors/storybook/source/configs.ts
kirchsuSICKAG: I have adjusted it. I am not sure if we will need more, but it is possible.  so I think it makes sense

packages/docs/stories/components/checkbox-group.stories.ts
kirchsuSICKAG: Yeah I think you are right and we can just remove it, as we have the information anyways in the overview pages now

packages/components/src/components/chart/themes/theme.ts
kirchsuSICKAG: The ECharts docs are a bit vague here, but darkMode controls internal contrast behavior (mainly auto label/text rendering).
When set to auto, ECharts infers light vs dark from backgroundColor and may adjust some text/label styles accordingly.

Since we already provide explicit light and dark themes with our own styles, we do not want this automatic adjustment to override or interfere with our intended styling.
So I removed it to keep theme behavior fully deterministic and controlled by our custom theme definitions.

packages/_private/angular-demo/src/AllComponentParts/LineChart.ts
kirchsuSICKAG: I added you proposal, i like it 👍

packages/_private/react-demo/src/AllComponentParts/LineChart.tsx
kirchsuSICKAG: I added you proposal, i like it 👍

packages/components/src/components/chart/chart.component.ts
kirchsuSICKAG: I added a comment with all the thought on theming to our theming ticket: https://github.com/synergy-design-system/synergy-design-system/issues/1220

packages/docs/stories/charts/line-chart.stories.tsx
kirchsuSICKAG: oops 😁
Removed it

packages/docs/stories/charts/line-chart.stories.tsx
kirchsuSICKAG: oops 😁
Removed it

packages/docs/stories/charts/line-chart.stories.tsx
kirchsuSICKAG: oops 😁
Removed it

packages/docs/stories/charts/line-chart.stories.tsx
kirchsuSICKAG: Unfortunately the decorator is not working. I can't overwrite the play function there.

I added a shared play function, which can be added at story file level to the default meta object

```js
const meta: Meta = {
  component: 'syn-chart',
  ...
  play: waitForFinishedChartPlayFunction,
  ...
};
```

packages/docs/stories/components/checkbox-group.stories.ts
kirchsuSICKAG: For this PR, I'll leave it as is.

But I will remove the figma plugin and this utilties in one of the next PRs. Maybe with the dependency update

packages/docs/stories/components/fieldset.stories.ts
kirchsuSICKAG: For this PR, I'll leave it as is.

But I will remove the figma plugin and this utilties in one of the next PRs. Maybe with the dependency update

reichst1SICKAG (approved):
Design stuff approved

schilchSICKAG (approved):

PR #1353, opened 2026-08-13
feat: ✨ Full featured chart axes

### LINKED ISSUE #1238: feat: ✨ full featured chart axes

# Component

I suggest to add the feature to the following component: syn-chart (axes)

# Story

As a user of the Synergy Design System, I would like to use axes that a are styled according to the Figma design guide.

# Acceptance criteria

- [ ] All features/styles below are defaults and can be overwritten:
  - [x] Category axis
    - [x] labels
    - [x] icons
    - [x] title
    - [x] Positioning: left, middle, right and everything in between
    - [x] min/max value
  - [ ] Values axis
    - [x] labels
    - [ ] indicator (colored rectangular patch)
    - [x] icons
    - [x] shorthand labels
    - [x] min/max fraction digits
    - [x] local formatting (e.g. 1.2 or 1,2)
    - [x] title
    - [x] adjustable tick count
    - [x] adjustable number formatting
    - [x] positioning: left, middle, right and everything in between
    - [x] min/max value (default: none -> auto-sizing)
    - [x] optionally auto sizing based on input values (default: on)
    - [ ] multiple axes for support of various units / color-coded (series color matches axis color)
  - [x] category / value axes can be flipped
  - [x] Vertical axis width
  - [ ] Horizontal and vertical auxiliary lines can be defined: visibility and distance
  - [x] Label space justification on the x-Axis (SpaceGreedily/SpaceEvenly)

kirchsuSICKAG:
# Pull Request

## 📖 Description
This PR adds the full featured axes theming and examples.

### 🎫 Issues
Closes #1238 

### Formatter utility functions
I added formatter utility functions for the axis label formatting (see `packages/components/src/components/chart/configs/axes/formatter.ts`). 
- [ ] Does it make sense, that they are exported there or should they be exported somewhere else? Stakeholder need to do following to use them currently:
```js
import {
  numberFormatter, numberShorthandFormatter, unitFormatter,
} from '@synergy-design-system/components/chart/configs/axes/formatter.js';

const myFormatter = unitFormatter('°C')
```

### Indicator for multiply y-axis
- [ ] check out which is the correct handling for showing which series belongs to which line. What if several series belong to one of the multiple lines? Currently possible solutions: color-coded axes line with color of series(--> what to do when several lines to one axis?) or colored patch beneath the axis title

## 📑 Test Plan

Do as always

## ✅ DoD


.changeset/1238-full-featured-chart-axes.md
```diff
@@ -0,0 +1,9 @@
+---
+"@synergy-design-system/components": minor
+"@synergy-design-system/metadata": minor
+"@synergy-design-system/tokens": none
```
schilchSICKAG: Should be minor

packages/components/src/components/chart/configs/formatter.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @param unit - The unit string to append (e.g. `'kg'`, `'°C'`).
+ * @returns A formatter function that concatenates the value and unit with a space.
+ */
+export const unitFormatter = (unit: string) => (value: string | number) => `${value} ${unit}`;
+
+/**
+ * Maps powers of 10 (in steps of 3) to their SI prefix symbols.
+ */
+const SI_PREFIXES = new Map<number, string>([
```
schilchSICKAG, line 12: What is this? :)

packages/components/web-test-runner.config.js
```diff
@@ -10,17 +10,20 @@ const allBrowsers = [
   // @see https://github.com/modernweb-dev/web/issues/2374
   playwrightLauncher({
     concurrency: process.env.CI ? 1 : undefined,
+    createBrowserContext: async ({browser}) => browser.newContext({ locale: 'de-DE'}),
```
schilchSICKAG, line 13: Whats the reason for creating new contexts here?

schilchSICKAG (changes requested):
>  Does it make sense, that they are exported there or should they be exported somewhere else? Stakeholder need to do following to use them currently:

Yeah, probably we should have an export barrel that handles this and future exports, basically a single source export of sorts?

packages/components/src/components/chart/configs/formatter.ts
schilchSICKAG: Learned something new today. Thanks for explaining.

.changeset/1238-full-featured-chart-axes.md
kirchsuSICKAG: done

kirchsuSICKAG:
> > Does it make sense, that they are exported there or should they be exported somewhere else? Stakeholder need to do following to use them currently:
> 
> Yeah, probably we should have an export barrel that handles this and future exports, basically a single source export of sorts?

I moved the formatters out of the axes context up to general config context, as this formatter could also be used for other cases like tooltip content, other label texts etc.

I added a index.ts at packages/components/src/components/chart/index.ts so we now have a barrel export for such utility functions and other stuff

packages/mcp/test/watermarks/baseline.latest-release.json
```diff
@@ -34,7 +34,7 @@
     },
     {
       "id": "component-info:examples:syn-chart",
-      "tokens": 8935
+      "tokens": 13008
```
schilchSICKAG, line 37: Should we maybe remove this entry for the time being as it is regulary updated?

schilchSICKAG (approved):
Seems nice from my pov. Just thinking about removing the watermark until we are really ready as it must be annoying to update it all the time. What do you think?

packages/mcp/test/watermarks/baseline.latest-release.json
kirchsuSICKAG: Yes, this might be a good idea. I will do it in my next PR

PR #1363, opened 2026-08-25
feat: ✨ added donut-chart

### LINKED ISSUE #1356: feat: ✨ donut chart
# Component

# Story
As a user of the Synergy Design System, I would like to show donut chart in my application so values are related to each other in an appealing and clear way.

# Acceptance criteria

- [x] definable color pallet
- [x] one complete grey circle in the center
- [x] show arcs for each value in a certain angle
- [x] value arcs have definable color
- [x] value arcs have definable size
- [ ] values are shown as strings on the outer edge of each value arc (can be switched off)
- [x] labels can be overwritten with any text
- [ ] format values (decimal places + shorthand)

klaarseSICKAG:
# Pull Request

## 📖 Description
This PR adds a new "synDonut" series as well as a "seriesDonut" preset to the syn-chart

### 🎫 Issues
Closes #1356 

## ✅ DoD

### DIFF
--- a/packages/components/CHARTING_README.md
+++ b/packages/components/CHARTING_README.md
@@ -419,6 +419,62 @@ chart.config = handle =>
 // ]
 ```
 
+### Gauge series presets
+
+| Preset function | Options                    | Description                                                                                                                                               |
+| --------------- | -------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `seriesGauge`   | `GaugeSeriesPresetOptions` | Adds a custom `synGauge` series. Renders the gauge progress arc and optional outer sections ring, plus value/min/max labels and optional trend indicator. |
+
+`GaugeSeriesPresetOptions` supports the following fields:
+
+| Option                | Type                        | Description                                                                                                                 | Default                                                                          |
+| --------------------- | --------------------------- | --------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------- |
+| `value`               | `number` _(required)_       | Current value of the gauge.                                                                                                 |                                                                                  |
+| `min`                 | `number`                    | Minimum value of the gauge scale.                                                                                           | 0                                                                                |
+| `max`                 | `number`                    | Maximum value of the gauge scale.                                                                                           | 100                                                                              |
+| `icon`                | `string`                    | SVG data URL rendered as an image below the value.                                                                          |                                                                                  |
+| `color`               | `string`                    | Custom color of the progress arc. Overrides automatic section-based coloring when set.                                      |                                                                                  |
+| `backgroundColor`     | `string`                    | Custom background color of the progress arc track.                                                                          | --syn-chart-track-color                                                          |
+| `formatter`           | `GaugeFormatterOptions`     | Formatter functions for the displayed gauge labels.                                                                         |                                                                                  |
+| `formatter.value`     | `(value: number) => string` | Formatter applied to the displayed gauge value.                                                                             | String(value)                                                                    |
+| `formatter.min`       | `(value: number) => string` | Formatter applied to the displayed minimum label.                                                                           | String(value)                                                                    |
+| `formatter.max`       | `(value: number) => string` | Formatter applied to the displayed maximum label.                                                                           | String(value)                                                                    |
+| `sections`            | `GaugeSectionsOptions`      | Outer section boundaries and colors.                                                                                        |
+| `sections.show`       | `boolean`                   | Shows the outer section ring when enabled.                                                                                  | false                                                                            |
+| `sections.boundaries` | `number[]`                  | Boundary values for outer sections (for example `[0, 20, 60, 100]`). Adjacent pairs define ranges.                          | [0, 20, 60, 100]                                                                 |
+| `sections.colors`     | `string[]`                  | Colors for each outer section range. Repeated cyclically when fewer colors than ranges are provided.                        | [--syn-namur-success-color, --syn-namur-warning-color, --syn-namur-error-color ] |
+| `trend`               | `GaugeTrendOptions`         | Trend indicator text and icon options.                                                                                      |                                                                                  |
+| `trend.show`          | `boolean`                   | Shows the trend indicator when enabled.                                                                                     | false                                                                            |
+| `trend.direction`     | `'up' \| 'down'`            | Direction of the trend indicator.                                                                                           | 'up'                                                                             |
+| `trend.value`         | `string`                    | Trend label text shown in the indicator pill.                                                                               |                                                                                  |
+| `trend.iconUp`        | `string`                    | SVG data URL used as the icon when `trend.direction` is `'up'`. Falls back to the Synergy default up icon when not set.     | Default arrow up icon                                                            |
+| `trend.iconDown`      | `string`                    | SVG data URL used as the icon when `trend.direction` is `'down'`. Falls back to the Synergy default down icon when not set. | Default arrow down icon                                                          |
+
+Example with custom colors and formatters:
+
+```ts
+import { formatter } from "@synergy-design-system/components/components/chart/index.js";
+
+chart.config = handle =>
+  handle.seriesGauge({
+    min: 0,
+    max: 1000,
+    value: 450,
+    unit: "kWh",
+    color: "#2f9e44",
+    backgroundColor: "#e8f5e9",
+    formatter: {
+      value: formatter.unitFormatter("kWh"),
+      min: formatter.unitFormatter("kWh"),
+      max: formatter.unitFormatter("kWh"),
+    },
+  });
+```
+
+Array merge strategy:
+- `seriesGauge({...})` uses `arrayStrategy: 'append'`.
+- The generated `synGauge` series entry is appended to `series`.
+
 ### Donut series presets
 
 | Preset function | Options                    | Description                                                                                                                                                   |
@@ -463,6 +519,20 @@ chart.config = handle =>
       },
       { value: 8, name: "Other", color: "#5e97fc" },
     ],
+  handle.seriesGauge({
+    min: 10,
+    max: 120,
+    value: 72,
+    sections: {
+      boundaries: [10, 40, 70, 120],
+      colors: ["#d92f2f", "#f5a623", "#2f9e44"],
+      show: true,
+    },
+    trend: {
+      direction: "down",
+      show: true,
+      value: "6.5%",
+    },
   });
 ```
 
@@ -499,7 +569,7 @@ import {
 
 | Function                   | Signature                                                                                      | Description                                                                                                    |
 | -------------------------- | ---------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
-| `unitFormatter`            | `(unit: string) => (value: string \| number) => string`                                        | Appends a unit with a space (for example `42` -> `42 kg`).                                                     |
+| `unitFormatter`            | `(unit: string) => (value: string \| number) => string`                                        | Appends a unit (for example `42` -> `42kg`).                                                                   |
 | `numberFormatter`          | `(locale?: string, options?: Intl.NumberFormatOptions) => (value: string \| number) => string` | Localizes numeric values via `Intl.NumberFormat`. Non-numeric values are returned unchanged.                   |
 | `numberShorthandFormatter` | `(locale?: string, options?: Intl.NumberFormatOptions) => (value: string \| number) => string` | Formats values with SI prefixes for large/small magnitudes (for example `1500000` -> `1.5M`, `0.002` -> `2m`). |
 
@@ -543,7 +613,7 @@ shorthand(0.002); // '2m'
 
 ## Bundle Size
 
-`syn-chart` uses [ECharts tree-shaking](https://echarts.apache.org/en/tutorial.html#Use%20ECharts%20with%20bundler%20and%20NPM) internally. Only the modules that are actually needed (currently `LineChart`, `CanvasRenderer`, `GridComponent`, `LegendComponent`, `TitleComponent`, `TooltipComponent`) are imported and registered via ECharts' `use([...])`.
+`syn-chart` uses [ECharts tree-shaking](https://echarts.apache.org/en/tutorial.html#Use%20ECharts%20with%20bundler%20and%20NPM) internally. Only the modules that are actually needed (currently `LineChart`, `GaugeChart`, `CanvasRenderer`, `GridComponent`, `LegendComponent`, `TitleComponent`, `TooltipComponent`, `DataZoomComponent`) are imported and registered via ECharts' `use([...])`.)
 
 > ⚠️ You do **not** need to register anything yourself.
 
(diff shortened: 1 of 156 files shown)


packages/components/src/components/chart/configs/donut-series/presets.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Renders a donut chart made of two concentric rings: a static inner track
+ * ring and a segmented outer ring. The data values are distributed evenly
+ * around the outer ring, with each segment sized proportionally to its value
+ * (normalized to 360 degrees).
+ *
+ * @param {DonutSeriesPresetOptions} [options] Preset options.
+ * @param {DonutDataItem[]} [options.data] Data values/items used to size and label the outer ring segments.
+ * @param {number} [options.data[].value] Numeric value of the segment; determines the slice angle relative to the other values.
+ * @param {string} [options.data[].name] Label shown for the segment.
+ * @param {string} [options.data[].icon] Optional icon as SVG data url used to render a segment icon alongside the label.
```
klaarseSICKAG: it might make sense to call it prefixIcon, so it is clear where the icon is places it its also future proof if we decide to enable postfixIcons as well.

packages/components/src/components/chart/configs/donut-series/presets.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param {[number|string, number|string]} [options.center] Center position within the donut layout area. Each value accepts pixels or percentages.
```
klaarseSICKAG: what is the default?

packages/components/src/components/chart/configs/donut-series/presets.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param {number|string} [options.radius] Outer donut radius in pixels/percentage.
```
klaarseSICKAG: what is the default?

packages/metadata/data/layers/examples/component/component__syn-chart.md
```diff
@@ -1375,6 +1375,201 @@ The data zoom slider (the dataZoom option with type: 'inside') allows users to d
...
+              value: 30,
+              name: "React",
+            },
+            {
+              value: 20,
+              name: "Vue",
+            },
+          ],
+        })
+        .legendShow();
```
klaarseSICKAG, line 1572: donut gets significantly smaller if disabling 2 or 3 values via the legend

kirchsuSICKAG: That is because the label is always shown at the middle of the section, which is at the top in case of all others are toggled away. Because there is not enough space for the full circle and the text, the circle adapts its radius

packages/components/src/components/chart/configs/donut-series/presets.ts
kirchsuSICKAG: I added the default

packages/components/src/components/chart/configs/donut-series/presets.ts
kirchsuSICKAG: i added the default

packages/components/src/components/chart/configs/donut-series/presets.ts
kirchsuSICKAG: I renamed the property to "prefixIcon"

packages/components/src/components/chart/configs/donut-series/donut-series-view.ts
```diff
@@ -0,0 +1,556 @@
...
+  polarPoint,
+  sanitizeFiniteNumber,
+} from '../utilities.js';
+import type { ExtensionAPI, GlobalModel, LayoutValue } from '../types.js';
+
+const PIXEL_VALUE_PATTERN = /^[+-]?\d+(?:\.\d+)?$/;
+const PERCENT_VALUE_PATTERN = /^[+-]?\d+(?:\.\d+)?%$/;
+
+/** Converts a pixel/percent input to pixels relative to the given base size. */
+const toPixels = (value: LayoutValue | undefined, baseSize: number, fallback = 0): number => {
```
schilchSICKAG, line 30: I wonder if we don´t need this on other types of charts too. Maybe we could move this to a generic helper library?

packages/components/src/components/chart/configs/donut-series/donut-series-view.ts
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  return Math.max(0, toPixels(radius, radiusBase, radiusBase / 2));
+};
+
+/**
+ * Returns `true` when the radius is an absolute value (number or non-percent string),
+ * meaning adaptive scaling should not change it.
+ */
+const isFixedRadius = (radius: LayoutRadiusInput): boolean => {
```
schilchSICKAG, line 119: This duplicates parts of the logic in the toPixels function. Maybe we can rename this function and use it directly in toPixels to avoid duplicate code?

packages/components/src/components/chart/configs/donut-series/types.ts
```diff
@@ -0,0 +1,93 @@
+import type { LayoutValue } from '../types.js';
+
+export type LayoutBounds = {
```
schilchSICKAG, line 3: I think its a good rules of thumb to add comments when we export types, what do you think? Are the types fine like this? I mean they are meaningfully named and this is just an observation.

packages/components/src/components/chart/configs/gauge-series/gauge-series-view.ts
```diff
@@ -1,21 +1,24 @@
 import { ChartView, graphic } from 'echarts/core.js';
```
schilchSICKAG, line 1: After looking at the folder structure, maybe it would benefit from another folder that gets inserted? Something like:

```bash
+ chart
  + config
    + series
      + donut
      + gauge
      + line
```

This would make it easier to structure later on as we may build much more series graphs, right? However, I am fine with this too if we don´t export anything from here for users directly.

packages/components/src/components/chart/configs/presets/index.ts
```diff
@@ -1,4 +1,5 @@
 import * as axes from '../axes/presets.js';
+import * as donutSeries from '../donut-series/presets.js';
```
schilchSICKAG, line 2: This would also be easier to do when we have a barrel export chart/configs/series.ts that exports all the things from a hypotetically existing chart/config/series folder. You could eventually use this for the typeof stuff in line 22 below by just extending it like the following:

```
import * as series from './configs/series.js';

export const ChartPresets = {
  ...axes,
  ...donutSeries,
  ...gaugeSeries,
  ...legend,
  ...lineSeries,
  ...tooltip,
} as typeof axes & keyof typeof series & typeof legend & typeof tooltip;

File: packages/components/CHARTING_README.md [hunk 10]
```diff
@@ -450,27 +450,6 @@ chart.config = handle =>
 | `trend.iconUp`        | `string`                    | SVG data URL used as the icon when `trend.direction` is `'up'`. Falls back to the Synergy default up icon when not set.     | Default arrow up icon                                                            |
 | `trend.iconDown`      | `string`                    | SVG data URL used as the icon when `trend.direction` is `'down'`. Falls back to the Synergy default down icon when not set. | Default arrow down icon                                                          |
 
-Example:
```
schilchSICKAG, line 453: Just for curiosity, why was this example dropped?

packages/docs/stories/charts/axes.stories.tsx
```diff
@@ -31,6 +31,7 @@ const meta: Meta = {
   component: 'syn-chart',
   parameters: {
     chromatic: {
+      delay: 300,
```
schilchSICKAG, line 34: Does this remove the test flakiness?

packages/docs/stories/charts/donut-chart.stories.tsx
```diff
@@ -0,0 +1,275 @@
...
+  },
+  play: waitForFinishedChartPlayFunction,
+  tags: ['Charting', 'Data Visualization'],
+  title: 'Charts/Series Types/Donut Chart',
+};
+export default meta;
+
+type Story = StoryObj;
+
+export const Default: Story = {
```
schilchSICKAG, line 56: The default stories does not seem to match the current version of Synergy-Chart-Library, which shows 8 colored parts instead of 4. I like your example more, because it also shows uneven distribution, but wanted to point it out regardless.

packages/docs/stories/charts/donut-chart.stories.tsx
```diff
@@ -0,0 +1,275 @@
...
+    docs: {
+      description: {
+        story: generateStoryDescription('chart', 'donut-series-colors'),
+      },
+    },
+  },
+  render: () => html`
+    <syn-chart id="donut-colors"></syn-chart>
+    <script type="module">
+      // import { ResolvedTokens as ChartTokens } from '@synergy-design-system/tokens/charts/resolved';
```
schilchSICKAG, line 94: I would opt to remove this

packages/docs/stories/charts/donut-chart.stories.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+      const charts = document.querySelectorAll('#donut-colors');
+      const getChartColor = (token) => {
+        return ChartTokens[token]['light'];
+      };
+
+      charts.forEach(chart => {
+        chart.config = handle => handle
+        .baseConfig({ color: [getChartColor('SynChartSequential01_100'), getChartColor('SynChartSequential01_90'), getChartColor('SynChartSequential01_80'), getChartColor('SynChartSequential01_70'), getChartColor('SynChartSequential01_60'), getChartColor('SynChartSequential01_50'), getChartColor('SynChartSequential01_40')] })
```
schilchSICKAG, line 103: Can you run this through prettier to make it multi line? The line is so long it took my whole scroll wheel to see the end :)

packages/tokens/src/figma-tokens/_docs.json
```diff
@@ -967,6 +967,86 @@
           "type": "text"
         }
       },
+      "donut-series-default": {
```
schilchSICKAG, line 970: Needs a copy in metadata. I know, we are dragging this quite long :(

packages/docs/stories/charts/donut-chart.stories.tsx
```diff
@@ -0,0 +1,275 @@
...
+        .seriesDonut({
+          radius: '90%',
+          data: [10, 20, 30, 40],
+        });
+      });
+    </script>
+  `,
+};
+
+export const Center: Story = {
```
schilchSICKAG, line 175: I am not sure that Center is the correct name for this story. I needed to look at the code to see that it is the opposite: You define a different center position instead of centering. Maybe it should be more like `Adjust center` or something?

packages/docs/stories/charts/donut-chart.stories.tsx
```diff
@@ -0,0 +1,275 @@
...
+    <syn-chart id="donut-labels"></syn-chart>
+    <script type="module">
+      const charts = document.querySelectorAll('#donut-labels');
+
+      charts.forEach(chart => {
+        chart.config = handle => handle
+        .seriesDonut({
+          data: [
+            {
+              value: 50,
```
schilchSICKAG, line 130: Would be cool if we could use the real numbers here. They can be seen in the internal page.

schilchSICKAG (changes requested):
A couple of small adjustments and we should be good to go.
One thing I noticed is that scaling is done via width, not height. I would have thought that if the chart has not enough height to be shown full screen, it would shrink. Don´t know if this is a common problem, though.

packages/docs/stories/charts/donut-chart.stories.tsx
kirchsuSICKAG: done

packages/docs/stories/charts/donut-chart.stories.tsx
kirchsuSICKAG: done

packages/tokens/src/figma-tokens/_docs.json
kirchsuSICKAG: done

packages/docs/stories/charts/donut-chart.stories.tsx
kirchsuSICKAG: done

packages/docs/stories/charts/donut-chart.stories.tsx
kirchsuSICKAG: I added a more explanatory comment to this, so users additionally get an explanation, why this is needed

packages/components/src/components/chart/configs/presets/index.ts
kirchsuSICKAG: I added a series.ts as barrel import as proposed

packages/components/src/components/chart/configs/series/gauge/gauge-view.ts
```diff
@@ -1,21 +1,24 @@
 import { ChartView, graphic } from 'echarts/core.js';
```
kirchsuSICKAG: nice idea. I adjusted the folder structure

packages/components/src/components/chart/configs/series/donut/types.ts
```diff
@@ -0,0 +1,93 @@
+import type { LayoutValue } from '../types.js';
+
+export type LayoutBounds = {
```
kirchsuSICKAG: all types now have descriptions

packages/components/src/components/chart/configs/donut-series/donut-series-view.ts
kirchsuSICKAG: I created a third method, which can is reused in boths functions

packages/components/src/components/chart/configs/donut-series/donut-series-view.ts
kirchsuSICKAG: done

packages/components/src/components/chart/configs/series/gauge/gauge-view.ts
schilchSICKAG: Looks way better :)

schilchSICKAG (approved):
Looks fine from my pov

packages/components/src/components/chart/configs/series/donut/donut-model.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @returns The normalized donut data item.
+ */
+const normalizeDonutDataItem = (dataItem: DonutDataValue): ResolvedDonutDataItem => {
+  if (typeof dataItem !== 'object') {
+    const value = sanitizeFiniteNumber(dataItem);
+    return { label: String(value), value };
+  }
+
+  const value = sanitizeFiniteNumber(dataItem.value);
+  const hasLabel = Object.prototype.hasOwnProperty.call(dataItem, 'label');
```
schilchSICKAG: `Object.prototype.hasOwnProperty.call` is very old school. We could move to `Object.hasOwn(obj, 'key')` instead which is shorter and native,too. Works in all our latest supported browsers (checked via canIUse and our 3.0 browser baseline).

I would suggest to remove all occurences of `Object.prototype...` which this pattern as its just shorter

schilchSICKAG (approved):
Still an approve, would just remove the Object.hasOwn stuff, but that can also be done later on (as long as we dont forget it)

PR #1380, opened 2026-09-15
fix: 🐛 noSpinButtons doesn't apply attribute for React wrapper

### LINKED ISSUE #1375: fix: 🐛 noSpinButtons doesn't apply attribute for React wrapper

# Version

The bug occurred with the following version of the component library: 3.24.0

# Component

The issue arises when using the following component(s): syn-input

# Environment

The issue occurred with the following operating system(s): Windows 11
The issue occurred in the following browser(s): Microsoft Edge Version 152.0.4191.66 (Official build) (64-bit)

# Description

Providing `noSpinButtons` attribute for `SynInput` in React 18 (also 19) doesn't apply `no-spin-buttons` correctly to the web component. Seems to be a React wrapper issue, as using `no-spin-buttons` version of the attribute correctly applies it.
The spin buttons are not visible as they shouldn't, but it does not correctly apply the css. It does apply  `:host([size="medium"][type="number"]:not([no-spin-buttons]))` although it should not.
This results into different width sizing behavior depending on adding `no-spin-buttons` or `noSpinButtons` (see gif)

Image

Reproducible examples: https://stackblitz.com/edit/vitejs-vite-wcemkmv5?file=src%2FApp.tsx

Working version:

```tsx
<SynInput type="number" value="5" no-spin-buttons />
```

Non-working version:

```tsx
<SynInput type="number" value="5" noSpinButtons />
```

# Acceptance criteria

- [ ] `no-spin-buttons` attribute is correctly applied to `syn-input` web component when used on React wrapper

schilchSICKAG:
# Pull Request

## 📖 Description

This release fixes an issue that could lead to wrong sizes for `syn-input` when `noSpinButtons` is provided as a property. Reason is that react does not use attributes at all, falling back to properties. This means reflection for `<syn-input no-spin-buttons />` is not done. Because of this, the size of a `syn-input` does not match the expected size.

### 🎫 Issues

Closes #1375 

## 👩‍💻 Reviewer Notes

Fixed by adding reflection to the property. I also refactored the input styles and removed some other unused ones.

## 📑 Test Plan

Use `<syn-input no-spin-buttons />` and `<SynInput noSpinButtons />`. Both should behave the same on resize and have the `no-spin-button` attribute in the DOM.

## ✅ DoD


packages/components/src/components/input/input.styles.ts
```diff
@@ -147,109 +173,132 @@ export default css`
...
+    padding-right: var(--input-password-toggle-padding-inline-start);
   }
 
-  .input--large .input__clear,
-  .input--large .input__password-toggle {
-    width: calc(1em + var(--syn-input-spacing-large) * 2);
+  /* stylelint-disable-next-line plugin/no-unsupported-browser-features */
+  .input--large .input__clear:has(+ .input__password-toggle) {
+    padding-right: var(--input-password-toggle-padding-inline-start);
   }
```
kirchsuSICKAG: This can be simplified by just removing the sizes classes in front and having only one of them (like with the css styling before this):


```suggestion
  /* stylelint-disable-next-line plugin/no-unsupported-browser-features */
  .input__clear:has(+ .input__password-toggle) {
    padding-right: var(--input-password-toggle-padding-inline-start);
  }
```

packages/components/src/components/input/input.styles.ts
```diff
@@ -147,109 +173,132 @@ export default css`
...
+    --input-clear-padding-inline-end-with-suffix: var(--syn-spacing-medium);
+    --input-clear-padding-inline-start: var(--syn-spacing-medium);
+    --input-control-height: calc(var(--syn-input-height-large) - var(--syn-input-border-width) * 2);
+    --input-control-padding: var(--syn-spacing-small) var(--syn-input-spacing-large);
+    --input-font-size: var(--syn-input-font-size-large);
+    --input-height: var(--syn-input-height-large);
+    --input-icon-size: var(--syn-font-size-2x-large);
+    --input-number-divider-height: var(--syn-font-size-2x-large);
+    --input-number-divider-margin: 0 var(--syn-spacing-x-small);
+    --input-number-stepper-margin-inline-end: var(--syn-spacing-x-small);
```
kirchsuSICKAG, line 236: The css variable `--input-number-stepper-margin-inline-start` is missing for size="large"


```suggestion
    --input-number-stepper-margin-inline-start: 0;
    --input-number-stepper-margin-inline-end: var(--syn-spacing-x-small);
```

packages/components/src/components/input/input.styles.ts
```diff
@@ -147,109 +173,132 @@ export default css`
...
-  .input__clear,
-  .input__password-toggle {
-    display: inline-flex;
-    align-items: center;
-    justify-content: center;
-    font-size: inherit;
-    color: var(--syn-input-icon-color);
-    border: none;
-    background: none;
-    padding: 0;
```
kirchsuSICKAG, line 250: It seems like we still need the `padding: 0` or at least a `padding-block: 0`. 
That is the reason, why chromatic complains about visual changes. 
Not having set the top and bottom padding explicitly to zero leads to having different top and bottom paddings in chrome (1px top, 1px bottom) and in safari (2px top, 3px bottom).

Before:

Image

Now in chrome:

Image

Now in safari:

Image

kirchsuSICKAG (changes requested):

packages/components/src/components/input/input.styles.ts
schilchSICKAG: Thanks, fixed.

packages/components/src/components/input/input.styles.ts
schilchSICKAG: You are right, fixed.

packages/components/src/components/input/input.styles.ts
schilchSICKAG: Fixed

kirchsuSICKAG (approved):

PR #1388, opened 2026-09-18
fix: 🐛 dependency updates (#258)

### LINKED ISSUE #1081: chore: 🔧 Migrate npm publish to trusted publishers
# Description

Due to the security strengthening approach from NPM the according tokens have only a very limited lifetime.
The traditional tokens should be replaces by the proposed OIDC approach.

# Rationale

Frequent token rotation is cumbersome.

NPM will phase out long running tokens. Old tokens are still valid and don´t need to be changed, however we should upgrade to the [latest recommended publishing workflow](https://docs.npmjs.com/trusted-publishers) to make sure we are correctly aligned.

# Details
Background information: https://github.blog/changelog/2025-09-29-strengthening-npm-security-important-changes-to-authentication-and-token-management/

OIDC approach: https://docs.npmjs.com/trusted-publishers

### LINKED ISSUE #1244: chore: 🔧 Remove unneeded devDependencies
# Description

As per [this Blog-Post](https://nodesource.com/blog/nodejs-features-replacing-npm-packages), there are now some pretty stable build in alternatives for commonly used `devDependencies` like `globby` or `chalk` or `rimraf`. We should reduce our dependency tree by leveraging those standard utilities instead.

- [x] Remove rimraf from all packages
- [x] Remove `globby` from all packages
- [x] Remove `chalk` from all packages
- [x] Remove custom solutions like mocha/chai for `node:test` and `node:assert` where appropriate

# Rationale

As a developer working on the Synergy Design System, I want to have less unneeded `devDependencies`.

schilchSICKAG:
# Pull Request

## 📖 Description

This release updates dependencies for all packages.
It also introduces `pnpm@12` as the default package manager.

### 🎫 Issues

Closes #258
Closes #1081 
Closes #1244

## 👩‍💻 Reviewer Notes

- This branch now uses NPMs Trusted Publishing. We cannot test this without merging, but I made it fallback save by leaving the token auth as a second variant.
- I removed `globby` entirely from all projects. Its still there as a transitive dependencies, but gone for our own software. One point closer to #1244
- `rimraf` also went the way of the dodo 🦤 in this release.
- Chalk was already removed everywhere, nothing to be done there
- Last thing was getting rid of `chai` for tests in the metadata package. Was done automatically and substituted with `node:assert`.
- Storybook had some more __"special"__ behavior that I had to tackle. Note the order of the argTypes table was once again altered - nothing we could do about this one unfortunately. I also had to patch our custom logic as otherwise he would now duplicate all attributes and properties. I adjusted it so it only shows properties that have **no** attribute associated. Hope this turn out well.
- Because of how Storybook now really removed default props, some metadata stories have changed too. Also notice that he now uses `booleanValue="true"` in the output instead of `booleanValue=""`. I did not find a way around this as it seems to be an internal wc-toolkit decision.
- Playwright was updated, too.
- Assets and Tokens Packages where tested to make sure they still receive the Figma assets.

## ✅ DoD


packages/styles/README.md
```diff
@@ -75,15 +75,16 @@ import "@synergy-design-system/styles/typography.css";
...
+- dist.css
+  - link-list.css
+  - link.css
+  - tables.css
+  - typography.css
+  - typography.css
+  - typography.css
+  - typography.css
+  - tables.css
+  - tables.css
```
kirchsuSICKAG: Why the change here? Nothing should have changed (and also didn't) for the dist output files. Also there is no such `dist.css` file and other files occure several times. 
I would just remove the change

packages/metadata/data/core/token/token:tokens-charts-js-resolved-d-ts.json
```diff
@@ -1,31 +0,0 @@
-{
-  "custom": {
```
kirchsuSICKAG, line 2: Why are the resolved token files deleted? They are there and I cannot find a change, that would resolve into the deletion

kirchsuSICKAG (changes requested):

packages/metadata/data/core/token/token:tokens-charts-js-resolved-d-ts.json
schilchSICKAG: Because it has the invalid `:`. The file is still there as token__tokens-chartjs..., but this was a leftover.

kirchsuSICKAG (approved):

